PR #56, opened 2015-02-07
Update the guide and API for two-way binding with v-with in 0.11.5

oddic:
Why was it changed to a unidirectional flow in 0.11 anyway? Why was it changed back?


yyx990803:
Thanks a lot!

PR #62, opened 2015-03-11
Updated the demo to use v-class

crswll:
I noticed you were using `class="{{done ? 'done' : ''}}"` instead of `v-class="done: done"`. I wasn't sure if there was a particular reason for this but I figured I'd submit a pull request either way.


yyx990803:
The two are largely equivalent, but in this particular case I think the interpolation is less confusing. `v-class` is more useful when there are multiple classes to be toggled.

crswll:
Sweet.

PR #76, opened 2015-06-08
Added missing links between pages

jperasmus:
Hi,

Very basic fixes just to add a couple of missing links between pages I found when reading through the docs.


yyx990803:
Thanks a lot! Just a note that there's no need to change `themes/vue/source/js/vue.js`, since when Vue releases a new version it will be updated with new source code. It should be fixed in the main repo itself. (which I already did).

PR #77, opened 2015-06-09
Tips for beginners

azamat-sharapov:
Let me know if I need to change anything or add some more.
Did my best, but feel free to change whatever you want, if you merge it :)

### DIFF
--- a/source/_posts/tips-for-beginners.md
+++ b/source/_posts/tips-for-beginners.md
@@ -143,7 +143,14 @@ Instances can also communicate using [Events](/api/instance-methods.html#Events)
 
 <p class="tip">In ECMAScript 5 there is no way to detect when a new property is added to an Object, or when a property is deleted from an Object. To deal with that, observed objects will be augmented with three methods: $add(key, value), $set(key, value) and $delete(key). These methods can be used to add / delete properties from observed objects while triggering the desired View updates. The difference between $add and $set is that $add will return early if the key already exists on the object, so just calling obj.$add(key) won’t overwrite the existing value with undefined.</p>
 
-The quote above is taken from guide and is important to mention, because many beginners run into issues with it. If you need to add new or delete existing data properties, use augmented methods as mentioned above. Another important thing you should know is, when you change data-bound Array directly using indices, Vue.js will not pick your changes and so View will not be updated. Again, you can use augmented methods to notify Vue.js about those changes. Also, augmented methods can take path as a first argument, which is useful. See example below.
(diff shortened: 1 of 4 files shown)


source/_posts/tips-for-beginners.md
```diff
@@ -0,0 +1,270 @@
...
+<div id="demo">
+  <!-- works, because this is parent template -->
+  <a v-on="click: parentMethod">click here</a>
+
+  <!-- does NOT work, because childMethod does not exist in this scope -->
+  <a v-on="click: childMethod"></a>
+</div>
+```
+
+## Communication between instances
```
yyx990803: I'm thinking about whether we should list the use of events as best practice here. From what I've seen so far it is very easy to end up with complex state change flow when using events, because you have to simulate how events travel inside your component tree correctly to reason about what is happening.

Since I plan to release this post after 0.12, I think we should emphasis the usage of passing down methods as prop callbacks as the first recommended approach, with events filling in the gap of other scenarios.

yyx990803:
Great job! I think we will release this post after 0.12 launch, I'll take a closer look after that, but it looks great for now :)

source/_posts/tips-for-beginners.md
azamat-sharapov: I saw your message about "passing callbacks" after writing this all actually :D It was unexpected, but good functionality to have. I was going to update this section, but decided to wait for your comments. I'll update it soon to demonstrate passing callbacks instead of events, but will probably mention using events too (just mention, without examples).

azamat-sharapov:
Updated *Communication between instances* to emphasis `props` usage. Let me know if something else needs to be edited.

Thanks!

source/_posts/tips-for-beginners.md
```diff
@@ -0,0 +1,253 @@
...
+    // adds new item and sets 'value' of it
+    this.arrayItems.$add(0, 'someValue')
+    
+    // Head Up! This will *not* change arrayItems[0], use $set() for that
+    // arrayItems[0] will still have 'someValue' as value
+    this.arrayItems.$add(0, 'newValue')
+
+    // using path! so objectItems becomes
+    // {a: { b: 'value' } }
+    this.objectItems.$add('a.b', 'value')
```
yyx990803: Hmm, $add actually doesn't support paths. The only methods that accept paths are `vm.$get` and `vm.$set`.

azamat-sharapov: Ooops! will fix now.

PR #79, opened 2015-06-13
Fix api options 

kazupon:


yyx990803:
wow, missed that totally, thanks!

PR #84, opened 2015-06-18
update $remove arguments reference

miljan-aleksic:
Additionally I have found in [this](http://vuejs.org/examples/firebase.html) example ```app.users.$remove(user)``` where user seems an object. I'm not sure about that one as in the code I don't see cb being used in some other way.


yyx990803:
Thaks, that `$remove` is an augmented method on observed arrays.

The `withTransition` flag is largely internal - end users should not have to care about it, that's why I left it out.

miljan-aleksic:
Ah, got it. Thanks :)

PR #95, opened 2015-07-24
Fix typo

kazupon:


source/guide/index.md
```diff
@@ -49,7 +49,7 @@ A slightly modified plain JavaScript object.
 vm.$data // The Model
 ```
 
-In Vue.js, models are simply plain JavaScript objects, or **data objects**. Once an object is used as data inside a Vue instnace, it becomes **reactive**. You can manipulate their properties and Vue instances that are observing them will be notified of the changes. Vue.js achieves transparent reactivity by converting the properties on data objects into ES5 getter/setters. There's no need for dirty checking, nor do you have to explicitly signal Vue to update the View. Whenever the data changes, the View is updated on the next frame.
```
kazupon: `instnace` -> `instance`

yyx990803:
Already fixed when addressing #97 :) Thanks anyway

PR #103, opened 2015-08-07
Added documentation of ability to use boolean expressions in v-show

afischer:


yyx990803:
This is actually not `v-show` specific - all Vue.js directives support inline expressions, as documented [here](http://vuejs.org/guide/directives.html#Inline_Expressions). Thanks for the PR though.

PR #109, opened 2015-08-25
Add vue-router, vue-async-data in extending.md

iJackUA:
It seems these libs a worth mentioning in official docs


yyx990803:
thanks!

PR #117, opened 2015-09-07
Add custom filter function syntax explanation

StoyanDimitrov:


source/api/filters.md
```diff
@@ -139,6 +139,7 @@ Or, just pass in a custom filter function:
 ``` html
 <div v-repeat="user in users | filterBy myCustomFilterFunction">
 ```
+Where [myCustomFilterFunction] is regular [vm] method accepting three arguments [user], [index] and [users] from the example above.
```
yyx990803: Can we use backticks (`) instead of square brackets here?

source/api/filters.md
```diff
@@ -139,6 +139,7 @@ Or, just pass in a custom filter function:
 ``` html
 <div v-repeat="user in users | filterBy myCustomFilterFunction">
 ```
+Where `myCustomFilterFunction` is regular `vm` method accepting three arguments `user`, `index` and `users` from the example above.
```
yyx990803: It probably make for sense to point out here that the method is expected to be a function that you'd pass to `Array.prototype.filter`.

yyx990803:
This PR is no longer relevant for the 1.0 docs - feel free to re-submit against 0.12 branch.

PR #125, opened 2015-09-22
"will refuse the set" vs "will refuse to set"

SERVANT14:


yyx990803:
Thanks for all the fixes!

PR #152, opened 2015-10-24
v-html="html" should be same as {{{html}}}

okoala:


yyx990803:
Thanks, can you submit against master instead?

okoala:
Yes

PR #177, opened 2015-11-04
Fix `v-for`'s documentation (Example #2)

jamen:
No major fix.  Just a minor error in the example I noticed.

The HTML provided in the example doesn't contain the hyphens that the output does...


yyx990803:
Thanks!

jamen:
Should I have done it for the HTML on the gh-pages branch as well? Or is that automatic?

yyx990803:
It's automatic, you should only modify `src`.

PR #203, opened 2015-11-30
small correction in binding class example

mackenza:
just a small change to be more precise about what is returned based on the example above.


yyx990803:
Thanks!

PR #223, opened 2016-01-11
Transition effects with dynamic components

maoosi:
Adding information about how to apply transition effects with dynamic components without creating any fragment instance.


yyx990803:
I'm not sure if this is correct. Fragment instance warnings is all about the component's template - if all dynamic components conform to the single-root-node rule then it doesn't matter where you place the `transition` attribute.

maoosi:
Mistake is on me, sorry about that. I was facing a problem with a component fragment instance since few days and the only way for me to fix it was applying the method described in my pull request.

I guess I have been a little too eager to share it with the community, because since your comment I've been running several independent tests in order to verify what I was saying, and I have not been able to re-create the issue I was facing.

For any reason it was coming from the cache that Google Chrome applies by default for Javascript files, and this put me in error. Anyway, I guess my pull request can be closed. Again, sorry about that.

PR #238, opened 2016-02-21
Fix component example

gimenete:


yyx990803:
Thanks, but the demo code is correct. We are creating a root instance that serves as an entry point, which contains the registered child component.

PR #240, opened 2016-02-22
add 'What are Components' section

hootlex:
Added new section that describes Components

### DIFF
--- a/src/guide/components.md
+++ b/src/guide/components.md
@@ -5,8 +5,9 @@ order: 12
 ---
 
 ##What are Components?
-Components are one of the most powerful features of Vue.js, helping you extend basic HTML elements and create reusable code.
-At a high level, Components are markers on a DOM element that tell Vue.js compiler to attach a specified behavior to that DOM element, or even to transform the DOM element and its children.
+Components are one of the most powerful features of Vue.js, helping you extend basic HTML elements and create reusable code. At a high level, Components are custom elements that Vue.js’ compiler attaches a specified behavior, or even transforms the elements and their children.
+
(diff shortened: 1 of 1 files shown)


src/guide/components.md
```diff
@@ -4,6 +4,10 @@ type: guide
 order: 12
 ---
 
+##What are Components?
+Components are one of the most powerful features of Vue.js, helping you extend basic HTML elements and create reusable code.
+At a high level, Components are markers on a DOM element that tell Vue.js compiler to attach a specified behavior to that DOM element, or even to transform the DOM element and its children.
```
yyx990803: I think it's better to use "custom elements" instead of "markers on a DOM element" here - the latter sounds like directive instead of components.

hootlex: "custom elements" sounds better but I gone with "markers on DOM element" for cases like `<tr is="my-component"></tr>`

Maybe "custom elements or special attributes" is better?

yyx990803: Yeah, that sounds good

hootlex:
I think I like it better now.

src/guide/components.md
```diff
@@ -4,6 +4,11 @@ type: guide
 order: 12
 ---
 
+##What are Components?
+Components are one of the most powerful features of Vue.js, helping you extend basic HTML elements and create reusable code. At a high level, Components are custom elements that Vue.js’ compiler attaches a specified behavior, or even transforms the elements and their children.
```
yyx990803: minor grammar issue: probably should be "Components are custom elements that Vue.js’ compiler would attach specified behavior to"

hootlex:
I changed the second sentence.

PR #254, opened 2016-03-22
documented the trick for mixing up two syntax styles of :class

fullfs:
Based on the discussion https://github.com/vuejs/vue/issues/2530
And the example https://jsfiddle.net/rhyzx/xzm80mc5/


yyx990803:
Hmmm I don't really recommend this style. I think we should just support objects inside arrays:

``` html
<div :class="[a, { b: b, c: c }]">
```

PR #270, opened 2016-03-31
added activate hook into API hooks list

fullfs:
To my point of view the hook is missing in the hooks list at API page. Fixed that :)

### DIFF
--- a/src/api/index.md
+++ b/src/api/index.md
@@ -658,6 +658,16 @@ type: api
 
 - **See also:** [Lifecycle Diagram](/guide/instance.html#Lifecycle-Diagram)
 
+### activate
+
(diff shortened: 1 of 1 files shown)


fullfs:
@yyx990803 ping :)

yyx990803:
I think this needs a bit more explanation because the hook is special in that it expects a callback and has async implications. Better to explain that right here in the API reference (and with an example).

posva:
ping @fullfs 
Any progress on this?

chrisvfritz:
Closing due to inactivity, but just opened [a new PR](https://github.com/vuejs/vuejs.org/pull/381) to re-attempt this change.

PR #287, opened 2016-04-19
Amend doc for new currency argument

phanan:
To follow up with https://github.com/vuejs/vue/pull/2676.

### DIFF
--- a/src/api/index.md
+++ b/src/api/index.md
@@ -1897,6 +1897,7 @@ type: api
 
 - **Arguments:**
   - `{String} [symbol] - default: '$'`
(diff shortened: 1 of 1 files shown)


ludo237:
Is should use the `£` sign to keep it coherent with the previous example

phanan:
Actually the `₫` sign should better demonstrate the purpose of the argument, as VND doesn't have any decimal places.

yyx990803:
I assume most users wouldn't know that symbol stands for VND, and they wouldn't know that VND uses no decimal places, so maybe add a bit of note on that ;)

phanan:
Done.

PR #289, opened 2016-04-21
Added twoWay binding note to compoent.props

cseufert:


src/api/index.md
```diff
@@ -424,7 +424,9 @@ type: api
       // type check plus other validations
```
kazupon: I think that this comment need to update.
e.g.
```js
// type check, other validations and prop syncing
```

ludo237:
Any update about this @cseufert ?

PR #296, opened 2016-05-05
Modify #ad width

abouthiroppy:
hello :)

I modified `#ad` width because `#ad` was overflowed at mobile screen.

## before
scroll-test

## check browsers
- safari (iPhone 6s plus iOS 9.3)
- chrome (nexus 5x)


yyx990803:
Thanks for the PR - however it seems the fix alone isn't enough so I had to apply a different fix. Still, thanks a lot!

PR #297, opened 2016-05-05
Add a note about frozen objects and v-for

sirlancelot:
Part of vuejs/vue#2809

### DIFF
--- a/src/guide/list.md
+++ b/src/guide/list.md
@@ -223,6 +223,10 @@ You can just do:
 this.items.$remove(item)
 ```
 
+#### Using `Object.freeze()`
+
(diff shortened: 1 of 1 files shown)


src/guide/list.md
```diff
@@ -223,6 +223,10 @@ You can just do:
 this.items.$remove(item)
 ```
 
+#### `Object.freeze()`
+
+When iterating over an array of objects frozen with `Object.freeze()`, you need to explicitly use a `track-by` key. A warning will be displayed in this scenario when Vue.js is unable to track objects automatically
```
yyx990803: period at the end please? ;)

sirlancelot:
Fixed ;)

PR #299, opened 2016-05-10
Add a note about v-else caveat with components and v-show

posva:
Refers to https://github.com/vuejs/vue/issues/2639


posva:
Should I add a not to the API too?

yyx990803:
Just a link in the API should be enough

posva:
I noticed there's already a _See also_ link in the api that refers to the `v-else` guide. Nevertheless I added a second one because it looks more explicit to me and draws user attention because of the word caveat

PR #309, opened 2016-05-26
update style of #ad

ZoomZhao:
avoid horizontal scroll on mobile web


chrisvfritz:
I wasn't able to reproduce this problem, so I think it may be fixed now. Closing for now.

PR #312, opened 2016-06-07
Fixed `is` attribute in the dynamic component example

sebastiandedeyne:


sebastiandedeyne:
Just had an a-ha moment and realised I'm the one that's wrong, sorry!

PR #324, opened 2016-06-20
minor grammar fix

kas:
Hi Evan,

Thank you for creating the framework! I have been using it recently and it is amazing compared to some of the other more clunky front end frameworks available.

I was reading the guide and saw a sentence that I thought should be reworded. I made the minor change in this commit.

I appreciate you reviewing this pull request!

### DIFF
--- a/src/guide/comparison.md
+++ b/src/guide/comparison.md
@@ -8,7 +8,7 @@ order: 19
 
 There are a few reasons to use Vue over Angular, although they might not apply for everyone:
 
-- Vue.js is much simpler than Angular, both in terms of API and design. You can learn almost everything about it really fast and get productive.
+- Vue.js is much simpler than Angular, both in terms of API and design. You can learn everything about it quickly and become more productive.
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks for the contribution @kas! Unfortunately, I think this might be a bit more than a grammar fix. To me, it changes the meaning with the:

- removal `almost` to make an even bolder claim
- addition of `more` to possibly imply that Vue devs are more productive than Angular devs, which may be the case but is probably not a productive poitn to make here

If you update to `You can learn almost everything about it quickly and become productive` or `You can learn everything to become productive quickly` then I'd be happy to merge. 😃

kas:
Hey @chrisvfritz,
I made changes like you requested. Although I don't necessarily think that `more` implies that Vue devs are more productive -- I am noticing Vue is improving my workflow more so than Angular 😆. 

I appreciate you looking into the pull request, please let me know what you think!

PR #334, opened 2016-07-14
Added a tip for local-components

LukasKnuth:
Small clarification on the `el`-element, which I was previously using wrong. See https://github.com/vuejs/vue/issues/3169


chrisvfritz:
@LukasKnuth Thanks for bringing this up! I just [included a note](https://github.com/vuejs/vuejs.org/commit/9e6c1d056b9ee4b44b5d7b7026d2a96d63a1111b) further up about this. 😃

PR #349, opened 2016-08-02
Rendering problems

RioloGiuseppe:
Rendering problems found using vuejs in an aspx page.


chrisvfritz:
@RioloGiuseppe I'll change `"valid"` in this section to `normal`, since these shorthand attributes (starting with `:` and `@`) are technically valid in attribute names. From the [HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2):

> Attribute names must consist of one or more characters other than the space characters, U+0000 NULL, U+0022 QUOTATION MARK ("), U+0027 APOSTROPHE ('), U+003E GREATER-THAN SIGN (>), U+002F SOLIDUS (/), and U+003D EQUALS SIGN (=) characters, the control characters, and any characters that are not defined by Unicode.

That means if a server-side renderer isn't correctly parsing these, it's actually a bug in the renderer.

PR #354, opened 2016-08-05
2.0

ssalcio:
alla linea 5030 aggiungere
var contenteditable = {

  	bind: function bind()
  	{
  		var self = this;
  		var el = this.el;
  		var lazy = this.params.lazy;
  		var debounce = this.params.debounce;

  		// handle composition events.
  		//   http://blog.evanyou.me/2014/01/03/composition-event/
  		// skip this for Android because it handles composition
  		// events quite differently. Android doesn't trigger
  		// composition events for language input methods e.g.
  		// Chinese, but instead triggers them for spelling
  		// suggestions... (see Discussion/#162)
  		var composing = false;
  		if (!isAndroid)
  		{
  			this.on('compositionstart', function ()
  			{
  				composing = true;
  			});
  			this.on('compositionend', function ()
  			{
  				composing = false;
  				// in IE11 the "compositionend" event fires AFTER
  				// the "input" event, so the input handler is blocked
  				// at the end... have to call it here.
  				//
  				// #1327: in lazy mode this is unecessary.
  				if (!lazy)
  				{
  					self.listener();
  				}
  			});
  		}

  		// prevent messing with the input when user is typing,
  		// and force update on blur.
  		this.focused = false;
  		if (!lazy)
  		{
  			this.on('focus', function ()
  			{
  				self.focused = true;
  			});
  			this.on('blur', function ()
  			{
  				self.focused = false;
  				// do not sync value after fragment removal (#2017)
  				if (!self._frag || self._frag.inserted)
  				{
  					self.rawListener();
  				}
  			});
  		}

  		// Now attach the main listener
  		this.listener = this.rawListener = function ()
  		{
  			if (composing || !self._bound)
  			{
  				return;
  			}
  			var val = el.innerHTML;
  			
  			self.set(val);
  			// force update on next tick to avoid lock & same value
  			// also only update when user is not typing
  			nextTick(function ()
  			{
  				if (self._bound && !self.focused)
  				{
  					self.update(self._watcher.value);
  				}
  			});
  		};

  		// apply debounce
  		if (debounce)
  		{
  			this.listener = _debounce(this.listener, debounce);
  		}

  		// Support jQuery events, since jQuery.trigger() doesn't
  		// trigger native events in some cases and some plugins
  		// rely on $.trigger()
  		//
  		// We want to make sure if a listener is attached using
  		// jQuery, it is also removed with jQuery, that's why
  		// we do the check for each directive instance and
  		// store that check result on itself. This also allows
  		// easier test coverage control by unsetting the global
  		// jQuery variable in tests.
  		this.hasjQuery = typeof jQuery === 'function';
  		if (this.hasjQuery)
  		{
  			var method = jQuery.fn.on ? 'on' : 'bind';
  			jQuery(el)[method]('change', this.rawListener);
  			if (!lazy)
  			{
  				jQuery(el)[method]('input', this.listener);
  			}
  		} else
  		{
  			this.on('change', this.rawListener);
  			if (!lazy)
  			{
  				this.on('input', this.listener);
  			}
  		}

  		// IE9 doesn't fire input event on backspace/del/cut
  		if (!lazy && isIE9)
  		{
  			this.on('cut', function ()
  			{
  				nextTick(self.listener);
  			});
  			this.on('keyup', function (e)
  			{
  				if (e.keyCode === 46 || e.keyCode === 8)
  				{
  					self.listener();
  				}
  			});
  		}

  		//// set initial value if present
  		//if (el.hasAttribute('value') || el.tagName === 'TEXTAREA' && el.value.trim())
  		//{
  		//	this.afterBind = this.listener;
  		//}
  	},

  	update: function update(value)
  	{
  		// #3029 only update when the value changes. This prevent
  		// browsers from overwriting values like selectionStart
  		value = _toString(value);
  		if (value !== this.el.innerHTML) this.el.innerHTML = value;
  	},

  	unbind: function unbind()
  	{
  		var el = this.el;
  		if (this.hasjQuery)
  		{
  			var method = jQuery.fn.off ? 'off' : 'unbind';
  			jQuery(el)[method]('change', this.listener);
  			jQuery(el)[method]('input', this.listener);
  		}
  	}
  };

replace:
var handlers = {
    text: text$2,
    radio: radio,
    select: select,
    checkbox: checkbox
  };
with:
var handlers = {
    text: text$2,
    radio: radio,
    select: select,
    checkbox: checkbox,
    contenteditable: contenteditable
  };

nella routine linea 5213
bind: function bind() {

replace:

if (tag === 'INPUT') {
        handler = handlers[el.type] || handlers.text;
      } else if (tag === 'SELECT') {
        handler = handlers.select;
      } else if (tag === 'TEXTAREA') {
      	handler = handlers.text;
      } else {
        'development' !== 'production' && warn('v-model does not support element type: ' + tag, this.vm);
        return;
      }

with:

if (tag === 'INPUT') {
        handler = handlers[el.type] || handlers.text;
      } else if (tag === 'SELECT') {
        handler = handlers.select;
      } else if (tag === 'TEXTAREA') {
      	handler = handlers.text;
      } else if (el.getAttribute("contenteditable") == "true") {
      	handler = handlers.contenteditable;
      } else {
        'development' !== 'production' && warn('v-model does not support element type: ' + tag, this.vm);
        return;
      }


chrisvfritz:
@ssalcio It looks like this pull request is requesting a merge from 2.0 into master, which we are not ready for yet - though it looks like you have other code you are looking to change. Please see [this document](https://gist.github.com/Chaser324/ce0505fbed06b947d962) for the process to submit a pull request on GitHub.

PR #357, opened 2016-08-07
vueify@next needs global envify

zigomir:
With browserify & Vue 2.0 you actually need to install `envify` yourself and apply it as global transform like [described here](https://github.com/vuejs/vueify/tree/next#building-for-production-1).

And there's also extract-css plugin included now so browserify users can extract styles into file as well.


chrisvfritz:
@zigomir Thanks for this!

PR #363, opened 2016-08-12
fix appropriateListComponent function syntax

vinicius73:


chrisvfritz:
@vinicius73 Thanks for this! I just implemented this change separately so that the style also complies with ESLint Standard.

PR #371, opened 2016-08-15
Use if statement instead of ternary

gaearon:
I think this is more consistent with how people use React in practice.

React is JS, and it is expected that you use ternaries in the same case as you would use them in JS: only if they help readability. In this case they seem to obscure it.

You still get to say that React is verbose after this change ;-)

### DIFF
--- a/src/guide/comparison.md
+++ b/src/guide/comparison.md
@@ -56,14 +56,12 @@ Take this example below:
 
 ``` jsx
 render() {
-  const { items } = this.props
+  let { items } = this.props
 
   let children
   if (items.length > 0) {
     children = items.map(item =>
-      <li key={item.id}>
-        {item.name}
-      </li>
+      <li key={item.id}>{item.name}</li>
     )
   } else {
     children = <p>No items found.</p>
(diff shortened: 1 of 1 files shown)


src/guide/comparison.md
```diff
@@ -55,17 +55,23 @@ With React's JSX, there are minor differences from HTML, such as using `classNam
...
+render() {
   const { items } = this.props
 
+  let children
+  if (items.length > 0) {
+    children = items.map(item =>
+      <li key={item.id}>
+        {item.name}
+      </li>
+    )
```
kasperpeulen: closing parens without a beginning :)

gaearon: Typed it on iPhone so maybe I missed something but this paren should correspond to opening `items.map(`.

kasperpeulen: oh wait nevermind haha, you are right

chrisvfritz:
Thanks @gaearon! I just merged and will update the surrounding text accordingly, but I do have one question: Why update `const { items } = this.props` to use `let` instead? It doesn't look like it's being overwritten later on.

gaearon:
People reading this might not have good knowledge of ES6 and might assume React forces them to use "const" or something.

chrisvfritz:
Got it. Thanks for the explanation!

PR #376, opened 2016-08-20
Update JSFiddle Hello World example link

amirrustam:


src/guide/index.md
```diff
@@ -6,7 +6,7 @@ order: 1
 
 Let's start with a quick tour of Vue's data binding features. If you are more interested in a high-level overview first, check out this [blog post](!!TODO: HIGH LEVEL OVERVIEW FOR RELEASED 2.0). If you are already familiar with Vue 1.0, the [migration guide](!!TODO: MIGRATION GUIDE FROM 1.0 to 2.0) may be a good place to start.
 
-The easiest way to try out Vue.js is using the [JSFiddle Hello World example](https://jsfiddle.net/chrisvfritz/ng41aepx/). Feel free to open it in another tab and follow along as we go through some basic examples. If you prefer downloading / installing from a package manager, check out the [Installation](/guide/installation.html) page.
+The easiest way to try out Vue.js is using the [JSFiddle Hello World example](https://jsfiddle.net/yyx990803/okv0rgrk/). Feel free to open it in another tab and follow along as we go through some basic examples. If you prefer downloading / installing from a package manager, check out the [Installation](/guide/installation.html) page.
```
LinusBorg: The script tag in this fiddle points to the latest version on npm, which is `v1.0.26`until `v2.0` becomes officially released.

So this should not be added to the docs until then.

LinusBorg:
The script tag in this fiddle points to the latest version on npm, which is v1.0.26until v2.0 becomes officially released.

So this should not be added to the docs until then.

chrisvfritz:
Also, we may want to update that example when we update, as the 4-space indentation doesn't comply with the ferross/standard style we're trying to use in examples.

src/guide/index.md
chrisvfritz: Also, we may want to update that example when we update, as the 4-space indentation doesn't comply with the ferross/standard style we're trying to use in examples.

chrisvfritz:
I'm going to close this for now - we'll make a pass to update versions used in all fiddles when we go gold.

PR #390, opened 2016-09-01
Attributes  Mustaches cannot be used inside HTML attributes 

zerofront:
Attributes  Mustaches cannot be used inside HTML attributes, instead use a v-bind directive


chrisvfritz:
Thanks for this! We just made a [bunch of updates](https://github.com/vuejs/vuejs.org/commit/8732e7c2ed54270c29d07a04577b24e8afbc4a93) to this page that should resolve the issue.

PR #393, opened 2016-09-03
Remove the description unsuited for Vue 2 any more

BigPage:
Removed content:
" You may have noticed this achieves the same result as an attribute interpolation using href="{{url}}": that is correct, and in fact, attribute interpolations are translated into v-bind bindings internally."

This content in contradiction with "Mustaches cannot be used inside HTML attributes, instead use a v-bind directive.".

### DIFF
--- a/src/guide/syntax.md
+++ b/src/guide/syntax.md
@@ -116,7 +116,7 @@ Some directives can take an "argument", denoted by a colon after the directive n
 <a v-bind:href="url"></a>
 ```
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Those are the docs for Vue 1. The soon-to-be-published Vue 2 docs incude [updated language](https://github.com/vuejs/vuejs.org/blob/2.0/src/guide/syntax.md).

PR #409, opened 2016-09-20
Fix a warning and a bad reference in the "Components - Dynamic Components" example

jvanbrug:
http://rc.vuejs.org/guide/components.html#Dynamic-Components


LinusBorg:
Good catch!

PR #412, opened 2016-09-22
Fixed "type: string" to "type: String"

dodas:


dodas:
Actually, it should probably be Number, as validator and default value are for number.

PR #413, opened 2016-09-22
Fixed advanced prop definition

dodas:
Changed type to Number, as both default and validator are for number and changed default value to pass validator.


LinusBorg:
Good Catch! LGTM

PR #415, opened 2016-09-25
add $slots property to API docs

kazupon:
I realized that $slots property is nothing in `guide/api/index.md`, when I found that dead link in [render function](http://rc.vuejs.org/guide/render-function.html).

### DIFF
--- a/src/api/index.md
+++ b/src/api/index.md
@@ -894,6 +894,51 @@ type: api
 
   The direct child components of the current instance. **Note there's no order guarantee for `$children`, and it is not reactive.** If you find yourself trying to use `$children` for data binding, consider using an Array and `v-for` to generate child components, and use the Array as the source of truth.
 
+### vm.$slots
+
+- **Type:** `Object`
+
+- **Read only**
+
+- **Details:**
+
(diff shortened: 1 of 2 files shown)


kazupon:
@chrisvfritz thanks!

chrisvfritz:
@kazupon Thanks for catching it!

jbruni:
Great! Just stumbled upon it. Cool that it has already been resolved.

jbruni:
Hmmm... the explanation and code samples seem confusing to me.

kazupon:
sorry, my poor english ... 😞

LinusBorg:
@jbruni Have you read the sections linked under the example? The content of `$slots` can only be of any use to you when you use render functions - in which case you will have to understand what vNodes are, which is explained in the linked guide sections.

jbruni:
@LinusBorg - Yes, I do understand the context, in ttheory and practice. The text could be much simpler, and the code (only perceived now) has a little error (reference to "slot2" and "slot1"), and in my opinion it does not help clarifying the contents. I'll see if I can quickly write and send a suggestion as PR.

PR #419, opened 2016-09-26
vm.$slots documentation improvement

jbruni:


jbruni:
@LinusBorg - here is my suggestion.

Thanks a lot to @kazupon who laid the foundations for this suggestion.

src/api/index.md
```diff
@@ -902,42 +902,38 @@ type: api
 
 - **Details:**
 
-  A hash of VNode children of component that should resolve with slot. VNode children resolved with Single Slot are stored as the `default` key. VNode children resolve with Named Slot are stored as the key that specified with `slot` attribute. those VNode children are stored an Array.
+  An object that holds virtual nodes containing parent contents used for content distribution (transclusion) inside [slots](/guide/components.html#Content-Distribution-with-Slots). The `default` property contains [single slot](/guide/components.html#Single-Slot) contents. Each [named slot](/guide/components.html#Named-Slots) has its own corresponding property, with the same name of the slot.
```
LinusBorg: May be because I'm not a native speaker, but I had to read the beginning twice. Maybe this is clearer:

"An object that holds virtual nodes which contain parent contents.."

jbruni: Ok. Pushed a commit with the change.

But I've made a mistake in the commit message... should I create another PR?  (With a single commit and correct message.) Or just let it go?

chrisvfritz: @jbruni No need to create another PR - we're going to squash before we merge anyway. I have a couple other changes I'd like to make, but I'll just push up those commits on your branch and then merge. 😃 👍

PR #420, opened 2016-09-28
fix comment in a documentation snippet

aoloe:


chrisvfritz:
This is not accurate. You really do have access to the entire Vue instance with `this`.

aoloe:
well, i must believe you!

but -- out of curiosity --  could you just tell me how? `this.name` gives me the content of `data: {name: "....."}`...

as an example, how can i access a function?

chrisvfritz:
You access a method with this.nameOfYourMethod(). In a created hook, try to console.log(this) to see everything that's available.

PR #427, opened 2016-10-01
some words correct

hijiangtao:
- In Conditionals and Loops chapter, the `second chapter` in content `This second example demonstrates that we...` may be moved from old version overview page directly? I think the example should be the third one in this page, rather than the second.
- In Composing with Components chapter, the `prop` in content `... to make it accept a prop:` should be `props`.

Thanks for your great effort in developing Vue.js, these are some really small mistakes I found when I read the Introduction page.

I choose master branch to commit the pull request because I don't know which branch should I choose from guidance. The link in the bottom of this page lead me to here.

### DIFF
--- a/src/guide/index.md
+++ b/src/guide/index.md
@@ -115,7 +115,7 @@ var app3 = new Vue({
 
 Go ahead and enter `app3.seen = false` in the console. You should see the message disappear.
 
-This second example demonstrates that we can bind data to not only text and attributes, but also the **structure** of the DOM. Moreover, Vue also provides a powerful transition effect system that can automatically apply [transition effects](transitions.html) when elements are inserted/updated/removed by Vue.
+This third example demonstrates that we can bind data to not only text and attributes, but also the **structure** of the DOM. Moreover, Vue also provides a powerful transition effect system that can automatically apply [transition effects](transitions.html) when elements are inserted/updated/removed by Vue.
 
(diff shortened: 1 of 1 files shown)


src/guide/index.md
```diff
@@ -115,7 +115,7 @@ var app3 = new Vue({
 
 Go ahead and enter `app3.seen = false` in the console. You should see the message disappear.
 
-This second example demonstrates that we can bind data to not only text and attributes, but also the **structure** of the DOM. Moreover, Vue also provides a powerful transition effect system that can automatically apply [transition effects](transitions.html) when elements are inserted/updated/removed by Vue.
+This third example demonstrates that we can bind data to not only text and attributes, but also the **structure** of the DOM. Moreover, Vue also provides a powerful transition effect system that can automatically apply [transition effects](transitions.html) when elements are inserted/updated/removed by Vue.
```
chrisvfritz: I decided to just remove the number in a separate commit, since it's not really important.

src/guide/index.md
```diff
@@ -262,7 +262,7 @@ Now you can compose it in another component's template:
 </ul>
 ```
 
-But this would render the same text for every todo, which is not super interesting. We should be able to pass data from the parent scope into child components. Let's modify the component definition to make it accept a [prop](/guide/components.html#Props):
+But this would render the same text for every todo, which is not super interesting. We should be able to pass data from the parent scope into child components. Let's modify the component definition to make it accept a [props](/guide/components.html#Props):
```
chrisvfritz: `prop` is correct, as we are adding only one.

PR #428, opened 2016-10-01
Improve responsive design on mobile devices

lin-lu:


chrisvfritz:
@lin-lu Can you please describe (and possibly print screenshots of) the problem you're seeing that you're trying to fix?

lin-lu:
The first problem is that when the screen width is between 720px and 900px, the hamburger menu button on the home page does not work. You can reproduce this bug by resizing the browser.

The second problem is that the search text box is displayed on top of the vue logo, when the width is around 800px. 
screen shot 2016-10-01 at 17 37 25
I changed the breakpoint of the media queries from 720px to 900px, so the mobile version will be displayed when the width is smaller than 900px.

After I changed that, I noticed that the max-width of the content is fixed at 600px, so when the screen width is larger than that and smaller than 900px, the content will appear on the left side of the page, leaving some empty space on the right side.
screen shot 2016-10-01 at 17 57 12
To make it symmetry, I also centered the content for mobile devices.

chrisvfritz:
Thanks! This looks good. 👍

PR #448, opened 2016-10-03
add warnings about using arrow functions on instance properties

chrisvfritz:
@LinusBorg What do you think?


qwexvf:
Thanks for this. i was like wtf? because of that. in some cases you can't really use arrow keys.

PR #454, opened 2016-10-04
Browserify users can use hint in package.json for standalone build

woodcoder:
See https://github.com/defunctzombie/package-browser-field-spec

### DIFF
--- a/src/guide/installation.md
+++ b/src/guide/installation.md
@@ -61,7 +61,13 @@ resolve: {
 }
 ```
 
-For Browserify, you can use [aliasify](https://github.com/benbria/aliasify) to achieve the same.
+For Browserify, add the following to your module's `package.json`:
+
+``` js
+  "browser": {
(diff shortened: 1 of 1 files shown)


src/guide/installation.md
```diff
@@ -61,7 +61,13 @@ resolve: {
 }
 ```
 
-For Browserify, you can use [aliasify](https://github.com/benbria/aliasify) to achieve the same.
+For Browserify, add the following to your module's `package.json`:
+
+``` js
+  "browser": {
+      "vue": "vue/dist/vue.js"
```
woodcoder: I realised that I'd seen a similar thing working before, but the browser field in that case was on the equivalent of the vue.js module itself.  That would then work if you wanted to default to the standalone on the client and the runtime on the server -- but I can understand that the preferred option is to default to the runtime by default in both cases!

Thanks for all the great work -- it's really good documentation.

PR #457, opened 2016-10-04
falsy -> false

yorts52:


simplesmiler:
"Falsy" is a common term for values, that JavaScript treats as false: `false`, `0`, `''`, `undefined`, `null`.

PR #461, opened 2016-10-04
"v-on:click" needs "native" modifier

rstoenescu:


chrisvfritz:
The `.native` modifier is [only necessary on components](http://vuejs.org/guide/components.html#Binding-Native-Events-to-Components), as native elements only support native events.

rstoenescu:
Just wanted to drop the PR as I just figured that out :)

PR #468, opened 2016-10-05
Update list.md to correctly reflect param order

scottsandersdev:
Docs currently show that index is second param, but it should be the first param, at least in vue 1.


LinusBorg:
The prams order switched: http://vuejs.org/guide/migration.html#v-for-Argument-Order-for-Arrays

PR #478, opened 2016-10-06
Update mixins.md (Vuex version)

jbruni:


jbruni:
Maybe we could specify that the example applies for Vuex version 1.x

Issue #475, opened 2016-10-06
[Suggestion] Make the search bar accessible from anywhere on the page


SoullessWaffle:
You guys have really user-friendly documentation and search, but at the moment the user needs to scroll to the top of the page every time they want to search for something.
I believe it would be a meaningful UX improvement if the user could search from anywhere on the page.
Making the header fixed so it always stays in view is one possible solution.

chrisvfritz:
I think this could be a good change. 👍  I may not have time to implement it right away, but a PR would be welcome and I can check it out.

LinusBorg:
This is live now :+1:

PR #487, opened 2016-10-08
Added some css changes to fix navigation for search quick access

Rubemlrm:
Possible fix for suggestion #475 , now navbar is fixed for search quick access and quick access to other stuff on nav.


phanan:
Bad guy here. If we can reduce the height of the bar, that would be great. It's taking too much space if you ask me.

themes/vue/source/js/common.js
```diff
@@ -124,28 +124,19 @@
...
+        offset: 0
       })
     }
 
     // listen for scroll event to do positioning & highlights
     window.addEventListener('scroll', updateSidebar)
     window.addEventListener('resize', updateSidebar)
 
-    function updateSidebar () {
+    function updateSidebar (event) {
```
posva: We can remove the argument

chrisvfritz:
@phanan We have the technology. 😄  What were you thinking? It's at 25px vertical padding right now. I agree it may be a bit large for a fixed header right now, but I also don't want to make it too cramped.

themes/vue/source/js/common.js
chrisvfritz: Just removed this.

chrisvfritz:
@phanan I just updated it to 20px. You can let me know what you think. It's also now possible to just change [this variable](https://github.com/vuejs/vuejs.org/pull/487/files#diff-43c2044b37446102035d286329cd5b8bR22) to play around with that value.

phanan:
@chrisvfritz IMHO 10px looks just nice:

image

But I guess what I really mean is, the new padding should only apply when the menu detaches into the `fixed` position, maybe with a smooth transition.

chrisvfritz:
Yeah... I was afraid you'd say that, but I agree. I'll try to update tomorrow.

Issue #285, opened 2016-04-18
Better add some placeholders texts to firebase example.


qiansen1386:
on page https://github.com/vuejs/vuejs.org/blob/master/src/examples/firebase.md

From
```html
    <input v-model="newUser.name">
    <input v-model="newUser.email">
```
to
```html
    <input v-model="newUser.name" placeholder="Username">
    <input v-model="newUser.email" placeholder="demo@email.com">
```
Like this: https://jsfiddle.net/qiansen1386/2d3htmpr/89/

PR #502, opened 2016-10-12
Updated firebase example inputs

Rubemlrm:
Updated inputs with placeholders and missing input type from jsfiddle . this is relative to issue #285


chrisvfritz:
Thanks! I had to fork the fiddle so that it's still owned by a member of the core team, but I just edited it in your branch so that you still get commit credit.

PR #506, opened 2016-10-13
Image path Update

andela-iadeniyi:
Update the image path in the hackernews.md documentation

### DIFF
--- a/src/guide/installation.md
+++ b/src/guide/installation.md
@@ -56,7 +56,7 @@ By default, the NPM package exports the **runtime-only** build. To use the stand
 ``` js
 resolve: {
   alias: {
-    vue: 'vue/dist/vue.js'
+    'vue$': 'vue/dist/vue.js'
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


LinusBorg:
I don't think it's a good idea to add absolute link to the repository. this link would persist through all forks of this repo, and all forks would break if we wver thange our repo's content.

andela-iadeniyi:
Yes, the images doesn't display. I get your point, I'll have to make the path relative.

chrisvfritz:
Images seem to display for me. Are you looking at [http://vuejs.org/examples/hackernews.html](http://vuejs.org/examples/hackernews.html)?

andela-iadeniyi:
I'm looking at https://github.com/vuejs/vuejs.org/blob/master/src/examples/hackernews.md.

chrisvfritz:
Ah, I see. The docs aren't actually made to be read from GitHub. They only render to HTML because that's what GitHub does with Markdown files.

PR #509, opened 2016-10-15
update to vue 2.0.3

defcc:


chrisvfritz:
@defcc Could you also please deploy this if not already done?

PR #522, opened 2016-10-19
Complete the doc on custom events

AlexandreBonneau:
When reading the documentation about event handling, one can assume `vm.$emit` is an alias to the DOM `dispatchEvent` method, which is not if I read the source correctly.

Hence this can lead to misconception about how to use those event methods.

This is a very first try to explain this in the doc, but surely it needs to be rephrased by a native english speaker as well as be corrected if I'm wrong, so please correct me!


chrisvfritz:
Thanks for prompting us to clarify this. I just updated the guide to be a bit clearer. I did not add anything to the Event Handling guide however, as `v-on` is capable of listening for both native DOM events and those created by Vue's event system.

PR #524, opened 2016-10-20
update wrapper width to 120px

ruiming:
I think the wrapper is a little crowded, and my chrome can not display it normally. So I change it from 110px to 120px.
image

### DIFF
--- a/themes/vue/layout/partials/sidebar.ejs
+++ b/themes/vue/layout/partials/sidebar.ejs
@@ -4,7 +4,7 @@
     </ul>
     <div class="list">
         <div id="donate" style="margin-bottom: 20px">
-            <span class="wrapper" style="width: 110px">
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Could you please tell me more about your system or the conditions producing this problem? The only way I'm able to reproduce this in Chrome so far is by zooming out to 75%.

ruiming:
This problem happens in my chrome without any zooming. My system is Arch Linux and my chrome version is  54.0.2840.59 (64-bit). 
I just found that this may happened because of my fonts settings of the chrome. I try to change it to other fonts but the result is that some fonts will makes it display normally but some fonts will not. I am not sure whether others has the same problems with me, but I still think is a little crowded and maybe try a not width-fixed way or increase the width to avoid it.

chrisvfritz:
I improved the CSS so that the width is now dynamic.

PR #526, opened 2016-10-20
Update index.md

emanuelgaspar:
I tried to follow the guide and every example worked, except for the components. Upon further investigation, I found the issue: the component must be registered before the Vue instance. This was not clear in the Introductory guide, so I suggest it's added ( I copied the same sentence as seen on https://vuejs.org/guide/components.html )


chrisvfritz:
I want to avoid going too deep into the component system and component registration on this introductory page, so I just included the component registration in the final example so that readers don't have to figure out where it goes.

PR #530, opened 2016-10-22
more idiomatic JSX code

mgutz:
The preceding comment "vetted by the React community" implies, at least to me, idiomatic JSX code.


chrisvfritz:
It sounds like it doesn't match your personal style, but the example was [provided by a member of the React team](https://github.com/vuejs/vuejs.org/pull/371), so it's definitely "idiomatic" JSX code.

PR #539, opened 2016-10-24
Add _target="blank" deprecation on `<router-link>`

fnlctrl:


src/guide/migration-vue-router.md
```diff
@@ -267,6 +267,8 @@ You'll need to update it like this:
 <router-link to="/about">About</router-link>
 ```
 
+Note: `target="_blank"` is not supported on `<router-link>` component, please use `<a>` tags instead.
```
hogart: I think it would be nice to mention about `router.match()` here, so people won't get impression that manually maintaining URLs is recommended way.

PR #541, opened 2016-10-25
"Webpack" term capitalization + runtime-only build size extracted to a variable

gbezyuk:
Subj. Also, I have no idea why eb3aef9 still shows up as if wasn't merged o_O
UPD: found another minor typo, also fixed (see 425055f)


chrisvfritz:
It may sometimes take up to a day before a merge becomes live, as I sometimes like to continue tweaking things after they're merged, then will merge when I'm done. eb3aef9 does look live to me though. Maybe a caching issue?

PR #542, opened 2016-10-25
Change the background color of code element when used in a tip

AlexandreBonneau:
When using a `code` in a `tip`, both background are of the same color, which leads to weird looking padding.

This fixes this by changing the code background color slightly.
Before :
Before

After :
After


chrisvfritz:
Looks good! Thanks for this.

AlexandreBonneau:
Hehe, my bad, I know SCSS, not Stylus ;)

PR #555, opened 2016-10-30
Rephrase info about there being no controllers.

Francisc:
"Some users" is too impersonal and it's safe to say many people who read the guide are not yet users.


Francisc:
Thank you.

PR #557, opened 2016-10-30
Minor grammatical fix.

Francisc:


Francisc:
Not sure why the diff looks so bad, I only changed one word:

"computed properties are cached based on *its* dependencies" => "computed properties are cached based on **their** dependencies"

chrisvfritz:
Thanks! GitHub's had a few quirks lately - I don't think you did anything wrong. 🙂

Francisc:
Thank you.

PR #560, opened 2016-10-31
Update list.md

Dawlatzai:
Replace `item` with `index` in `v-for(item, index)` because it outputs the following.

- Parent- [object Object] -

- Parent- [object Object] - 

Changing it to this form `v-for(index, item)` outputs the correct result.

### DIFF
--- a/src/guide/list.md
+++ b/src/guide/list.md
@@ -60,7 +60,7 @@ Inside `v-for` blocks we have full access to parent scope properties. `v-for` al
 
 ``` html
 <ul id="example-2">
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Are you seeing this rendered result on the website? It renders correctly for me. My guess is you may be reading the Vue 2 docs while externally testing the examples with Vue 1. Could that be the case?

Dawlatzai:
Yes it is the case I was test it in Vue 1 and in Vue 2 it works as it is.

chrisvfritz:
OK, I'll close then.

PR #567, opened 2016-11-04
Update components.md

DominikSerafin:
It was confusing for me in the beginning where and when to pass props as initial data so I think it might be confusing for others too.

### DIFF
--- a/src/guide/components.md
+++ b/src/guide/components.md
@@ -357,7 +357,7 @@ There are usually two cases where it's tempting to mutate a prop:
 
 The proper answer to these use cases are:
 
-1. Define a local data property that uses the prop's initial value as its initial value;
+1. Define a local data property (e.g. inside `created` lifecycle hook) that uses the prop's initial value as its initial value;
 
 2. Define a computed property that is computed from the prop's value.
 


chrisvfritz:
That actually wasn't quite what we meant, so it's good you're bringing this up. 😄  I just added [a couple examples](https://vuejs.org/guide/components.html#One-Way-Data-Flow) and would love to hear your thoughts.

DominikSerafin:
@chrisvfritz I wasn't sure 100% that's the correct way to do it (I've even asked a [question about it on SO](http://stackoverflow.com/questions/40408096/whats-the-correct-way-to-pass-props-as-initial-data-in-vue-js-2)) So I was kinda hoping for two things with this pull request:

1. Either my solution will be correct one and the pull will be merged;
2. Or someone like you will correct me and shows the good way to do it and include it in the docs (and it happened so thank you!)

And that method of directly passing the props to the data is 💯. I don't have much feedback about it because I think that's perfect (the method and how it's described in docs now).

chrisvfritz:
Fantastic! Thanks for helping make the docs a little better for everyone that comes after you. 👏🏻

samlaf:
@chrisvfritz How can we deal with the case when the data we are passing down to the child component is gotten from a database asynchronously during the create() phase of the parent, and is still not available during the create() phase of the child?

sonnt189:
@samlaf well, this case you can use `watcher` with `immedieate:true` flag to get props to assign new state in child components

PR #568, opened 2016-11-04
_common.styl: unify #donate a font-family

JLHwung:
Unify font-family setting of `#donate a` with `$body-font`.
Previously the font-family of `#donate a` is hard-coded and does not consist with `$body-font` and this PR solves this discrepancy.

Edit: typo

### DIFF
--- a/themes/vue/source/css/_common.styl
+++ b/themes/vue/source/css/_common.styl
@@ -198,7 +198,9 @@ a.button
         width 14px
         height 14px
     a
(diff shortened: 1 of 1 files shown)


chrisvfritz:
@JLHwung Can you clarify what problem you're solving here?

JLHwung:
@chrisvfritz Description updated.

chrisvfritz:
Thanks. I made a couple more tweaks. The `11px/14px` style is short for `font-size: 11px` and `line-height: 14px` by the way.

PR #574, opened 2016-11-08
components' documentation update base on issue#4117

MarcoBoffo:
https://github.com/vuejs/vue/issues/4117

I think this resource would be useful for single file components' users.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -1020,6 +1020,28 @@ template: '<div><stack-overflow></stack-overflow></div>'
 
 A component like the above will result in a "max stack size exceeded" error, so make sure recursive invocation is conditional (i.e. uses a `v-if` that will eventually be `false`).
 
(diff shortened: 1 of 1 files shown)


posva:
Thanks for that! It may indeed help some people, however, the important part is the necessity of lazy loading. This can be achieved in many different ways and depends on the bundler. Also, I think the examples would be clearer if you added the require/import parts for both MyComponent and StackOverflow

MarcoBoffo:
Totally agree with you, tell me if you need anything else.

chrisvfritz:
@MarcoBoffo Thanks for this. Just wanted to let you know that it hasn't fallen off my radar. I'll need to rework this slightly to conform to our internal style guide, but should find time in the next couple days.

yyx990803:
Also - I don't think "recursive" is the correct term here, rather it's about circular references between two components.

PR #580, opened 2016-11-14
Update comparison.md

fengsi:


chrisvfritz:
Both versions are acceptable - no need to change.

PR #588, opened 2016-11-16
a note on incremental adoptability in Vue vs React comparison.md

gbezyuk:
I believe we should put a little more attention on the possibilities provided by Vue in terms of incremental adoptability. So here I make a humble step in this direction.


yyx990803:
Thanks for the PR, but in terms of incremental adoptability and migrating in a piecemeal fashion, React is technically in the same camp with Vue. (although you have to make more decisions in picking the additional pieces)

PR #589, opened 2016-11-17
Ecosystem Nav

lukepolo:
Its annoying to switch from different docs cause I have to google it each time


themes/vue/layout/partials/ecosystem_dropdown.ejs
```diff
@@ -0,0 +1,7 @@
+<li class="nav-dropdown-container">
+  <a class="nav-link">Echosystem</a><span class="arrow"></span>
```
rpkilby: echosystem :P

lukepolo: ha :-)

yyx990803:
Thanks! Wondering why we haven't thought of this earlier.... ;)

Issue #565, opened 2016-11-02
Difficulties in understanding vue inner workings


thewhitetulip:
Note: this might sound arrogant, but this is a real issue, don't mean to to arrogant!

I have 0 experience in front end dev. There are a host of issues I have for getting started with Vue, despite the detailed and otherwise wonderful docs. I think Vue should have a "tutorial" like the Python tutorial, if you guys are busy to write it, I'll write it (on the lines of this https://github.com/thewhitetulip/web-dev-golang-anti-textbook/), I want a document which teaches vue.js with example. It is one thing to tell `how to bind a JS variable to a <p>` and how to use vue in general and how it works.

The docs just skip the essential "getting started with vue" part and go ahead teaching the basics, there are 0 docs about how to set up & explain the setup, the docs have the `npm install` thing, but npm is like magic to me, don't understand a thing about it. I would like to know what that npm does, or what the vue cli does, how things work, I read the docs for three days and yet I can't understand _how_ vue works, do I import it? do I have it in a `<script>` tag? how do things work?

How exactly is the project run? what if I do not want to have a JS based server? how do I do that in vue? (I think this is an advanced topic, but since the docs don't tell about how to get started, it is difficult for newbies to front end dev like me to understand how to use vue and get started with the front end world which is full of js frameworks)

writeee:
<script> works.
"Simply download and include with a script tag. Vue will be registered as a global variable." (http://vuejs.org/guide/installation.html)

You can write your vue code inside another script tag (or include).

-----

Npm is a tool for managing packages, docs: https://docs.npmjs.com

writeee:
Script tag example:
```
<body>
  <script src="vue.js"></script>
  <div id="app">
    {{ message }}
  </div>
  <script>
    var app = new Vue({
      el: '#app',
      data: {
        message: 'Hello Vue!'
      }
    })
  </script>
</body>
```


Npm is needed if you want to build more complex web (front-end) apps.
See https://docs.npmjs.com/

And this: https://github.com/lukehoban/es6features
(How the "import" works in JavaScript.)

Then: https://github.com/vuejs/vuex

More: http://webpack.github.io/

thewhitetulip:
@writeee This might suffice if I go ahead and dive deep into the things, but don't you think we need a chapter to explain this in the bare minimum essentials in the docs itself?

You linked to the TODOmvc project, it doesn't have a  `<script>` tag where vue.js is placed as src

writeee:
[...]

PR #591, opened 2016-11-18
Update index.md

thewhitetulip:
As raised in issue #565, added few more lines which would immensely help total new comers to front end. Also added about changing the delimiter, because I think the first page itself should contain the information to change the delimiter, so the need to do an internet search for it is saved.


yyx990803:
Thanks for the PR, but honestly I don't find these additions necessary. (Note the current version is already adjusted based on the feedback in #565).

For the first: you don't really need `<html>`, `<head>` or `<body>` if you are running the example code in a decently up-to-date browser. The brevity is intentional. The snippet simply tells the user how to "install" Vue on the page. They are supposed to then copy the examples below.

For the second: the guide has nothing to do with server side templates, although the tip itself is useful, this doesn't seem to be the right place for it.

thewhitetulip:
>The brevity is intentional.

So we are deliberately teaching people not to use valid HTML in the official docs?

PR #599, opened 2016-11-21
Update select2.md

larsboldt:


chrisvfritz:
We're not trying to demo every feature of select2 here - just give a simple example of a typical integration. Since the focus is on Vue and not select2 specifically, I think we can leave out mentions of all the non-default features we're not including in the example.

PR #600, opened 2016-11-23
add larger example of full async component

james2doyle:
I was pretty confused about async components and how they worked with `props` and `data`. I didn’t realize that you just moved your whole application logic into the object being resolved.

I feel this additional paragraph makes that much clearer. It also showcases an approach to loading a remote template and then rendering the results.

I attached a screenshot below to show the visible changes:

screen shot 2016-11-22 at 5 08 05 pm

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -994,7 +994,38 @@ Vue.component('async-example', function (resolve, reject) {
 })
 ```
 
-The factory function receives a `resolve` callback, which should be called when you have retrieved your component definition from the server. You can also call `reject(reason)` to indicate the load has failed. The `setTimeout` here is simply for demonstration; How to retrieve the component is entirely up to you. One recommended approach is to use async components together with [Webpack's code-splitting feature](http://webpack.github.io/docs/code-splitting.html):
+The factory function receives a `resolve` callback, which should be called when you have retrieved your component definition from the server. You can also call `reject(reason)` to indicate the load has failed. The `setTimeout` here is simply for demonstration; How to retrieve the component is entirely up to you.
+
+Using this convention, you can move the logic of your application into the object that is being resolved. A full example of an async component might look like the following:
+
+```html
+<async-example message="Done!"></async-example>
+```
+
+```js
+Vue.component('async-example', function(resolve, reject) {
+  // our template is a remote HTML partial
+  fetch('/async.html')
+    .then(function(res) {
+      return res.text();
+    }).then(function(template) {
+      resolve({
+        props: ['message'],
+        data: function() {
+          return {
+            time: Date.now()
+          };
+        },
+        template: template,
+        created: function() {
+          console.log('component is ready.')
+        }
+      });
+    }).catch(reject);
+})
+```
+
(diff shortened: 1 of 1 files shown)


chrisvfritz:
You don't actually move your _application_ logic to the object being resolved. The resolved object is simply a component definition, so works the same as any other component definition. I just [added a comment](https://github.com/vuejs/vuejs.org/commit/008888fd02fa3ed752e1e953f114a186ec6f9e27) to try to make that clearer. Would love to hear your thoughts.

james2doyle:
Yeah, _application logic_ is not right, component definition is better.

Thinking as if I didn't know what I do now, I don't believe that additional comment would truly communicate the usage that I needed to see.

Reading the section over again, I think the main issue is that it doesn't even utilize the `resolve` function in a way that showcases the most common reason for using async components: _remote data_.

When I read it for the first time, it seemed like the `resolve` function _only took an object with a `template` property_. The convention of a promise-style constructor doesn't exist in any other way in the component docs, and I feel the example glosses over the usage very quickly.

I'm guessing that most people are either using the async component to load an object with their component definition, or to load a remote template (like in my example above).

But the example shows neither of those and doesn't showcase the actual use-case for the async component. The timeout example is a little contrived.

chrisvfritz:
Ah, thanks for clarifying. I'll give this some more thought and see what I can do to improve these.

chrisvfritz:
@james2doyle Would you actually mind providing a bit more information on your use case for loading a remote template? That's one I've never actually encountered before.

james2doyle:
Sure. I was actually building a component that would load content based on the device features, platforms, or authentication.

So I would make a request to the server, then load up the component template based on what kind of user it is. If the user is authed, my template would be showing additional data, and the reverse for un-authed.

An example would be:

```js
Vue.component('profile', function(resolve, reject) {
  fetch('/api/profile.json')
    .then(function(res) {
      return res.json();
    })
    .then(function(res) {
      resolve({
        data: function() {
          return res.data;
        },
        template: "#some-inline-template"
      });
    ).catch(reject);
})
```

I could also just have a static object that resolves normally, but the request for the template (like in my example above) is a path that is depending on what device features are available.

Something like:

```js
Vue.component('slider', function(resolve, reject) {
  var componentTemaplatePath = (Modernizr.touchevents) ? "/components/mobile/": "/components/desktop/";
  fetch(componentTemaplatePath + '/slider')
    .then(function(res) {
      return res.json();
    })
    .then(resolve)
    .catch(reject);
})
```

You can see I can now load templates based on a request, and not hard coded strings or `template` tags in my HTML page.

The strongest use case is probably for authenticated data requests.

Does this make sense?

chrisvfritz:
Thanks again for the clarification. I'll take a stab at making these examples less contrived, while keeping them simple enough to understand right away. Also, I really love your async component with modernizr, so I think we'll be adding a similar example to the work-in-progress cookbook. 😄

james2doyle:
Awesome! If there is anything else I can make another PR if this one needs a different story

chrisvfritz:
Thanks! Rather than forcing you to read my mind though, I'll spend some time trying to rework it with the style I try to maintain and then get your feedback. What you've already provided has been extremely helpful!

Issue #587, opened 2016-11-16
Explain Usage of the special "key" attribute outside of v-for more throughly.


LinusBorg:
While the docs state that the use of `key` is not limited to the context of `v-for`, it is never prominently explained what those other usecases really are (except briefly  in the Advanced Section about Transitioning Components).

Issues like vuejs/vue#4216 and questions on the forum show that people often see "strange" behaviour where all that's missing is a `key` to keep Vue from re-using an element where it shouldn't.

So we need to find a good place (or maybe multiple places) to explain this more thoroughly.

yyx990803:
@LinusBorg do you have links to forum posts that have to do with this? Would like to see a few to help pick example code to use.

LinusBorg:
Will have to look for some, don't have them bookmarked I think.

LinusBorg:
This one, for example:

https://forum.vuejs.org/t/vnodes-produced-by-createelement-in-render-function-shares-same-data-object/3013

Will post more (older ones) when I find them.

chrisvfritz:
@LinusBorg I'll close this for now, in the hope that this change resolves the confusion. If you find that we still need to include info in more parts of the guide, please don't hesitate to open another issue. 🙂

LinusBorg:
@chrisvfritz Yeah that's ok. I will come back to this if I find additional pain points from my experience on the forum etc.

PR #602, opened 2016-11-23
Add section to conditionals about key attribute

chrisvfritz:
@LinusBorg Do you think this is a step in the right direction towards resolving #587? Also, is it short enough and explained in simple enough terms so as not to confuse beginners?

### DIFF
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -4,7 +4,7 @@ type: guide
 order: 7
 ---
 
-## v-if
+## `v-if`
 
 In string templates, for example Handlebars, we would write a conditional block like this:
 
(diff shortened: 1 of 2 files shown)


src/v2/guide/conditional.md
```diff
@@ -28,7 +28,7 @@ It is also possible to add an "else" block with `v-else`:
 <h1 v-else>No</h1>
 ```
 
-### Template v-if
+### Conditional Groups with `v-if` on `<template>`
```
yyx990803: Does this look okay in the sidebar?

yyx990803:
I like it. I had a rough draft a while ago but I like your version much better :)

src/v2/guide/conditional.md
chrisvfritz: Good catch! I just added a function to escape HTML in sidebar links.

LinusBorg:
Yeah this is really good! It shows that you are an educator - the way you introduce the upside of this behaviour before introducing a way to handle the "downside" is great!

And I think the message is transported very clearly.

PR #603, opened 2016-11-23
add language menu

kazupon:
released vue.js 2.0 translation for japanese
http://jp.vuejs.org
ref: https://github.com/vuejs/jp.vuejs.org/issues/189


yyx990803:
Let's also add a link for Chinese (中文)：cn.vuejs.org

kazupon:
done :)

PR #608, opened 2016-11-24
Request for more explanation

glittle:
My edit is just a question to the writer. Please remove my text!

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -424,6 +424,8 @@ computed: {
 
 Alternatively, you can also just use a method where computed properties are not feasible (e.g. inside nested `v-for` loops):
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
It _is_ feasible. The text mentions this as an alternative "where __computed properties__ are not feasible (e.g. inside nested `v-for` loops)." Does that answer your question?

glittle:
I'm still confused. It says "Alternatively, you can also just use a method where computed properties are not feasible (e.g. inside nested v-for loops):"

But the example given to explain that seems to be equivalent to the one just above it, so must be a case where they are feasible. 

I'm just suggesting that the HTML of the example should be different, showing a case when computed properties are not feasible!

PR #613, opened 2016-11-26
Update unit-testing.md

zesani:
edit expect 
references from https://github.com/Automattic/expect.js


chrisvfritz:
I don't think there's any reason to switch to expect.js-style syntax here. The style currently used in the doc can be used with Jasmine, Jest, and Mocha ([via Expectations](https://github.com/mochajs/mocha/wiki#assertion-libraries)), so it's actually much more universal.

PR #617, opened 2016-11-29
Mention `v-model` does not update state during input composition

phanan:


chrisvfritz:
@phanan Looks good, but is this something that should be fixed in Vue?

phanan:
No it's not. This is by design, as I understood from Evan in the related issue.

chrisvfritz:
Just found that issue - I understand now, thanks!

PR #619, opened 2016-11-30
Correcting a syntax error in an example

arnolem:
Semicolons were missing at the end of the instruction


chrisvfritz:
Not missing - [semicolons are unnecessary](https://github.com/yyx990803/semi#but-semicolons-are-required) and we opt for a cleaner syntax.

PR #621, opened 2016-11-30
Add static class and style to render function data object

defcc:


chrisvfritz:
@defcc How do the static versions behave differently from the dynamic versions? Are there any performance benefits? Will dynamic data bound to them not result in a DOM update?

defcc:
@chrisvfritz They will be merged together.

```js
h('div', {
  staticStyle: {
    color: 'red'
  },
  staticClass: 'class-name'
})
```

chrisvfritz:
Hmm, I still have a few more questions, but I'll bring it up on the #help channel on Slack, as I have a feeling I may not be the only one who'd like to learn more about the subtleties here.

chrisvfritz:
OK, deciding to leave this out of the docs, as people are unlikely to want to use this directly.

emileber:
The **documentation should be explicit and complete**. It's not a getting started tutorial, it's a reference.

> as people are unlikely to want to use this directly

The doc on data object is kind of misleading:

```
  // Normal HTML attributes
  attrs: {
    id: 'foo'
  },
```
Since `staticClass` isn't listed, I was wondering why my `attrs: { class: 'static-class' } }` wasn't added to an element. I find it odd to have to do `class: { 'static-class': true }`...

chrisvfritz:
I think `attrs: { class: 'my-class' } }` might make sense as an alias for `class: { 'my-class': true }`. I'd recommend opening an issue on the [main repository](https://github.com/vuejs/vue) to make that happen.

The reason we don't list `staticClass` and `staticStyle` is because they're really a compilation optimization that's very, very easy for users to misuse, while not really opening up any new possibilities. For example, `staticClass: this.myClass` will cache the initial value of `this.myClass`, but then won't update when `this.myClass` changes.

fnlctrl:
For the record, `staticClass` updates when bound value changes: https://jsfiddle.net/yym9hc30/
And it's indeed best to leave undocumented since it's really counter-intuitive...

PR #624, opened 2016-12-01
Fix broken transition example

callumacrae:
Padding is set to `0 !important`, so this example didn't work. It's also
bad practice to transition padding unless you really have to: in this
case, a transform is more performant, so I changed it to that instead.


callumacrae:
Will this need fixing on the other repos too, or are they completely different docs? Happy to send more PRs.

chrisvfritz:
I believe this is the only version of the docs with this specific example. Thanks!

PR #625, opened 2016-12-01
The use of mixed quotes for src= is not intuitive

rickatech:
Providing example here will help avoid such confusion!

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1518,7 +1518,9 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
   <!-- shorthand -->
   <img :src="imageSrc">
-  <img v-bind:src="'/storage/userimages/' + item.imagefile">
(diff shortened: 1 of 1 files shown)


src/v2/api/index.md
```diff
@@ -1518,6 +1518,7 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
   <!-- shorthand -->
   <img :src="imageSrc">
+  <img v-bind:src="'/storage/userimages/' + item.imagefile">
```
kazupon: Here is shorthand.
If you add the mixed quotes, you should be added the above into example.

rickatech: Or perhaps go with more concise pseudo prose ....

chrisvfritz:
Tweaked it a bit. Thanks!

rickatech:
:-)

PR #630, opened 2016-12-03
one small mistake

yanlee26:
I think here is the 'is',instead of 'if'.


chrisvfritz:
`if` is actually correct

yanlee26:
Alright...,but i was confused to got the message obviously,can you explain it in another way?
Many tks!

chrisvfritz:
I just tried rephrasing this comment to:

> The name of the slot, if this component is the child of another component

Does that make more sense?

PR #632, opened 2016-12-03
tweak table style

kazupon:
Grit Component table like:
https://vuejs.org/v2/examples/grid-component.html

table style example:
screen shot 2016-12-03 at 7 08 31 pm


chrisvfritz:
I like this way better! 😍

PR #636, opened 2016-12-07
update Comparison/TypeScript/decorator

kaorun343:
Because
* we have official decorator
* `vue-typescript` is not for the current version of Vue.js


ktsn:
LGTM

PR #645, opened 2016-12-11
replace h by createElement

sylvainpolletvillard:
I guess this is an oversight.


chrisvfritz:
h is a common alias for createElement, but it's only introduced further down, so good catch!

PR #647, opened 2016-12-12
add Circular References Between Components demo

xalexec:


xalexec:
pr is some wrong , delete

PR #649, opened 2016-12-12
Add Vue + TypeScript guide

ktsn:
I've wrote a beginner's guide of using Vue.js with TypeScript. It introduces official declaration files and [vue-class-component](https://github.com/vuejs/vue-class-component) briefly. I think readers can make simple applications after understanding this guide.

I added this guide after SSR guide but not sure it is appropriate place.

@kaorun343 @HerringtonDarkholme Your review would be appreciated if you have time :)


src/v2/guide/typescript.md
```diff
@@ -0,0 +1,104 @@
...
+  props: ['message'],
+  template: '<span>{{ message }}</span>'
+
+// annotate ComponentOptions with `as` operator
+} as Vue.ComponentOptions<Vue>
+```
+
+You can see all available types on the corresponding repository ([here are Vue's types](https://github.com/vuejs/vue/blob/dev/types/index.d.ts)).
+
+<p class="tip">The declaration files of Vue requires `--lib DOM,ES2015` as compiler option. So you need to pass the option via `tsc` command or `tsconfig.json`. For more details of compiler options, please see [Compiler Options · TypeScript](https://www.typescriptlang.org/docs/handbook/compiler-options.html)</p>
```
HerringtonDarkholme: I would suggest to require only `--lib DOM,ES2015.promise`. Because `Promise` is the only requirements in Vue(and it is optional, if users do not use async component.). IMHO, listing only API we used is in accordance with the rule of least power.

src/v2/guide/typescript.md
```diff
@@ -0,0 +1,104 @@
+---
+title: TypeScript
+type: guide
+order: 25
+---
+
+If you develop a large application, using a language having static type checking would be helpful. If your code seems to cause unexpected behavior, it provides useful errors before you run it on browsers. You can easily use [TypeScript](https://www.typescriptlang.org/) (one of popular AltJS having static type checking) with Vue since Vue officially provides its declaration files.
```
HerringtonDarkholme: > If your code seems to cause unexpected behavior

This sentence does not make much sense to me.... Maybe `If you have unexpected typos in your code` is better?

src/v2/guide/typescript.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Official Declaration Files
+
+There are official declaration files for Vue, vue-router and vuex. You do not have to use external tools like `typings` to use them since they are published via npm. You can install Vue's declarations with `npm install vue`. Then, all declarations can be imported with `import` statement.
+
+Example of component definition with TypeScript:
+
+``` ts
+// we cannot write `import Vue from 'vue'` in TypeScript
```
ryutamaki: We can use `import * as Vue from "vue";` this syntax instead (at least, working well to me).

src/v2/guide/typescript.md
```diff
@@ -0,0 +1,104 @@
...
+import Component from 'vue-class-component'
+
+// ECMAScript Decorator syntax
+// @Component indicates the class will be
+// defined as a Vue component
+@Component({
+  // all component options are allowed in here
+  template: '<button @click="onClick">Click!</button>'
+})
+export default MyComponent extends Vue {
```
kaorun343: `class` is missing

```typescript
export default class MyComponent extends Vue
```

kaorun343:
It looks so great to me! 😄

src/v2/guide/typescript.md
ktsn: Yeah, it would be clearer and understandable. So how about `For example, if you have typos in your code, ...` ? Because typos are always unexpected thing 😛

src/v2/guide/typescript.md
ktsn: @ryutamaki 
Actually, `import * as Vue from "vue";` and `import Vue = require("vue");` are the same thing. I choose the later syntax because it would be more intentional to import single object. 

But maybe writing the former syntax as a note would also be good. I'll add about that, thanks for pointing this out!

src/v2/guide/typescript.md
HerringtonDarkholme: `If you have unexpected mistakes in your code`?

src/v2/guide/typescript.md
ktsn: Thanks, I often miss it 😅

src/v2/guide/typescript.md
```diff
@@ -0,0 +1,104 @@
...
+title: TypeScript
+type: guide
+order: 25
+---
+
+If you develop a large application, using a language having static type checking would be helpful. If your code seems to cause unexpected behavior, it provides useful errors before you run it on browsers. You can easily use [TypeScript](https://www.typescriptlang.org/) (one of popular AltJS having static type checking) with Vue since Vue officially provides its declaration files.
+
+## Official Declaration Files
+
+There are official declaration files for Vue, vue-router and vuex. You do not have to use external tools like `typings` to use them since they are published via npm. You can install Vue's declarations with `npm install vue`. Then, all declarations can be imported with `import` statement.
```
ktsn: It seems that Vue is title case but vue-router and vuex is following github repository name in the docs.
Is that correct? /ping @chrisvfritz

src/v2/guide/typescript.md
DanielRosenwasser: Personally, I'd say that `import Vue = require("vue");` is the most appropriate way to import a CommonJS module given that the interop behavior hasn't been settled on yet.

chrisvfritz:
@ktsn @kaorun343 @HerringtonDarkholme I just revised this. Fantastic info here! Beyond tweaking some language and reorganizing a bit, I removed some explanations about TypeScript that weren't specific to Vue (e.g. different ways of importing JS modules), as I think we might want to assume some TypeScript experience here. I'm not an expert though, so let me know if you think that was a mistake.

src/v2/guide/typescript.md
```diff
@@ -0,0 +1,98 @@
+---
+title: TypeScript Support
+type: guide
+order: 25
+---
+
+## Official Declaration Files
+
+A strong type system can help prevent many potential runtime errors, especially as applications grow. That's why Vue ships with [official type declarations](https://github.com/vuejs/vue/tree/dev/types) for [TypeScript](https://www.typescriptlang.org/) - not only in Vue core, but also [for Vue Router](https://github.com/vuejs/vue-router/tree/dev/types) and [for Vuex](https://github.com/vuejs/vuex/tree/dev/types) as well.
```
HerringtonDarkholme: Would you mind replace `strong` with `static`? TypeScript isn't type sound in language design. `static` is probably a better choice to express compile time checking.

chrisvfritz: Done!

ktsn:
@chrisvfritz Looks great! Thanks for your revising 😄

STayinloves:
Could you please to tell me which color theme do you use in the VScode? Thank you.
pic

chrisvfritz:
@STayinloves It's all a lie, unfortunately. @phanan photoshopped a screenshot I took to better match Vue's colors. 🙂

phanan:
😛 I'm sorry…

STayinloves:
Oh, anyway, still thank both of you. Maybe can hack one some day. @chrisvfritz @phanan

STayinloves:
I write a color theme like this for VSCode today. Published at [here](https://marketplace.visualstudio.com/items?itemName=staychan.vue-color). Someone may have interest.😛

PR #650, opened 2016-12-12
Class name should be a string when binding

itainoam:


chrisvfritz:
Was it not working for you without the quotes?

itainoam:
Yeah, didn't work for me. 
Take a look at this [jsfiddle](https://jsfiddle.net/yMv7y/2011/).
Line 5, if you remove quotes it stops working.

LinusBorg:
That's because Javascript object properties have to be strings if they contain a hyphen (`-`). So that's expected behaviour.

Did you find a place in the docs where we missed to add the quotes in such a case?`Or why did you open this issue?

itainoam:
Got it. Thanks for detailed response!

PR #653, opened 2016-12-13
add Circular References Between Components  example

xalexec:


chrisvfritz:
Thanks for the contribution! Can you please provide a little more detail about the problem you're trying to solve here?

xalexec:
When I read the document, the above fragments make me understand some of the deviation, , so I want to write an example so that others can understand easily.

chrisvfritz:
Is there a way you think we can improve the existing example to make it clearer? For example, is there information we might be missing? I think I'd rather improve the existing example rather than just linking to another one.

xalexec:
I think it's better to provide a complete example than a snippet,this is i can do.

PR #654, opened 2016-12-13
Update comparison.md

Jameron:
I would like to get some comparisons to backbone.js in here. Backbone was huge and still has a lot of legacy users. Also, hoping to get Vue's core team to consider models and collections.


chrisvfritz:
I appreciate the contribution, but I'd like to reserve this page for actively maintained projects. Backbone's creator now considers it "finished", it hasn't had even a patch release in 8 months, and more projects in the ecosystem seem to be retiring than being created, as people move on to other frameworks. When choosing a UI library/framework for a new project, I just don't think many people will considering Backbone as an option and needing to weigh it against Vue.

As for models and collections, I certainly can't speak for the entire team, but I wouldn't want them in Vue. As a former Backbone user myself, I found those abstractions really useful at the time, but am much happier with more modern state management solutions like Vuex, Redux, Apollo, etc.

PR #655, opened 2016-12-14
Updated to 2.1.6

Uriziel01:


chrisvfritz:
Thanks for the reminder! We actually have an update script we run, so I ran that instead for a more complete update of the docs. We also prefer to run that ourselves for security purposes - I'm sure you can understand. It could be easy to miss a tiny change to the downloaded Vue version in a PR.

Uriziel01:
Its perfectly fine mate. I do realise that I would force you to review all the code and check against the current version, the fact that You have a script for that taks is a great news 👍

PR #662, opened 2016-12-19
In v-for index/item example swapped.

awoehler:
In the index example for v-for the index and item were in the wrong order.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -60,7 +60,7 @@ Inside `v-for` blocks we have full access to parent scope properties. `v-for` al
 
 ``` html
 <ul id="example-2">
-  <li v-for="(item, index) in items">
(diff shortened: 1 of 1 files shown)


kazupon:
Thank you for your PR.
This order is correct.
See the test code.
https://github.com/vuejs/vue/blob/dev/test/unit/features/directives/for.spec.js#L35

awoehler:
Thank you for the quickly reply, but the example does not work ask described. See https://jsfiddle.net/awoehler/2f1aay4q/4/ [Object] is displayed instead of the index.

kazupon:
The version of the library that you are using is Vue 1.x.
You can see the vue.js.org for Vue 1.x via selectbox from side-menu. 😉 
http://v1.vuejs.org

PR #663, opened 2016-12-19
Clarification for runtime vs standalone

ubershmekel:
Removed some redundant language by including all the details in the bullet point comparison. Also added a note about Vuex requiring the standalone-build. Hopefully this is a bit more complete and easier to read.

### DIFF
--- a/src/v2/guide/installation.md
+++ b/src/v2/guide/installation.md
@@ -46,13 +46,11 @@ $ npm install vue
 
 ### Standalone vs. Runtime-only Build
 
-There are two builds available, the standalone build and the runtime-only build. The difference being that the former includes the **template compiler** and the latter does not.
+There are two builds available:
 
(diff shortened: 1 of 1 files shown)


yyx990803:
FYI Vuex store does not require the standalone build. If it does then it's a bug.

kazupon:
oh, sorry...

ubershmekel:
I hit this when I imported the runtime only build, this.$store was undefined:  https://github.com/vuejs/vuex/issues/409

yyx990803:
@ubershmekel I think you have some misunderstanding about the builds. The recommended solution is configuring the alias via your bundler so that `import Vue from 'vue'` always points to the desired build everywhere in your application. We explicitly tell you to NOT do `import Vue from 'vue/dist/vue.common.js`.

DaveStein:
I currently have an alias like `vue: 'vue/dist/vue.esm.js'`, and am using Webpack 2. I am really lost as to why `this.$store` would not be be defined while `this.$router` is. I am passing both to constructor of `Vue` and am doing `use` for both plugins.

PR #671, opened 2016-12-27
Add disclaimer that $on cannot be used to listen to events from children

pholly:
I assumed parents could listen to child events using `$on` and that `v-on` was just convenient syntax. Adding a disclaimer may help others like me, especially those that come from an Angular or Vue 1.0 background.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -448,6 +448,8 @@ Every Vue instance implements an [events interface](../api/#Instance-Methods-Eve
 
 In addition, a parent component can listen to the events emitted from a child component using `v-on` directly in the template where the child component is used.
 
+<p class="tip">You cannot use `$on` to listen to events emitted by children. You must use `v-on` directly in the template, as in the example below.</p>
+
 Here's an example:
 
 ``` html


src/v2/guide/components.md
```diff
@@ -448,6 +448,8 @@ Every Vue instance implements an [events interface](../api/#Instance-Methods-Eve
 
 In addition, a parent component can listen to the events emitted from a child component using `v-on` directly in the template where the child component is used.
 
+<p class="tip">You cannot use `$on` to listen to events emitted by children. You must use `v-on` directly in the template, as in the example below.</p>
```
kazupon: duplicate the above paragraph ?

pholly: Explicitly state that you cannot use `$on` to listen to events emitted by children. My friend and I both tried to use `$on` and I saw some questions on the vue forums as well. 

The "Custom Events" section begins by introducing Vue's custom event system as a way to communicate from a child to parent. Then it speaks about the event interface using `$on` and `$emit`. One can only assume that `$on` and `$emit` are how to communicate from child to parent. There should be some language somewhere that states the only way is to use `v-on`. Either that or move the event system introduction out of components, because it applies to Vue instances and not just components, but that's a bigger change.

Any suggestions?

kazupon: Indeed, that is better.

PR #672, opened 2016-12-27
Arguments in $mount didn't match codebase.

trentmillar:
Because of this, readers are unable to link to "Source" in Github

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1178,10 +1178,10 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
 ## Instance Methods / Lifecycle
 
-<h3 id="vm-mount">vm.$mount( [elementOrSelector] )</h3>
+<h3 id="vm-mount">vm.$mount( el, hydrating )</h3>
 
 - **Arguments:**
-  - `{Element | string} [elementOrSelector]`
+  - `{Element | string} [el]`
   - `{boolean} [hydrating]`
 
 - **Returns:** `vm` - the instance itself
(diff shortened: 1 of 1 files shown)


trentmillar:
Sure, I agree. The issue then is the implied search to the codebase. This will be broken unless you change the actual argument name in the code from `el` to `elementOrSelector`.

src/v2/api/index.md
```diff
@@ -1178,10 +1178,10 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
 ## Instance Methods / Lifecycle
 
-<h3 id="vm-mount">vm.$mount( [elementOrSelector] )</h3>
+<h3 id="vm-mount">vm.$mount( el, hydrating )</h3>
```
kazupon: Oh, indeed.
Search is broken...
I think try to merge your PR as quick fix.
However, you should be fix arguments of `$mount`.

`<h3 id="vm-mount">vm.$mount( [el, [hydrating] )</h3>`

chrisvfritz:
@trentmillar Thanks for bringing up that this wasn't working correctly! I didn't like this particular solution, as it sacrifices the descriptiveness of the docs to improve the search, but I found an alternative. The generated search strings should now be a little smarter, yielding better results for all API entries. As an example, here's [the new search query](https://github.com/search?utf8=%E2%9C%93&q=repo%3Avuejs%2Fvue+extension%3Ajs+%22Vue.prototype.%24mount%22&type=Code) for `vm.$mount`.

PR #674, opened 2016-12-29
fix: added css override property to make certain links clickable

gesslerpd:
First noticed this [here](https://vuejs.org/v2/guide/custom-directive.html#Directive-Hook-Arguments).

The *VNode API* link click target was being covered by previous `ul` and `li` tags and I added a css override to allow for these links to be clickable. I am using Chrome 55.0.2883.87 and image of the link is below.

image


kazupon:
thanks!

PR #676, opened 2016-12-29
Add reminder about using quotes with kebab-case

regularmike:
It may be obvious to many users but the guide seems to suggest that you can use unquoted kebab-case CSS property names. A reminder to use quotes here would be helpful since the "invalid expression" warning could be vague in many cases.


kazupon:
Thanks! Indeed, it's user-friendly :)

PR #677, opened 2016-12-30
Update to 2.1.8

ChangJoo-Park:


kazupon:
Thank you for your PR.
However, that is updated with `make update`.
So, close. 😉

PR #681, opened 2016-12-31
Cookbook v clickoutside

LinusBorg:
Add Cookbook chapter about a clickoutside directive.

Anybody care to review? I'm good at having typos *everyhere* :D

### DIFF
--- a/src/v2/cookbook/clickoutside-directive.md
+++ b/src/v2/cookbook/clickoutside-directive.md
@@ -159,6 +159,11 @@ unbind(el, binding) {
 }
 ```
 
+## The Final Result
+
+<iframe width="100%" height="500" src="https://jsfiddle.net/Linusborg/g5sh3rcu/embedded/result,html,js,css" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
+<script async src="//"></script>
+
 ## Limitations
 
 #### Doesn't work with v-show
--- a/src/v2/cookbook/clickoutside-directive.md
+++ b/src/v2/cookbook/clickoutside-directive.md
@@ -132,3 +132,45 @@ Vue.directive('clickoutside', {
   }
 })
 ```
+## Argument: Using different event types.
+
+So far, your directive will only listen for `click` events. But there might be situations where we want to listen to other events, like `doubleclick` or `mousedown`.
+
+For this, we can use an argument on our directive. This is what this looks like in our HTML:
+```html
+<modal v-clickoutside:dblclick="handler">
+
+</modal>
+```
+
+To use the argument in our directive, we only have to change the first argument of our `addEventListener()` and `removeEventListener()` calls to use the argument, if present:
+
+```JavaScript
+bind(el, binding) {
+  // ...
+  const eventName = binding.arg || 'click'
+  document.addEventListener(eventName, /*...*/)
+}
+
+unbind(el, binding) {
+  // ...
+  const eventName = binding.arg || 'click'
+  document.removeEventListener(eventName, /*...*/)
+}
+```
+
+## Limitations
+
+#### Doesn't work with v-show
+
+`v-show` does not remove the element from the DOM, so `unbind()` is not triggered when `v-show` hides an element. This means that our clickoutside event is still active even if your modal is not visible, and would still fire for each click outside of the modal (which now, is basically *everywhere*)
+
+#### No inline expressions
+
+We can't use an inline expression instead of a handler function. The following will **not** work, because it will be run immediatly, not on `click`:
+
+```html
+<modal v-clickoutside="showModal= false">
+
+</modal>
+```
--- a/src/v2/cookbook/clickoutside-directive.md
+++ b/src/v2/cookbook/clickoutside-directive.md
@@ -6,7 +6,7 @@ order: 2.1
 
 ## What we are building
 
-Many UI elements need to react to clicks that happen *outside* of them. Common usecases are:
+Many UI elements need to react to clicks that happen *outside* of them. Common use cases are:
 
 * Modals
 * Dropdown menus
@@ -172,7 +172,7 @@ unbind(el, binding) {
 
 #### No inline expressions
 
-We can't use an inline expression instead of a handler function. The following will **not** work, because it will be run immediatly, not on `click`:
+We can't use an inline expression instead of a handler function. The following will **not** work, because it will be run immediately, not on `click`:
 
 ```html
 <modal v-clickoutside="showModal= false">
--- a/src/v2/cookbook/clickoutside-directive.md
+++ b/src/v2/cookbook/clickoutside-directive.md
@@ -77,7 +77,7 @@ Vue.directive('clickoutside', {
     const handler = binding.value
 
     // create a named function for the handler
-    function handler(event) {
+    function wrappedHandler(event) {
       const target = event.target
       if (!el.contains(target) && el !== target) {
         handler(event)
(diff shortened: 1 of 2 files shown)


chrisvfritz:
Nice!! I'll review shortly. 😄

simplesmiler:
There's one not immediately obvioius limitation I've stumbled upon with [vue-clickaway](https://github.com/simplesmiler/vue-clickaway). It's that you lose the ability to capture the "context" from the template. E.g. with `v-on` you can do:
```html
<li v-for="item in items" v-on:click="process(item.id)">{{ item.title }}</li>
```
, but you can't trivially do the same with a custom directive.

LinusBorg:
@simplesmiler yes that's right. But returning a closure function from that call is an easy enough workaround. 

I will add this to this article, thanks!

simplesmiler:
@LinusBorg something like `"makeProcessCb(item.id)"` or even `"ev => process(item.id)"`?

LinusBorg:
I don't honk you can define arrow functions inline in Vue templates, but yeah, thats the basic idea

chrisvfritz:
@LinusBorg Actually, Babel parses the JS within Vue templates too. 🎉  That's with Webpack at least. I'm not sure about Browserify.

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Limitations
+
+#### Doesn't work with v-show
+
+`v-show` does not remove the element from the DOM, so `unbind()` is not triggered when `v-show` hides an element. This means that our clickoutside event is still active even if your modal is not visible, and would still fire for each click outside of the modal (which now, is basically *everywhere*)
+
+#### No inline expressions
+
+We can't use an inline expression instead of a handler function. The following will **not** work, because it will be run immediatly, not on `click`:
```
intelliot: immediately

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
+---
+title: A clickoutside directive
+type: cookbook
+order: 2.1
+---
+
+## What we are building
+
+Many UI elements need to react to clicks that happen *outside* of them. Common usecases are:
```
intelliot: use cases

LinusBorg:
Thanks @intelliot, will fix those

sqal:
> v-show does not remove the element from the DOM, so unbind() is not triggered when v-show hides an element

Hmm correct but `update` hook is called and then we can check if vnode has v-show directive and if element is visible

```js
update(el, binding, vnode) {
  let isShown = false;
  const hasShowDirective = vnode.data directives.some(d => {
    if (d.name === 'show') {
      isShown = d.value;
      return true;
    }

    return false;
  });

  if (hasShowDirective) {
    if (isShown) {
      // bind
    } else {
      // unbind
    }
  }
}
```

LinusBorg:
@sqal That's a really great idea, thanks! 

I'm a bit unsure wether we should tread into vnode territory in these kinds of recipes that are kind of basic, though.

Maybe I can add it as a seperate "advanced" section at the end.

src/v2/cookbook/clickoutside-directive.md
LinusBorg: fixed

src/v2/cookbook/clickoutside-directive.md
LinusBorg: fixed

LinusBorg:
@sqal unfortunately that doesn'T work, at least not when v-show is used on a component (e.g. `<modal v-show="showModal">`). The directive is not present on the v-node in that case.

sqal:
hmm I'm afraid I don't get what you mean - https://jsfiddle.net/uhhdeqhq/

LinusBorg:
Hm, thanks ... 

Somehow this didn't work for me yesterday, but it seems I didn't save my fiddle.

Will try again ...

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
...
+    }
+    // ...
+  }
+})
+```
+## Argument: Using different event types.
+
+So far, your directive will only listen for `click` events. But there might be situations where we want to listen to other events, like `doubleclick` or `mousedown`.
+
+For this, we can use an argument on our directive. This is what this looks like in our HTML:
```
myst729: This is what ~~**this**~~ **it** looks like in our HTML:

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
...
+## The Final Result
+
+<iframe width="100%" height="500" src="https://jsfiddle.net/Linusborg/g5sh3rcu/embedded/result,html,js,css" allowfullscreen="allowfullscreen" frameborder="0"></iframe>
+<script async src="//"></script>
+
+## Limitations
+
+#### Doesn't work with v-show
+
+`v-show` does not remove the element from the DOM, so `unbind()` is not triggered when `v-show` hides an element. This means that our clickoutside event is still active even if your modal is not visible, and would still fire for each click outside of the modal (which now, is basically *everywhere*)
```
myst729: Missed a period signal at the end.

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
...
+
+#### No inline expressions
+
+We can't use an inline expression instead of a handler function. The following will **not** work, because it will be run immediately, not on `click`:
```
myst729: "...it will be run immediately" sound strange. "...it will ~~be~~ run immediately" or "...it will be **fired** immediately"? Not sure though.

LinusBorg:
Just a quick note that I haven't forgotten about this, but I haven't found the time and mood to re-work this with the visiblity check that sqal provided.

Hope to get to it soon.

gbezyuk:
@LinusBorg just a friendly reminder. I mean, maybe summer has somehow better mood )

chrisvfritz:
@gbezyuk Ha, we actually just talked about it today. I'm going to try to finish this page over the weekend.

chrisvfritz:
I didn't end up finishing this over the weekend, because I played a space-war-themed board game on Friday evening, where we set the mood with the soundtrack to Mass Effect 3. Then the rest of the weekend was mostly spent playing Mass Effect 3. No regrets, but I'm back to work on it today. 😄

gbezyuk:
@chrisvfritz fair enough, man )

src/v2/cookbook/clickoutside-directive.md
```diff
@@ -0,0 +1,181 @@
...
+    // ...
+    // el.__vueClickOutside__ = wrappedHandler
+    handlerCache.set(el, wrappedHandler)
+    // ...
+  },
+  unbind(el) {
+
+    if (handlerCache.has(el)) {
+      // get the handler from the Map
+      const hander = handlerCache.get(el)
```
LucasLeandro1204: handler

sdras:
Hi @LinusBorg! 
This example will be so useful for people! This is a really common question/use case. I'm sorry to come in so late in the game to this PR, especially when you've already put so much work into it, but we just created some [Contribution Guidelines](https://vuejs.org/v2/cookbook/cookbook-contributions.html) for The Cookbook, I'm working on keeping consistency across examples to reduce cognitive load for our users. Would you mind resubmitting/updating this PR following the template we now provide? Thanks for your patience.

sdras:
@LinusBorg since we talked and this needs a major refactor to work in the new style, would it make more sense to close it out and start a new one or do you think you could reformat this into the new style? I'm thinking that a fresh PR would be easier work on, but I leave it your judgement.

sdras:
I'm going to close this out so that it can be resubmitted aligned with our template, but feel free to reopen if you'd rather refactor and submit on this PR. :)

LinusBorg:
Alright :)

PR #689, opened 2017-01-07
enter-to-class, leave-to-class available in 2.1.8

ChangJoo-Park:


ChangJoo-Park:
Thank you @kazupon  have a good weekend :)

PR #700, opened 2017-01-12
chagne `Jade` to `Pug`

### LINKED ISSUE #695: Change Jade to Pug
In the file `https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/single-file-components.md`

We think (french translate team) which the

> As promised, we can also use preprocessors such as *Jade*, Babel (with ES2015 modules), and Stylus for cleaner and more feature-rich components.

Should be

>As promised, we can also use preprocessors such as **Pug**, Babel (with ES2015 modules), and Stylus for cleaner and more feature-rich components.

If we see others place with Jade term, we notify this here !

kazupon:
Closes #695


gbezyuk:
Guys, we at @translation-gang are waiting for this changes to be merged to merge our translations for the same thing )

PR #701, opened 2017-01-12
missing ES2015 for --lib

shomodj:


shomodj:
`tsc --version`
Version 2.1.5

_main.ts_
`import Vue = require('vue')`

not working
```
tsc --lib DOM,ES2015.Promise main.ts
error TS2318: Cannot find global type 'Array'.
error TS2318: Cannot find global type 'Boolean'.
error TS2318: Cannot find global type 'Function'.
error TS2318: Cannot find global type 'IArguments'.
error TS2318: Cannot find global type 'Number'.
error TS2318: Cannot find global type 'Object'.
error TS2318: Cannot find global type 'RegExp'.
error TS2318: Cannot find global type 'String'.
```

working
`tsc --lib DOM,ES2015,ES2015.Promise main.ts ; echo $?`

ktsn:
Thanks for correcting it!
These interfaces are defined on [lib.es5.d.ts](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts), so I prefer to `--lib DOM,ES5,ES2015.Promise` since it's minimal option.

PR #706, opened 2017-01-14
add japanese translation info

kazupon:


Alex-Sokolov:
Change minus to dash too! Not in (https://github.com/vuejs/vuejs.org/pull/707)

PR #707, opened 2017-01-14
README.md Change minus to dashes

Alex-Sokolov:


chrisvfritz:
While dashes might look slightly nicer, I think sticking with minus is OK and will be easier for people to maintain in the future. The ugliest thing would probably be inconsistency, which is more likely by introducing dashes I think.

PR #722, opened 2017-01-20
`cache` option for computed properties.

frankdiox:
I think the `cache` options works in Vue 2 but I could not find any information in the docs.
Is it just a documentation issue or was omitted on purpose for some reason? Thanks.


chrisvfritz:
I'd like to get at least one more review on this before merging though.

phanan:
LGTM.

PR #723, opened 2017-01-21
Warn about modifying instance in hooks

bmacnaughton:
I ended up spending a fair amount of time tracking down unpredictable behavior as a result of doing exactly that. I wrote components to display the "updated" count as I implemented the various parts of the guide for visual feedback on how often updates were taking place. I then wrote a component to do so and found that it worked when written certain ways but not in others, and yielded unpredictable results. I tried tracking them down.

I had thought about the circular reference/infinite loop but presumed that Vue had detection logic and broke the chain when it detected the looping (and it seemed to). So I was trying to figure out what I didn't understand about components when one didn't work as I expected. See https://stackoverflow.com/questions/41780872/vue-js-component-updates-inconsistently for details if interested.

It might make more sense not to introduce instance lifecycle events this early in the guide, but that's more major change.

### DIFF
--- a/src/v2/guide/instance.md
+++ b/src/v2/guide/instance.md
@@ -79,7 +79,7 @@ Consult the [API reference](../api) for the full list of instance properties and
 
 ## Instance Lifecycle Hooks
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I removed the warning, but added some links to the API for people who immediately want to go deeper. Since this page is just to give a very broad overview of how Vue works and explicitly states that people are not meant to understand it all yet, I want to avoid specific usage recommendations.

PR #728, opened 2017-01-25
Revert "`cache` option for computed properties."

kazupon:
@frankdiox 
From #722
>I think the cache options works in Vue 2 but I could not find any information in the docs.
Is it just a documentation issue or was omitted on purpose for some reason? Thanks.

We merged at #722, but I'll revert it due to not support `cache: false` for Vue 2 or later.
We can realize same as `cache: false` with methods.
For this reasons, `cache: false` explanation intentionally was removed it from Vue 2 docs.

/cc @yyx990803 @chrisvfritz 
Please comment if you have something unsatisfactory.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -447,7 +447,7 @@ type: api
 
   <p class="tip">Note that __you should not use an arrow function to define a computed property__ (e.g. `aDouble: () => this.a * 2`). The reason is arrow functions bind the parent context, so `this` will not be the Vue instance as you expect and `this.a` will be undefined.</p>
 
-  Computed properties are cached, and only re-computed on reactive dependency changes.
+  Computed properties are cached, and only re-computed on reactive dependency changes. Note that if a certain dependency is out of the instance's scope (i.e. not reactive), the computed property will __not__ be updated.
 
 - **Example:**
 
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -439,7 +439,7 @@ type: api
 
 ### computed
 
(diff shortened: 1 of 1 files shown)


frankdiox:
@kazupon Since `cache: false` works in 2.1.4, I guess you mean that it will be removed at some point? I'm wrapping Custom Elements inside Vue components and need to get some element properties that are not reactive. Methods can do the same but it would change the API.

chrisvfritz:
@frankdiox There are no plans to remove it in 2.x, but it will very likely be gone in v3 (whenever that might be). Please feel free to share more information about your use case though. Plans for the API do often change after learning more about how people are using specific features.

kazupon:
@frankdiox 
Yes @chrisvfritz said about it. If you have some use-cases, please share more information. :)

frankdiox:
@chrisvfritz @kazupon Thanks a lot! There are workarounds for my case so I guess `cache: false` is not a decisive feature. Here is some info: 

I have some [CustomElements](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Custom_Elements) (sliding menu, switch, etc.) that have methods and properties. The properties can be modified programatically by the user in some cases or just by interacting with the element with clicks or drag (e.g. `myDialog.isShown`). I'm now wrapping these elements within Vue Components (and making a Vue-like API to interact with them) and it was very handy to directly map these element properties to the Vue Component as computed properties (so `myVueComponent.$el.pages` becomes `myVueComponent.pages`). I guess I can modify the API and use methods or take the initial value and listen for any modification whenever possible, but I just thought `cache: false` made this very simple.

Issue #734, opened 2017-01-28
quick start


entaku19890818:
I am interested in the javascript framework and I want to use vue.js.
I attempted to set up referring to the official site.
There was a confusing point in the document.
https://vuejs.org/v2/guide/index.html

# Procedure

1. I installed vue.js by following the instructions on the ["Installation" page](https://vuejs.org/v2/guide/installation.html)

```
npm install vue
````

2. I attempted to move the installed vue.js in the procedure written on the  ["Introduction" page](https://vuejs.org/v2/guide/index.html#Getting-Started) .

3. However, since the ["Introduction" page](https://vuejs.org/v2/guide/index.html#Getting-Started) was a procedure to designate and use vue.js on the web,
I did not use vue.js that I dropped by installation.

# Expected behavior

I wanted it to be a procedure to use the installed vue.js.

# Possible remedy

Write code example of script tag when installing on "Introduction" page

What do you think?

**Comment by kazupon** [comment]
Thank you for your feedback! 
I think we should only need to improve about phrases at the introduction page, due to already mention about the installation as follows.

```
... Or you can simply create an .html file and include Vue with:

<script src="https://unpkg.com/vue/dist/vue.js"></script>
```

chrisvfritz:
As @kazupon mentioned, we _do_ mention how to include Vue the way you're hoping on the introduction and installation pages (it even appears in a separate section, before the separate instructions to use via NPM). I haven't encountered anyone having this particular problem before, so I'm going to close for now, but am happy to reopen if this appears to be a recurring problem with multiple people.

PR #735, opened 2017-01-28
add sample code to Getting-Started #734

entaku19890818:
I tried to fix with [my suggested improvement plan](https://github.com/vuejs/vuejs.org/issues/734).
How is it?


chrisvfritz:
As per [my comment here](https://github.com/vuejs/vuejs.org/issues/734#issuecomment-275891807), I don't think this is necessary. For your use case, I simply wouldn't use NPM. If you follow the instructions in one of the previous sections (_Standalone_ or _CDN_), then you should be fine.

PR #741, opened 2017-01-31
Accept Jamatto micropayment donations

jimmejardine:
Evan - if you'd like to accept one-click micropayments on your site, i have made the few changes to your html for you to do so.  To see how easy it is to pay, check out jamatto.com and do a quick sample purchase.

### DIFF
--- a/src/support-vuejs/index.md
+++ b/src/support-vuejs/index.md
@@ -10,7 +10,7 @@
   <span>Donate via PayPal (one time)</span>
 </a>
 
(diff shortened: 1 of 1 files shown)


yyx990803:
Hey, thanks for the PR, but I really prefer keeping donations centralized in PayPal so that I have an easier time for book-keeping.

jimmejardine:
Understood - was just thinking you might find that more than 10x people pay you 50c than do $5.  You would receive your donations aggregated up in monthly batches, so it might not be too much additional burden on the book keeping.

PR #745, opened 2017-02-01
Update example of currency-input component

termosa:
Update the example with more reactivity that is given by Vue.js
The topic about computed properties is above, so readers should be familiar with it at this point.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -569,32 +569,33 @@ Let's see it in action with a very simple currency input:
 Vue.component('currency-input', {
   template: '\
     <span>\
-      $\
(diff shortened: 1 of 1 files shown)


termosa:
Here is CodePen to make sure it works the same: https://codepen.io/termosa/pen/MJQEwq

chrisvfritz:
Sorry, I don't see this as an improvement. 😕  It just uses a more obscure feature (computed setters) instead of a method, which I think distracts from what we're really trying to demonstrate. The computed setter does allow the value to be used with `v-model`, but that also doesn't really matter in this case, since the data is only bound once.

PR #746, opened 2017-02-02
Change link to the broken example

termosa:
Change link to JsFiddle that was directed to the broken example with the new link to the fixed example

### DIFF
--- a/src/v2/guide/transitioning-state.md
+++ b/src/v2/guide/transitioning-state.md
@@ -390,7 +390,7 @@ function generatePoints (stats) {
 </style>
 {% endraw %}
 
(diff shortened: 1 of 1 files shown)


termosa:
The example is broken because of the broken link to the Vue.js in External Resources section.

chrisvfritz:
Thanks for pointing this out! We prefer to just update the base version rather than replacing links, in case the fiddle is also bookmarked or used in other documents and blog posts, etc. So that's the only reason I'm closing. 🙂  The fiddle has been updated now.

termosa:
Right choice! I'll update Russian translation.

PR #752, opened 2017-02-07
Detail Explanation of Proxied Objects

clintonyeb:
Updated the 'Properties and Methods' section to include what I think, will help newcomers understand proxied properties better. The explanation did not seem clear to me.


chrisvfritz:
I agree there's room for improvement here! The explanation you provided isn't _quite_ accurate, but I'll spend some time today working on an overhaul of this page in general. In the meantime, I'll keep this open.

clintonyeb:
Okay thanks for your feedback :)

PR #756, opened 2017-02-09
Output wrong values on page

baberuth22:


chrisvfritz:
> Wait! Shouldn't the value equal the expression if a string is passed, not the argument.

This seems like more of a question than a pull request, but the answer is no. The expression is the actual code that returns a value, while the value is what's returned _by that_ expression. Also note that it does not equal the argument, but I can understand that confusion since they are named very similarly here. I just updated the example to make it more clear.

PR #757, opened 2017-02-10
Fix typo: 'is' => 'id'

jipe:


Alex-Sokolov:
But `is` was correct!
It's not multiple li-items with identical id
It is a component which described below and named `Vue.component('todo-item',...`

znck:
Oh! My bad. Component is defined in later part. (Was not visible in snippet)

Nice catch @Alex-Sokolov 👍 

/ping @kazupon Can you revert this (accept #758)?

jipe:
Sorry guys, I jumped the gun. When reading on through the components section, this all made sense - and of course I should have thought it through that you were not trying to create multiple elements with the same id :-). Keep up the good work!

PR #759, opened 2017-02-11
Replace statement by expression

### LINKED ISSUE #754: Statement is a same word for « Instruction » or « Declaration » ?
I work on French translation of Vue.

All was clear from `guide/syntax.md` but with `event.md` I'm not really sure to have well understand.

## This is my current comprehension ##

In the document `syntax.md` at line 74, we can read

> `<!-- this is a statement, not an expression: -->` 

to describe this `{{ var a = 1 }}`

For Linguee, `Statement` means `Declaration` : http://www.linguee.fr/francais-anglais/search?source=anglais&query=statement
For Google Translate, `Statement` means `Declaration` : https://translate.google.fr/?um=1&ie=UTF-8&hl=fr&client=tw-ob#en/fr/statement

So I understant this :

`Instruction` is `Instruction`
`Statement` is `Declaration`
`Expression` is `Expression`

So, I can describe this :

```js
var num = 20 + true, 
    bool;
str = "Hello World"; 
delete str; 
if (num === 21) { 
    bool = true;
} else {
    bool = false;
    num = 0;
}
```

like this :

```
┌──────────────────────────────────────────────────┬───────────────────────────────┬───────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Instruction                                      │ Instruction                   │ Instruction       │ Instruction                                                                                                       │
├──────────────────────────────────────────────────┼──────────────────────────┬────┼──────────────┬────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│                                                  │                          │    │              │    │ Control Flow                                                                                                      │
│                                                  │                          │    │              │    ├────────────────────────────────────────────────────────┬──────────────────────────────────────────────────────────┤
│ Statement (declaration)                          │                          │    │              │    │ Block statement                                        │ Block statement                                          │
├─────┬───────────────────────────────────────┬────┤                          │    │              │    ├──────┬────────────────┬─────┬──────────────────────┬───┼────────┬───────────────────────┬─────────────────────┬───┤
│     │ Expression                            │    │ Expression               │    │ Expression   │    │      │ Expression     │     │ Instruction          │   │        │ Instruction           │ Instruction         │   │
│     ├─────┬────┬────┬────┬──────┬────┬──────┤    ├─────┬────┬───────────────┤    ├────────┬─────┤    │      ├─────┬─────┬────┤     ├─────────────────┬────┤   │        ├──────────────────┬────┼────────────────┬────┤   │
│     │     │    │    │    │      │    │      │    │     │    │               │    │        │     │    │      │     │     │    │     │ Expression      │    │   │        │ Expression       │    │ Expression     │    │   │
│     │     │    │    │    │      │    │      │    │     │    │               │    │        │     │    │      │     │     │    │     ├─────┬────┬──────┤    │   │        ├──────┬────┬──────┤    ├─────┬────┬─────┤    │   │
│     │ var │ op │ od │ op │ od   │ op │ var  │ sc │ prp │ op │ operand       │ sc │ op     │ var │ sc │      │ od  │ op  │ od │     │ var │ op │ od   │ sc │   │        │ var  │ op │ od   │ sc │ var │ op │ od  │ sc │   │
│     │     │    │    │    │      │    │      │    │     │    │               │    │        │     │    │      │     │     │    │     │     │    │      │    │   │        │      │    │      │    │     │    │     │    │   │
│ var   num   =    20   +    true   ,    bool   ;    str   =    "Hello World"   ;    delete   str   ;    if (   num   ===   21   ) {   bool  =    true   ;    }   else {   bool   =    true   ;    num   =    0     ;    } │
└─────┴─────┴────┴────┴────┴──────┴────┴──────┴────┴─────┴────┴───────────────┴────┴────────┴─────┴────┴──────┴─────┴─────┴────┴─────┴─────────────────┴────┴───┴────────┴──────────────────┴────┴────────────────┴────┴───┘
```

and that means

**These lines are one « Expressions » and work**

```html
<!-- data: { num: NaN, bool: false } -->
{{ num = 20 + true, bool }} <!-- 21 -- >
<div v-if="num = 20 + true, bool"></div> <!-- 21 evaluated to true so this is injected in DOM -- >
```

```html
<!-- data: { str: "" } -->
{{ str = "Hello World" }} <!-- "Hello World" -- >
<div v-if="str = 'Hello World'"></div> <!-- "Hello World" evaluated to true so this is injected in DOM -- >
```

```html
<!-- data: { str: "" } -->
{{ delete str }} <!-- true -- >
<div v-if="delete str"></div> <!-- true evaluated to true so this is injected in DOM -- >
```

and same for `num === 21`, `bool = true` and `num = 0`

That means also

_These lines are not only « expressions » but « declaration » and « controle flow » and **don't work**_

```html
<!-- data: { num: NaN, bool: false } -->
{{ var num = 20 + true, bool }} <!-- avoid using JavaScript keyword as property name: "var" -- >
<div v-if="var num = 20 + true, bool"></div> <!-- avoid using JavaScript keyword as property name: "var" -- >
```
```html
<!-- data: { num: NaN, bool: false } -->
{{ if (num === 21) { bool = true; } else { bool = true; num = 0 ; } }} <!-- avoid using JavaScript keyword as property name: "if" -- >
<div v-if="if (num === 21) { bool = true; } else { bool = true; num = 0 ; }"></div> <!-- avoid using JavaScript keyword as property name: "if" -- >
```

All is clear for me

BUT

## Maybe I'm not really understand Statement word ? ##

In the document `guide/events.md` at line 107, we can read

> Instead of binding directly to a method name, we can also use methods in an inline JavaScript statement:

to describe this `<button v-on:click="say('hi')">Say hi</button>`

and at line 144 in the same document we can read

> Sometimes we also need to access the original DOM event in an inline statement handler

to describe this `<button v-on:click="warn('Form cannot be submitted yet.', $event)">Submit</button>`

**So if I refer to my previous explaination/comprehension**, `say('hi')` and `warn('Form cannot be submitted yet.', $event)` are not _statement_ but _expression_**

So I think. Maybe the correct line are

> Instead of binding directly to a method name, we can also use methods in an inline JavaScript **expression**:
> Sometimes we also need to access the original DOM event in an inline **expression** handler

or 

> Instead of binding directly to a method name, we can also use methods in an inline JavaScript **instruction**:
> Sometimes we also need to access the original DOM event in an inline **instruction** handler

Because I'm not sure there is an error in two different place I think « maybe statement means instruction and not declaration »

So I ask to Google `statement javascript` and I see in first result this (http://www.w3schools.com/js/js_statements.asp) :

> JavaScript Statements

```js
document.getElementById("demo").innerHTML = "Hello Dolly.";
```

and I say « No ! It's not a statement, it's an instruction that only contain an « expression » (an affectation), not a « declaration (statement) ».

## But where is the thuth ? ##

So if I decide to say « ok, statement is not declaration, but instruction in general »

In this case the 

> `<!-- this is a statement, not an expression: -->`

should be

> `<!-- this is a declaration, not an expression: -->`

> **Someone for help me to really correctly understand ?**

posva:
Closes #754

### DIFF
--- a/src/v2/guide/events.md
+++ b/src/v2/guide/events.md
@@ -104,7 +104,7 @@ var example2 = new Vue({
 
 ## Methods in Inline Handlers
 
-Instead of binding directly to a method name, we can also use methods in an inline JavaScript statement:
+Instead of binding directly to a method name, we can also use methods in an inline JavaScript expression:
 
 ``` html
 <div id="example-3">


chrisvfritz:
Well, these _are_ actually statements rather than expressions. An expression will always return a value, whereas a statement is just executed code which does not necessarily return anything. In [this example](https://jsfiddle.net/wh0boo8g/) for instance, the inline `if` statement does not return anything, but it still works within `v-on`.

For some other directives, like `v-if`, we _do_ require an expression because we use the returned value to check for truthiness.

Does that make sense?

posva:
Oh, wow, didn't know that haha
Yes, makes sense!

chrisvfritz:
No worries! 🙂  There's so much jargon in computer science and the names are often very poor.

- Declaration
- Statement
- Expression

The meaning of these words is almost identical in human languages, so they seem like very poor choices to describe different things when programming. "Statement" is so general as to be almost useless and "expression" doesn't at all describe what it is. 😂  It should be called a "returnable" or "value producer" or something.

As the songwriter / amateur philosopher Harry Chapin once said:

> Sometimes words can serve me well. Sometimes words can go to hell.

😛

PR #762, opened 2017-02-15
Index is undefined.

vishalkakadiya:
Fix issue of index is undefined, as every time first element was deleted not the current one.


simplesmiler:
This appears to be PR to v1 branch. In Vue 1 the example with `$index` is valid.

vishalkakadiya:
@simplesmiler 
Sorry, I am checking with v2 version and checking docs for v1 site. Thanks for looking in to it.

PR #768, opened 2017-02-17
Use latest version on jsDelivr

DiegoRBaquero:
jsDelivr supports latest tag, it's not updated as instantly as unpkg but should do the trick.


jimaek:
The new backend is live. It pulls files directly from npm and github https://github.com/jsdelivr/jsdelivr#usage

Example https://cdn.jsdelivr.net/npm/jquery@3.1.1
Example directory listing https://cdn.jsdelivr.net/npm/jquery@3.1.1/

Issue #771, opened 2017-02-18
Any plan on service-worker for offline usage?


JLHwung:
The [hexo-offline](https://github.com/JLHwung/hexo-offline) supports using `sw-precache` to generate a service worker for offline usage on any hexo site.

[Browser Support](https://jakearchibald.github.io/isserviceworkerready/)

chrisvfritz:
As I just [mentioned here](https://github.com/vuejs/vuejs.org/pull/772), I'm definitely not opposed to adding offline support. 👍

gogamwar:
I tried [hexo-offline](https://github.com/JLHwung/hexo-offline) with the following configuration:
```yaml
offline:
  maximumFileSizeToCacheInBytes: 10485760
  staticFileGlobs:
    - public/**/*.{js,html,css,png,jpg,gif,svg,eot,ttf,woff}
    - public/**/**/*.{js,html,css,png,jpg,gif,svg,eot,ttf,woff}
    - public/manifest.json
    - public/browserconfig.xml
  stripPrefix: public
```
In general, it works, but I was confused by these errors. I also was not able to properly configure caching CDN scripts and Google fonts, it works only through the browser cache.
![Network errors](https://cloud.githubusercontent.com/assets/13474134/23102247/0236665c-f6b7-11e6-86ac-2b604fcf4ea9.png)

JLHwung:
CDN scripts should be cached using `runtimeCaching`, see https://github.com/GoogleChrome/sw-precache#runtimecaching-arrayobject

I will document the usage of `runtimeCaching` on `hexo-offline` later.

I cannot reproduce the error using the given configuration.

gogamwar:
@JLHwung  I tried using runtimeCaching, but I could not achieve results without the use of regular expressions.
I think you can not reproduce the error since it occurs because of files that I generated for use with manifest.json (#772).
Maybe you try yourself and send a request? It would be great :flushed:

JLHwung:
Hi morigs,

I've updated the [README](https://github.com/JLHwung/hexo-offline/blob/16cb3b85aec1eaf4d7d21c822c0456ae42526ff7/README.md) section to include CDN usage. The `runtimeCaching` works on the demo site.

I am happy to guide through, don't hesitate to contact me if there is any questions.

chrisvfritz:
Fixed in #918

Issue #769, opened 2017-02-17
manifest.json


morigs:
I think it's need to add manifest.json for ability to add site to the homescreen. But the question is where to put icons and manifest. And how to modify `_config.yml` :tired_face:

Also I think `#4fc08d` is good for `theme_color`

JLHwung:
> (where to put icons) and manifest

The `manifest.json` can be placed to `src/`.

> where to put icons

The icon is located in `themes/vue/source/images`.

> `And how to modify _config.yml`

It is unnecessary to modify `_config.yml`. If you mean referencing the `manifest.json`, you can revise `themes/vue/layout/layout.ejs`.

gogamwar:
@JLHwung Thanks for reply. I'll try this today and if all is OK send PR

gogamwar:
Omg, I forgot one import thing. This requires https. And Service Workers too (#771).
In my opinion the lack of https on the official website is a real problem.
https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/

chrisvfritz:
@gogamwar Hm, I'm not sure I see what you mean by "not by default". When I try to visit the site without SSL, I'm correctly redirected.

gogamwar:
@chrisvfritz It's really strange. On my Chrome for Android and Firefox on Ubuntu it's correctly redirect me too. But on Chrome on Ubuntu/Windows and Edge when I write ```vuejs.org``` and press enter:
```
Request URL:http://vuejs.org/
Request Method:GET
Status Code:200 OK
Remote Address:104.31.84.3:80
```
I can open site with SSL only by manually typing ```https://...```

chrisvfritz:
That is weird. I believe we're using HSTS for SSL, which [should be supported](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security#Browser_support) by all the browsers you listed. 😕

gogamwar:
Okay, now it's working on Chrome for Ubuntu but still does not work in Edge 38.13393.0.0 and Chrome 56.0.2924.87 for Windows. Maybe the problem is due to the browser but I don't see `Strict-Transport-Security`. :weary:
**Headers from Chrome**
<details> 
  <summary>Response headers</summary>
<pre>HTTP/1.1 200 OK
Date: Sun, 19 Feb 2017 17:03:23 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
[...]

PR #772, opened 2017-02-18
Add manifest and icons

morigs:
Added manifest and icons for ability to install web app (#769).
splash screenweb app in standalone modeapp in recents

However there is problem with offline. It's need to add service workers (#771).
in devtools


chrisvfritz:
@gogamwar Thanks for starting on this! I think it'll be a great addition to the docs. 😄  I'm not opposed to adding offline support.

chrisvfritz:
Consolidated into #918.

PR #795, opened 2017-03-01
Update ssr.md

sjordan1975:
I'm assuming you meant to write "dynamic" as this tutorial is contrasting the normal dynamic use of Vuejs with SSR, a fully static server rendering solution. Prerendering then is a solution that allows the user to have an essentially dynamic site while serving a few static pages.


chrisvfritz:
No, the use here is correct. The frontend is fully static, meaning the same static files are served on every request, rather than dynamically generated by a server.

PR #807, opened 2017-03-07
Grammatical change

gauravmak:


chrisvfritz:
`passing` isn't quite correct. Just as we say a function is _passed_ arguments, components are passed props and contents.

PR #808, opened 2017-03-08
Update index.md

zrw354:
To make the documents more user-friendly, we need to break long lines in sample HTML/JavaScript code into short lines to avoid horizontal scrollbars caused by limited column width


kazupon:
ah, sorry, I saw the your comment.
LGTM.

zrw354:
@kazupon Thanks for the quick response! In this case document users have to use a horizontal scrollbar to view the hidden part of only one long line, but if we don't pay attention to source code line length sometimes many lines could be hidden, causing significant usability issues. You guys have done a great job in making Vue user-friendly. Let's keep the momentum going!

PR #822, opened 2017-03-13
functional components invisible in devtools

stevekrouse:
I spent a number of hours trying to figure out why my components weren't showing up in the chrome devtools. I think a line as simple as this would solve this problem for others.


kazupon:
Thank you for your PR!

Hmm 🤔, I seem this is `vue-devtools` issue.

/ping
@yyx990803 
Is this a specification?

yyx990803:
This is correct - functional components are not traced in devtools.

kazupon:
okay, i see.

PR #836, opened 2017-03-21
More concise way to bind props from high order component to children

dotBits:
I've found a good implementation choice to wrap a set of small components into an higher order one. Consider this composition defined in a layout:

````html
<user-card :img="imgURL" :score="number" :name="username"></user-card>
````

You can split user card in small reusable components and pass only relevant props to them without much boilerplate. This would be user-card template

````html
<template>
    <div class="user-card">
        <!-- image will receive and use :img prop-->
        <an-image v-bind.prop="$props"></an-image>
        <!-- user details will receive just :score and :name -->
        <user-details v-bind.prop="$props"></user-details>
    </div>
</template>
````


robertcasanova:
Good PR!!! Thx

dotBits:
@kazupon @chrisvfritz I've recently discovered that .prop modifier no longer works as I've experienced above and this example is still in the docs https://vuejs.org/v2/api/#v-bind

PR #848, opened 2017-03-24
Better React example

zackify:
The current react example shows part of a class based component. This example is a full, functional component that attempts to match the vue component syntax as much as possible! :)


src/v2/guide/comparison.md
```diff
@@ -104,28 +104,19 @@ In React, everything is Just JavaScript, which sounds very simple and elegant -
...
-  )
-}
+    :
+      <p>No items found.</p>
+    }
+  </div>
+)
 ```
 
 Render functions with JSX have a few advantages:
```
ads1018: You'd have to update the documentation as well as it references the render function.

ads1018: You'd have to update the documentation as well as it references the render function.

PR #849, opened 2017-03-24
Update React Example in comparison.md

btnwtn:
There's no reason that the example in comparison.md shouldn't be a functional component. I've updated the docs to reflect how the component would be written by most React devs.


btnwtn:
Closed as it's a duplicate of: https://github.com/vuejs/vuejs.org/pull/848

PR #852, opened 2017-03-28
Added link to Vuecasts in the menu and guide

introwit:
Inspired by [Laravel](laravel.com) a link to Jeffrey Way's tutorials will be the best resource for anyone who wants to start learning Vue.

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -14,6 +14,8 @@ If you are an experienced frontend developer and want to know how Vue compares t
 
 <p class="tip">The official guide assumes intermediate level knowledge of HTML, CSS, and JavaScript. If you are totally new to frontend development, it might not be the best idea to jump right into a framework as your first step - grasp the basics then come back! Prior experience with other frameworks helps, but is not required.</p>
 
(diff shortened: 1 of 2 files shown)


introwit:
@kazupon wait what? One will first land up on the vuejs 's website to learn Vue and not on the awesome Vue list , many even don't know that list exists but everyone knows vuejs.org for sure. Also the tutorials by Jeff are definitely the best one , a really great alrenative to the docs. @yyx990803 your thoughts?

yyx990803:
First, thanks for the PR.

The dropdown only contains links that are officially associated with the Vue team (except jobs, which we have a partnership with). I agree that Jeffrey's casts are excellent, but they are after all 3rd party. I'm hesitant about designating it as "the best" - I'd rather let learners explore all available ones.

introwit:
@yyx990803 thank you for taking the time and responding! Also, since JS frameworks can be learned better by seeing things in action so a visual guide alternative to the textual guide would be the best for people like me starting with Vue (I am definitely not alone) for whom the docs are a bit overwhelming and Jeff comes to the rescue there. To the people I refer Laravel , when they go the docs and see the Laracast alrenative in the docs they immediately fell in love with Laravel seeing the guide/docs actually give you the every best guidance possible. Still since you are the one making final calls so it's up to you, no issues here :) 

PS: I disagree on the term 'third party' here.

PR #854, opened 2017-03-30
Update components.md

jameschiffey:
Added a reminder that when returning a fresh object from a function, you MUST place the opening curly brace on the same line as the "return". This is a really nasty gotcha if you get it wrong!

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -190,6 +190,8 @@ data: function () {
 }
 ```
 
+**It should be noted that the opening curly brace of the fresh data object MUST be on the same line as the `return`.**
+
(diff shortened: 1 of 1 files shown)


jameschiffey:
Hi. In the code example above my change, if you move the curly brace to a new line, it doesn't work.

Like this, it will error:

`data: function () {
  return
  {
    counter: 0
  }
}`

It's a really subtle difference, but really important.

kazupon:
Hmm, 🤔 
Vue.js official documentaion is assume that User can be used generally JavaScript programming.
Vue.js official documentaion is written about Vue.js.
It should not be written about JavaScript topic due to keep clear documentation.

PR #856, opened 2017-04-01
chrome manifest.json setting for csp

netroby:


chrisvfritz:
Yes, I agree that this seems to have more to do with Chrome extension development rather than with Vue specifically, so closing.

PR #861, opened 2017-04-04
Fix caveat 2 solution in list rendering section

lukaszkarbownik:
Currently list rendering section includes a wrong solution for Caveat #2.
`vm.items.length = newLength` is not equal to `example1.items.splice(newLength)`.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -420,10 +420,10 @@ Vue.set(example1.items, indexOfItem, newValue)
 example1.items.splice(indexOfItem, 1, newValue)
 ```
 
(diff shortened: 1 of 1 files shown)


kazupon:
Thank you for your feedback!

However, This is correct ...
See the below reference:

- length
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length

- splice
https://developer.mozilla.org/ja/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

lukaszkarbownik:
Thank you for fast reply and clarification.
I have indeed misread the documentation and missed the fact there is no assignment of `splice` call result.

PR #864, opened 2017-04-07
Change `node` to `Node.js` to avoid misunderstood.

Haeresis:
I participate to French translation and sometime node refers to a node from DOM and sometime it refers to Node.js (If I correctly understood).

I think it's a good thing to use Node.js and not just `node` to talk about Node.js

1. It's avoid non engilsh reader with few english skills to better understand Guide.

2. It's possbile to be a Front-end Expert and go into advanced section and to be a beginner with server-side Node.js techno and I think it's a good point to call a spade a spade.

3. Because it's the true name of application.

I understand because unix command is `node` it's a perfect shortcut and into a forum dedicated to Node.js it's a good shorthand but in the context of Vue it's a good thing to not use it.

### DIFF
--- a/src/v2/guide/ssr.md
+++ b/src/v2/guide/ssr.md
@@ -32,7 +32,7 @@ If you're using Webpack, you can easily add prerendering with the [prerender-spa
 
 ## Hello World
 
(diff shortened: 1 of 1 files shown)


Haeresis:
Thx a lot @kazupon!

PR #888, opened 2017-04-27
change the Introduction->Composing with Components

IanYet:
change the Introduction->Composing with Components->the first demo
when define a new component, you need to create an instance of that in javascript, and then you can use it in html


chrisvfritz:
At this point in the explanation, we're just trying to introduce the concept. A few paragraphs down, a more full example is provided.

PR #893, opened 2017-04-30
Add ES6 string template example to Vue template docs

crabmusket:
I didn't realise Vue supported ES6 string interpolations in templates until I tried it. I hope this is an acceptable example!

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -66,6 +66,8 @@ So far we've only been binding to simple property keys in our templates. But Vue
 {{ message.split('').reverse().join('') }}
 
 <div v-bind:id="'list-' + id"></div>
+
+<div v-bind:style="`background-image: url('${imageUrl}')`"></div>
 ```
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/syntax.md
```diff
@@ -66,6 +66,8 @@ So far we've only been binding to simple property keys in our templates. But Vue
 {{ message.split('').reverse().join('') }}
 
 <div v-bind:id="'list-' + id"></div>
+
+<div v-bind:style="`background-image: url('${imageUrl}')`"></div>
```
kazupon: Thank you for your feedback!

However, It is possible only when we are using `.vue` (vue-loader).
https://vue-loader.vuejs.org/en/features/es2015.html#es2015-in-templates

I think that we should be noted necessary as a comment about it.

crabmusket: Oh, I see! I hadn't seen vue-loader's extensive documentation. Maybe this change is unnecessary in that case.

chrisvfritz:
I definitely appreciate the thought, but I'd like to keep the essentials of the core docs as ES5. I'd consider adding something about a note to the page on Single-File Components though. 🙂

PR #896, opened 2017-05-03
Component naming convention explanation

padcom:
Added information which naming conventions explanation on which naming convention match declaration and usage


chrisvfritz:
Fantastic addition!

PR #897, opened 2017-05-03
Optimized the CSS in a transition example

jamestrenda:
Removed unnecessary CSS animation declaration and use. For this example, we can just re-use the first animation and reverse it.


chrisvfritz:
Nice! I learned something new today. 😄

PR #899, opened 2017-05-05
removed v-bind.prop example since is wrong

dotBits:
v-bind.prop no longer works as explained in that example and shouldn't be used to filter props between parent and child components


chrisvfritz:
Fixed the example rather than just removing it.

PR #903, opened 2017-05-11
`component.md` : Component Naming Conventions better consistency

Haeresis:
Hi everybody. You will find in this PR about Component Naming Conventions Guide part 4 points I propose to update:

1. **TitleCase**
There is one « Title case » term so to follow all others I change it into « TitleCase »

2. **Case vs Cased**
In first part and old second part version you used kebab-case**d**-component with letter « d ».
In second part you not use « d ». I think it's a better choice to kept consitency accross naming

3. **Opened and Closed tag in exemple**
I know it's possible to say « there is too verbose » but you have not already explain that is possible to use self-closed format `<my-component />` when you use following example
    ```html
    <kebab-cased-component />
 
    <camel-cased-component />
    <camelCasedComponent />
 
    <title-cased-component />
    <titleCasedComponent />
    <TitleCasedComponent />
    ```
    So there is two way:

     1. Use full tag declaration (in this PR)
     2. Explain more sooner that is possible to use self closed way (not in this PR)

4. **Space**
I have add a space in self-closed format to respect what you have done in part 3.


chrisvfritz:
Thanks! I made one tiny change to remove the space you added before the end of the self-closing component, but otherwise I like the changes. 🙂  Those previous spaces were before the end were something I missed from another contributor, but I really appreciate your attention to detail!

PR #905, opened 2017-05-13
Clarify v-html usage

lloy0076:
Originally, I'd thought that:

```
<div v-html="rawHtml">This {{ some_html}}</div>
```

(where we might have `vm.data.some_html='<b>hello</b>'`) ... would render something like:

**hello**

But instead we get:

```
Property or method "rawHtml" is not defined on the instance but referenced during render. Make sure to declare reactive data properties in the data option.
```

...if one is watching console errors.

My wording could be a little bit kludgy and others might be able to make it clearer but I think this may overcome some misunderstandings for those not familiar with vuejs.

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -36,7 +36,7 @@ The double mustaches interprets the data as plain text, not HTML. In order to ou
 <div v-html="rawHtml"></div>
 ```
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks! Just made a small update to the wording.

PR #909, opened 2017-05-15
Added explanation for rendering of multiple values

pverbeek:


src/v2/guide/class-and-style.md
```diff
@@ -202,3 +202,5 @@ Starting in 2.3 you can provide an array of multiple (prefixed) values to a styl
 ``` html
 <div :style="{ display: ['-webkit-box', '-ms-flexbox', 'flex'] }">
 ```
+
+This will only render the last value in the array which the browser supports. In this example, it will render `display: flex` for browsers that support the unprefixed version of flexbox.
```
kazupon: [comment](https://github.com/vuejs/vue/blob/dev/src/platforms/web/runtime/modules/style.js#L19) said:
`Set them one by one, and the browser will only set those it can recognize`

pverbeek: But doesn't it only set one value? For example, if you do `color: ['red', 'yellow', 'fake(value)']` it will only set `yellow`

kazupon: i checked multiple values.
sorry,  you're right.

DiMeneses: While I find the present explanation super confusing, the one you just gave is perfectly clear @paul-v-m 
I think it should be uptaded to: 
`color: ['red', 'yellow', 'fake(value)'] `
it will only set yellow

DiMeneses: Because, if I'm understanding it correctly, even if the browser supported prefixed versions of flexbox, `display: flex` would still be rendered, making the example ambiguous.

phanan:
I don't think this is 100% correct. IFAIK this is what happens:

1. The script sets the property to each of the array items regardless.
2. With each item, the _browser_ sets and triggers the style change. Here is where unrecognized/unsupported properties are silently ignored.

For this reason, if you observe the element, you'll see the style change is actually triggered multiple times and not once. [Demo](https://jsfiddle.net/3gLu17sv/).

PR #917, opened 2017-05-20
Add is attribute to API, fixes #804

### LINKED ISSUE #804: Add "is" to list of special attributes in API docs.
The "is" attribute, used for/with dynamic components, is currently missing in the list of special attributes.

It's documented indirectly in the Guide's section about "Dynamic components", and in the API docs about `<component>`, but I think that can be easy to miss if somebody is looking for it when he/she finds something like `<tbody :is="whatever">` in some code and want to look it up.

chrisvfritz:
Anything else I miss here?

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1785,7 +1785,7 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
   <!-- prop binding. "prop" must be declared in my-component. -->
   <my-component :prop="someThing"></my-component>
-  
+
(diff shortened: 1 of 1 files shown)


znck:
This is not required in single file components.

It should be mentioned, I guess.

chrisvfritz:
@znck Hmm, if someone isn't sure what's meant by "in-DOM templates", that section gives a little more information on the exact cases one has to worry about, so I think we're covered. This is something that's good to think about though and if someone opens up an issue because they were confused, we can clarify further. 🙂 👍

PR #919, opened 2017-05-21
Elaborate on passing props with JSX

zacacollier:
Seems like passing props is a common point of confusion for newcomers like myself coming from React. I had to look in at the babel plugin [example](https://github.com/vuejs/babel-plugin-transform-vue-jsx/blob/master/example/example.js) to figure out  the usage I've described above, so I thought it might be appropriate to briefly describe it in the docs.

### DIFF
--- a/src/v2/guide/render-function.md
+++ b/src/v2/guide/render-function.md
@@ -445,6 +445,49 @@ new Vue({
 
 <p class="tip">Aliasing `createElement` to `h` is a common convention you'll see in the Vue ecosystem and is actually required for JSX. If `h` is not available in the scope, your app will throw an error.</p>
 
+If you need to pass props from parent component to a child, make sure to reference them within your `render` function:
+
+``` js
+import AnchoredHeading from './AnchoredHeading.vue'
+
+new Vue({
+  el: '#demo',
+  render (h) {
+    const data = {
+      level: 1
+    }
+    return (
+      <AnchoredHeading level={data.level}>
+        <span>Hello</span> world!
+      </AnchoredHeading>
+    )
+  }
+})
+```
+
+If the parent in question is receiving the props which you'd like to pass down to a child component, simply declare them as you normally would, remembering to reference them within `render`:
+
+``` js
+import AnchoredHeading from './AnchoredHeading.vue'
+
+new Vue({
+  el: '#demo',
+  props: {
+    level: Number
+  },
+  render (h) {
+    const data = {
+      level: this.level
+    }
+    return (
+      <AnchoredHeading level={data.level}>
+        <span>Hello</span> world!
+      </AnchoredHeading>
+    )
+  }
+})
+```
+
 For more on how JSX maps to JavaScript, see the [usage docs](https://github.com/vuejs/babel-plugin-transform-vue-jsx#usage).
 
 ## Functional Components


chrisvfritz:
Thanks for reporting your confusion! Did you see [this section](https://vuejs.org/v2/guide/render-function.html#The-Data-Object-In-Depth)? If you were unable to find that page, more information on the search path you used would be helpful. If you did find it and it wasn't helpful, I'd love more information on which questions it didn't answer. 🙂

zacacollier:
Hi Chris, thanks for your quick follow-up! In my first attempt, I did read that section, however I didn't really connect the dots until you suggested it just now.

Looking back at my search history, I googled "pass props vue jsx", which led me to both the [babel plugin](https://github.com/vuejs/babel-plugin-transform-vue-jsx) and the [Render Functions](https://vuejs.org/v2/guide/render-function.html) section in the documentation.

I'd say the source of my confusion stems from how I'm accustomed to passing and accessing props in React, and trying to find out how to achieve a similar development pattern with Vue components. It wasn't immediately apparent that props declared in the component aren't directly accessible within a `render` function, that is, not until they're called within `render` itself.

This isn't to say that this discrepancy isn't covered in the documentation. I'm very new to Vue, having only used it for a few days now, so I may have missed on the first go-around.

This is the thing I've been hacking on:
```js
export default {
  name: 'Posts',
  props: {
    posts: Array,
    msg: String
  },
  render (h) {
    const data = {
      posts: this.posts
    }
    return (
      <div>
        {
          data.posts.map((post, index) =>
            <Post key={index}
              author={post.author}
              title={post.title}
              tags={post.tags}
              date={post.date}
            />
          )
        }
      </div>
    )
  }
}
```

At first, I tried to call `posts` within the render function using `this.props.posts`. Since, in my Vue component, I was declaring props on the component, and seeing them register in the DevTools, I was under the impression that I could call `props` as I would in a class-based React Component - that is, from the instance itself:

```js
class Posts extends React.Component {
    render() {
        return (
          <div>
            {
              this.props.posts.map((post, index) =>
                <Post key={index}
                  author={post.author}
                  title={post.title}
                  tags={post.tags}
                  date={post.date}
                />
              )
            }
         </div>
        )
    }
}
```

 When that didn't work, I tried passing in `props` as a second argument to the `render` function (reading the babel plugin's usage docs showed me the problem there).

I finally figured out that I have to explicitly reference props before calling them, by defining a `data` object within a `render` method as in the [example from the usage docs](https://github.com/vuejs/babel-plugin-transform-vue-jsx/blob/master/example/example.js).

Tldr; the source of my confusion is that with JSX in Vue, there seem to be a couple extra hoops to jump through as you pass props, compared with the fast-and-loose way of passing props in React. Not that I'm complaining - in fact, I really like how Vue requires you to explicitly declare props on the component before you can use them - it reinforces good type-checking habits during development. The extra hoops themselves weren't immediately apparent after my initial reading of the documentation. Perhaps a FAQ along the lines of "Migrating to Vue from React" would be in order? 

Sorry for the lengthy reply. Let me know if you need any additional clarification from me, as well as any edits you'd like to see in my PR.

chrisvfritz:
Ah, so the solution is a bit simpler than you're thinking. While the posts are not scoped under `this.props` like in React, they are still readily available within JSX the same way you accessed them  in your example outside of JSX: directly on the instance - just like data, computed properties, and methods. Try this:

``` jsx
export default {
  name: 'Posts',
  props: {
    posts: Array,
    msg: String
  },
  render (h) {
    return (
      <div>
        {
          this.posts.map((post, index) =>
            <Post key={index}
              author={post.author}
              title={post.title}
              tags={post.tags}
              date={post.date}
            />
          )
        }
      </div>
    )
  }
}
```

zacacollier:
ah, I had a sneaking suspicion that it was simpler than I thought. Cleverness strikes again. Much appreciated!

PR #921, opened 2017-05-23
A drop-dead happened in animate function, and then I fixed it.

longze:


chrisvfritz:
I'm afraid I'm not familiar with the term "drop-dead" - what is it? The usage of Tween is based on examples from their docs, so I think it was correct. Do you have an explanation for what was wrong with it and how your PR solves the problem?

longze:
I'm sorry for my English, "drop-dead" it means that the function of animate will be  execute and not stop for ever. You can get the recurrence, if you write `console.log(1)` in function of animate. I thank the example of Tween's docs is the simplest, But it dot suit in product.

chrisvfritz:
Thanks for the explanation! I'll investigate further and then probably merge this after making some small changes.

PR #928, opened 2017-05-26
Added note for render functions for pre-compiled templates

DKhalil:


chrisvfritz:
This is not actually true. When using a runtime-only build, it's necessary to only use templates in .vue files. For the standalone build, templates can be used anywhere.

PR #932, opened 2017-06-02
Docs EN : Webpack to webpack

Haeresis:
webpack is used without majuscule (and even in starting of phrase) : http://webpack.github.io/docs/

   - — see https://webpack.js.org/configuration/ for phrase majuscule example 
   - — see https://webpack.js.org/development/release-process/ for link/title example)


chrisvfritz:
You are definitely correct that it's the convention used within Webpack's docs, but it _does_ feel strange to me. 😕  More generally in open source, kebab-case (or snake_case in some communities) seems to be the convention when referring to a _package_ (e.g. in the context of install instructions), while Title Case is the convention when referring to projects and ecosystems. At least, this is what I _believe_ I've observed.

Since Webpack's internal convention breaks what I think might be a more general convention, I'm tempted to stick with the more general convention. I worry that lower-cased "webpack" next to "Browserify" and "Rollup" (for example) might look like a mistake and be off-putting to readers -especially at the beginning of a sentence.

Does that make sense? What do others think?

Haeresis:
Hi @chrisvfritz!

I totally understand what you explained because I have same conventions. Me too, if I follow my ego (the little voice that said, my believe is thruth and not others) I will say:

« Starting a sentence without a capital is not something authorised! » or « All nouns must begin by a capital! »

But I try to see the scene from an external point of view and understand more globaly if it's just my ego that speak. In the « 2.x » world, the case of name is important.

If someone write « Jquery » and not « jQuery » I think it's not professional, it's a person « not structured » / « not organized » / « not respectful » that write that. So it's possible others professionals think that when the Vue.js documentation is red.

Same if I see « Iphone » and not « iPhone » (https://www.apple.com/fr/iphone), « NPM » and not « npm » (https://www.npmjs.com/), « Javascript » and not « JavaScript » (https://en.wikipedia.org/wiki/JavaScript), « MicroSoft » and not « Microsoft » (https://www.microsoft.com/), « NodeJS » and not « Node.js », « Angular.js » and not « AngularJS » (https://angularjs.org/), « VueJS » and not « Vue.js » (or Vue) etc. 

So why because « webpack » not following some « Webpack », « wEbpack » or « WebPack » conventions (I have usually see) I will add a first capital?

If someone write my name « bruno » and not « Bruno », I have the same feeling of « not professional / not respectful » behavior so, if all occurences of webpack, included the name on project itself, are in lowercase, I think a respectful, profesionnal and structured behavior is to write « webpack ».

In my personal case I have a npm/github Node.js MVC framework called « NodeAtlas » with the `node-atlas` package name or CLI command (that's following the convention you have mention!) and it's also my conventions so I understand your point.

Ultimatly, it's your decision so if you choose to kept Webpack, the world will not collapse! So it's you that decide and that know better than me the english culture.

For french culture, « respectful » is important to allows credit of something so we will choose « webpack » in french translation :)

Thank you very, very much for taking time to respond to my PR

Très cordialement :)

chrisvfritz:
@Haeresis Outside of our own documentation, I don't think I mind whether people write Vue.js, VueJS, vuejs, Vue, or vue. 🤷‍♂️ 

I can understand your points though and it sounds like for you and some others, seeing "Webpack" with a capital "W" would be distracting. That's definitely something I'm concerned about, as I actually have the same (but opposite) concern: I don't want to distract readers with an exception to a rule followed in the rest of the core documentation, more generally in the software community, and by English grammar at the beginning of sentences.

As you point out, we each have our cultural biases and I agree with your decision for the French translation. 👍  Since the English docs have a more international readership and we have a very international core team and community, I'll open this up to a vote, so we can measure which would be the least distracting to most people. Once we have 10 votes, I'm happy to go whichever direction is decided. 🙂 

### Voting instructions

- 👍  **Change all instances of "Webpack" to "webpack", regardless of context**
- 👎  **Keep "Webpack" capitalized when used in a project context**

chrisvfritz:
Wow, 5-5! I guess in the case of a tie, I'll selfishly choose the one I prefer. 😅  I'm potentially open to revisiting this in the future though.

Haeresis:
For example, you use already « axios » without majuscule same as documentation : https://github.com/mzabriskie/axios#axios-api

PR #937, opened 2017-06-06
Clarify only custom event listeners are removed with `vm.$off`

MadLittleMods:
Clarify only custom event listeners are removed with `vm.$off`

From discussion here https://github.com/vuejs/vue/issues/5829#issuecomment-306400930


kazupon:
Thanks!

PR #939, opened 2017-06-07
Add note regarding explicit prop declaration requirement

KKSzymanowski:
When customizing the `v-model` behaviour, the new prop name has to be explicitly defined in the props list. This adds a note about it to the docs.

Regarding https://github.com/vuejs/vue/issues/5834


chrisvfritz:
Thanks!

Issue #941, opened 2017-06-07
Explicitly Create index.html File


alexfinnarn:
In the https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/index.md#declarative-rendering example, I only saw `{{message}}` until I added a document ready function wrapper...something like 

```
document.addEventListener('DOMContentLoaded', function(){
  var app = new Vue({
    el: '#app',
    data: {
      message: 'Hello Vue!'
    }
  });
}, false);
```

I'm not sure if I'm just doing something wrong, but whatever the case, more notes need to be added for my failure or the code example needs to check if the DOM is fully loaded.

alexfinnarn:
I basically just told them explicitly what .html file to make and what they could copy/paste into that to follow along.

chrisvfritz:
As stated in #943, I made a compromise with a download link for `index.html`. I think that should resolve the problem. 🙂

PR #943, opened 2017-06-07
Issue #941: Explicitly Create index.html File

alexfinnarn:


chrisvfritz:
I like the direction you're going: making the intro simpler easier for those who want to get started with an HTML file. I _would_ like the keep the example focused on just Vue though. As a compromise, I added a download link for a starting HTML file. 🙂

PR #947, opened 2017-06-09
Create list.md

mPyth:
added 'this' in the method near to the end of the file


chrisvfritz:
`numbers` is actually an argument to the `even` method, so this is correct as it is. 🙂

PR #948, opened 2017-06-10
Add explanation for implicit component attribute passing

chrisvfritz:
From https://github.com/vuejs/vue/issues/5846.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -436,7 +436,9 @@ When a prop validation fails, Vue will produce a console warning (if using the d
 
 ## Non-Prop Attributes
 
-While explicitly defined props are preferred for passing information to a child component, authors of component libraries can't always foresee the contexts in which their components might be used. That's why components can accept arbitrary attributes, just like HTML elements.
+A non-prop attribute is an attribute that is passed to a component, but does not have a corresponding prop defined.
+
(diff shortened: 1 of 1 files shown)


chrisvfritz:
@posva Good call. 😄  I was looking [at a cat](https://en.wikipedia.org/wiki/Calico_cat) when making up that example. Updated!

src/v2/guide/components.md
```diff
@@ -434,6 +434,44 @@ In addition, `type` can also be a custom constructor function and the assertion
 
 When a prop validation fails, Vue will produce a console warning (if using the development build).
 
+## Non-Prop Attributes
+
```
Austio: 👋 Hi, vue user here and followed to this pr from the issue discussion vuejs repo.

Have you considered adding a bottom line up front statement of the behavior and then the detail below that?  For instance, maybe a sentence here with something like.

```
A non-prop attribute is an attribute that is passed to a component that does not have a corresponding prop defined.

Non-prop attributes are passed as attributes to the root element of the component they are sent to.
```

Austio: btw, the documentation site is absolutely incredible, thank you for making it.

chrisvfritz: @Austio Great thinking. 🙂  I just updated the explanation.

PR #949, opened 2017-06-11
fix: composing with components in guide/v2

PibeDx:
- add new Vue(...) in sample


chrisvfritz:
We only register apps on this page for rendered examples. In those code snippets, we're simply introducing the idea of components. Just below them, we complete the idea and show the complete app that you're looking for. 🙂

PR #954, opened 2017-06-14
Riot comparison - VueJS interacts better with inline JS

DuskoV:
- Riot doesn't allow direct access to components/tags from inline JS without using complex mixins. Vue has straight-forward approach.

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -224,5 +224,6 @@ Riot 2.0 provides a similar component-based development model (which is called a
 
 - [Transition effect system](transitions.html). Riot has none.
 - A far more powerful router. Riot’s routing API is extremely minimal.
+- Riot doesn't allow direct access to components/tags from inline JS without using complex mixins. Vue has straight-forward approach.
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I'm not personally familiar with trying to use this feature in Riot. If others have used both Vue and Riot and found this to be true (or not true), please comment! 🙂

DuskoV:
I experienced it. For example, if I make AJAX call from regular JS, and want to inject JSON in data component, I can do ot easy in Vue. Riot is another story (need mixin, etc).

chrisvfritz:
@DuskoV We simply want to be careful about adding anything to this page that we don't have first-hand experience with (or clear community consensus on). If you created a couple JSFiddles demonstrating the difference, then I could play around with them and do a little extra research to validate the claim.

DuskoV:
Here you go, Chris:
https://jsfiddle.net/oys8avje/

- Vue allows direct access out from scope (from inline JS)
- Riot requires mixins


Hope this helps.

chrisvfritz:
Hmm, even though I understand what you're saying now and do agree, I worry the use case is a little difficult to explain in a short amount of text and also not encountered that frequently. I do appreciate you helping me learn more about Riot, but I worry about sounding nitpicky by including a (relatively) niche feature like this. It's also introduced very early in the Vue docs, so I think there's less risk in someone missing it.

PR #956, opened 2017-06-15
`render-function.md` — tick consistency

Haeresis:
- tick consistency to highlight code part in text comment


chrisvfritz:
Hmm, I'm wondering if we want to go the other way and consistently remove backticks from comments. For those not familiar with markdown, they could be confusing. What are others' thoughts?

Haeresis:
Hi @chrisvfritz !

This is my throught.

In a normal text (.txt) all you right is only ponctuation. So if you want highlight something important, one of the way is to place some « " » (in english). Ex 👍 

```js
/*...*/
{
  route, // the ES6 « route » property is the same as « route: route » in ES5. We recommand « ES6 » format as Bob said « ES6 is better than ES5 because 6 is after 5 ».
  data: () => {}
}
/*...*/
```

But this is really poor to indicate the different level of informations.

That's why for web presentation some `<strong>`, `<samp>`, `<em>`, `<var>` etc. tag was created.

But the main propose of Markdown is to « allows plain text to be **readable** and in same time to allows all this level of conplexity « without » use parsing to create `.something` from `.md` (e. g. `.htm`).

One of the capability of that, is Markdown is _easy_ to **write**, _easy_ to **read** and _easy_ to **transform** because text contains differents level of importance.

---------

I understand your problematic as following :

« Because our actual process not allows us to « parse » Markdown inside code comment, maybe it is a better choice to not set into ».

But for exemple, JSDoc parser to allows some comment to be a part of API documentation generated transform markdown into valid HTML for a webdoc (or maybe valid Word tag for a `.docx` doc, etc.).

So I think :
- because markdown was create to allows more readibility to plain text, and a comment **is** a plain text : **is a good idea to use it**.
- it's not because our current tool do not allows parse Markdown that we must not used it.

So, in french documentation we use the following rules:
- if the part of text is a **valid JavaScript/HTML/what-you-want code**, we surrounded it with tick.
- if the part of text is a code-like syntax that's only serve an example, we do nothing. e. g. { Array<String> }
Because it's what we have in majority observed accross all vue, vue-loader, vue-router, vue-server-renderer, vuex, etc. documentation.

So the previous code could become : 

```js
/*...*/
{
  route, // the ES6 `route` property is the same as `route: route` in ES5. We recommand **ES6** format as Bob said _ES6 is better than ES5 because 6 is after 5_.
  data: () => {}
}
/*...*/
```

that bring more level of information than :

```js
/*...*/
{
  route, // the ES6 route property is the same as route: route in ES5. We recommand ES6 format as Bob said ES6 is better than ES5 because 6 is after 5.
  data: () => {}
}
/*...*/
```

---------

Another point that not directly concern the English documentation.

For french, we have another point, (that's probably the same for language that is not in english).

If you said (speaking about a variable called `children`):

```js
// Use children is more important`
```

Because in french the word `children` become `enfants`, we prefer clearly identify that's a part of code and that's a part of plain english word because in one case we translate by

```js
// Utiliser des enfants est plus important.
```

or

// Utiliser `children` est plus important.

------

That's all :p

chrisvfritz:
You've convinced me. 😄  For people who don't understand Markdown yet, it could initially be confusing, but I now agree that it's not difficult to get used to the convention and that for those who _can_ already read markdown, it will be simpler.

Haeresis:
And *learn* Markdown is not a big difficulty and it's a great tools that all developer should have in our skills to « communicate » accross plain text document, chat, email, code comment, etc. I'm proud to say : _Vue provided it with code comment_ :P

PR #957, opened 2017-06-16
Tip: UI libraries using hidden form fields #1194

richardtallent:
Added a tip based on personal experience using SemanticUI's dropdown field, which relies on a hidden input field to store the current state.

### DIFF
--- a/src/v2/guide/forms.md
+++ b/src/v2/guide/forms.md
@@ -12,7 +12,7 @@ You can use the `v-model` directive to create two-way data bindings on form inpu
 
 <p class="tip" id="vmodel-ime-tip">For languages that require an [IME](https://en.wikipedia.org/wiki/Input_method) (Chinese, Japanese, Korean etc.), you'll notice that `v-model` doesn't get updated during IME composition. If you want to cater for these updates as well, use `input` event instead.</p>
 
-<p class="tip">Some common form UI libraries directly change the value of hidden input fields to store the current state. Since this does not raise an `onchanged` event on the element, `v-model` cannot automatically detect the changed value and update the model.</p>
+<p class="tip">Note that `v-model` listens for the `input` (default) or `change` (if you use the `lazy` modifier, more on this below) events on the element. These only fire when the *user* modifies the field. Vue cannot automatically update based on other code making changes to the element's value.</p>
 
 ### Text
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Hmm, I'm hesitant about this, because it seems more a quirk of some UI libraries, rather than having to do specifically with Vue. For example, if you were trying to detect the change with plain JavaScript, the same caveat would apply.

richardtallent:
It's a pretty common "quirk" for UI libraries that replace plain HTML elements with more stylish or functional surrogates (checkboxes, date pickers, drop-downs, radio buttons, rich text editors, etc.) because it's a convenient way to get a "drop-in" replacement "field" that still provides a value to the parent form.

This could be rewritten to better emphasize that because of the way Vue detects changes (the onchange event), that *any* outside code manipulation of an input value won't trigger this event (including but not limited to hidden fields modified by UI libraries). Would that allay your hesitation?

chrisvfritz:
Another issue is that to _truly_ understand how to handle this use case, you have to know about refs, which aren't introduced until later. I _do_ agree that it would be good to provide more guidance for integrating non-Vue UI libraries, but since this is not really a Vue-specific problem and could use a more lengthy explanation, I think a cookbook article would be better suited.

richardtallent:
Good point. I still think it bears mentioning how Vue does its v-model magic, hopefully that's enough to trigger the right thought process for someone like me trying to figure out why they can't get some fancy component working. I'll see if I can tweak the submission into something less specific to UI component integration. I may also write such an article...

richardtallent:
@chrisvfritz, I gave it another shot, is this any better for you? I guess the point of this patch is that eventually someone *will* figure out what is going on by reading the docs, but if they're like me, it will take too long for it to sink in that `el.value="new value"` (whether by your own code or some UI library) isn't going to be caught by the events Vue is listening for. I think a tip to point that out is still a Good Thing(tm), as I suspect it's a common issue for people migrating.

chrisvfritz:
@richardtallent Isn't this information already on the page though? We do mention in the introductory paragraph that "v-model is essentially syntax sugar for updating data on user input events." Since `el.value = 'new value'` is not user input, I think it makes sense that it does not trigger a user input event.

chrisvfritz:
In the cookbook, we'll be including more examples of integration with 3rd-party UI libraries, so I'm going to close this for now, as I still worry adding a note here will cause more confusion than it relieves.

PR #961, opened 2017-06-18
Update components.md

mPyth:
I've added additional few sentences which should not finish in final document, they describe the problem and there is a fix suggestion.


chrisvfritz:
@mPyth I just updated the example, so the bug should now be fixed for anything but extremely large numbers. Let me know what you think. 🙂

chrisvfritz:
I'll actually close this for now, but I'll reopen if you notice something I missed. 😄

mPyth:
Seems it now works as expected.
I'm a little curious what's the case where multiplying and dividing with 100 is needed (now 100 is replaced with 1000000000000).
AFAIK its not advisable to do such things with floating point numbers because they are quite unpredictable (if such multiplying/dividing is needed its probably better to do it with a 2^N).

PR #965, opened 2017-06-20
Update components.md

codexp:
I mistakenly thought a `<template>` could be anything, but it turns out this feature is limited to this particular tag.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -933,6 +933,8 @@ If we render the above, the output will be:
 </div>
 ```
 
+<p class="tip">Note that the slot element has to be a `<template>` tag!</p>
+
 A more typical use case for scoped slots would be a list component that allows the component consumer to customize how each item in the list should be rendered:
 
 ``` html


chrisvfritz:
@codexp I may be misunderstanding what you're trying to say, but it isn't true that slot elements can only be `template` tags, as is evidenced in examples on this page. Can you show a JSFiddle demonstrating the behavior you didn't see documented?

codexp:
just created JSFiddle and found out one can use any tag actually.
But on my project I've spent some hours trying to figure out why my implementation did not work.
Then after renaming my slot tag from <p> to <template> it immediately kicked in.
So my wrong assumption was that only template tags were allowed.

codexp:
ow, in my project I have renamed template tag (scoped slot) renamed to div and that component did not render with Vue errors in console:

codexp:
sorry closed it too early, problem still exists.

chrisvfritz:
@codexp A paragraph inside a paragraph is technically invalid HTML, so that could be part of the problem. But could you produce a very _minimal_ JSFiddle demonstrating the problem? It's much easier to figure out what's going on when looking at the smallest possible code that still demonstrates your issue.

codexp:
as the header in JSFiddle suggests: 
> try to change the &lt;template&gt; tag of the scoped slot to any tag you wish and see what happens...

you can choose any tag you like, and it brakes the vue component.
For example a div around a p tag is perfectly fine, but produces same error.

codexp:
here is the updated (minimized) JSFiddle example:

https://jsfiddle.net/hyofspsr/2/

LinusBorg:
From the docs:

> In the parent, a `<template>` element with a special attribute scope indicates that it is a template for a scoped slot.

Technically, this already does say that it has to be a `template` element, but I assume this wasn't clear enough and you want to stress that point?

codexp:
yes that's right, it could have saved a lot of time in my case, instead of trying to figure out what I have done wrong (in a quite complex setup at that moment).

chrisvfritz:
@codexp Ah, OK. Which part of the docs were you looking at, where you expected a note that could have helped you?

LinusBorg:
@chrisvfritz I think he expected it where he added it in this PR ;)

However I think we cold clarify it right in the paragraph that I quoted.

codexp:
that would suffice :)

PR #966, opened 2017-06-20
Clarify .prop modifier interacting with components

davidhewitt:
As per discussion on https://github.com/vuejs/vue/issues/5925


src/v2/api/index.md
```diff
@@ -1800,6 +1800,8 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
   ```
 
   `.camel` is not needed if you are using string templates, or compiling with `vue-loader`/`vueify`.
+  
+  If the tag is a component then `.prop` will set the property on the component's `$el`.
```
posva: Can you move it up to the `.prop` section?

davidhewitt:
Sure, have done so, how's that? Feel free to squash the commits when merging if desirable.

PR #973, opened 2017-06-26
Update docs for keep-alive include/exclude with array syntax

jkzing:
[#5956](https://github.com/vuejs/vue/pull/5956)


jkzing:
🎉

PR #977, opened 2017-06-29
Update class-and-style.md

martinnaughton:
if you do not have the single quote around the class when using the tenary operate then vuejs complains the it can not find that class. I ran in to the is problem yesterday.


chrisvfritz:
It does work, as you'll see in [this fiddle](https://jsfiddle.net/z982uk8v/). It sounds like you simply don't have a property called `activeClass` defined, but rather are trying to use a class literally called `activeClass`.

PR #986, opened 2017-07-02
Docs: awesome-vue => vue-curated

Aladdin-ADD:
there are so many people didn't know it.


chrisvfritz:
@Akryum Just double checking - does vue-curated now replace awesome-vue?

nickmessing:
I think it makes sense to add vue curated without removing awesome vue to the drop down, what do you think @Akryum, @chrisvfritz, @Aladdin-ADD?

chrisvfritz:
Done. 🙂

PR #990, opened 2017-07-03
npm is always in lowercase

Haeresis:
I know npm is the acronyme of Node Package Manager and normally an acronyme is in uppercase. But a true acronyme take also some dot N.P.M., U.R.S.S. etc. So, I think the best way to « write » npm is the way npm itself want to use it.

On https://www.npmjs.com/, you will see « npm » is written in lowercase, and same as « webpack » : at begin of a sentence, it use always a minuscule.

It's the way will use it in French transalation. We propose you to follow with rules (or not) if that make sence for you.


chrisvfritz:
Hmm, I think this is another case where I just worry it'll look like a mistake to most people - particularly at the beginning of a sentence. I just tried Googling "npm" with an exclusion rule for npmjs.com and found that the vast majority of sites also capitalize it like a normal acronym. So based on the precedent of the earlier decision, I think I'll once again break NPM's internal convention in favor of the more global convention.

> But a true acronyme take also some dot N.P.M., U.R.S.S. etc.

That's actually not true in modern English. Most acronyms do not require dots and many would even look strange with dots.

Haeresis:
Thx you both for your time!

Understood @chrisvfritz, i'm not really sure for the dot rules, even in french so no problems.

We will kept official term in our hand but I better understand the process. You check the most usage of a word done by others. In our culture will be more intersted by officials terms because if we use this method, we will think no one had always walk on the moon! ;)

chrisvfritz:
@Haeresis Thanks for bringing up these topics and forcing us to make conscious decisions!

PR #991, opened 2017-07-03
tweak profile

kazupon:
😉


chrisvfritz:
Only saw this after publishing, so updated on master. 🙂

PR #992, opened 2017-07-04
Tweak profile

znck:
😉 Switched job.


chrisvfritz:
Just saw this after publishing and so I updated on the master branch. 🙂

PR #994, opened 2017-07-06
Corrected an example in `render-function.md`

riophae:


chrisvfritz:
Nice work! I wrote this example before `v-if`/`v-else` combinations were allowed on the root and it definitely deserves an update now. 🙂

PR #995, opened 2017-07-06
Changed backslash string into ES template string.

Jinjiang:


chrisvfritz:
Unfortunately, this would break in any version of Internet Explorer, so we can't use template literals in the core docs. 😞  I find it a bit painful too, but once we start adding language features like that, then we start assuming a build step, which we try to avoid.

Jinjiang:
I also find [here](https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/installation.md#runtime--compiler-vs-runtime-only), [here](https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/installation.md#runtime--compiler-vs-runtime-only) and [here](https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/list.md#components-and-v-for) using ES template string in code snippets (but still backslash in the `{% raw %}` code to ensure the running demos work in IE).
Is that OK or we should change these above back?

chrisvfritz:
@Jinjiang Good find. I see that @yyx990803 added those, so I'll bring him into the conversation to ensure we're on the same page. My thinking was that since Vue 2 maintains IE9+ compatibility and can be easily used without a build step, that we should avoid examples that actually _require_ a build step - at least in the Essentials section. Evan, are you thinking we should adjust that policy?

chrisvfritz:
(In the meantime, I've updated those examples to normal strings for consistency. We can change them back pending the result of this discussion.)

yyx990803:
I think we should start adopting ES6 features in the docs because:

- Any up-to-date major browser support most ES6 features now, so there shouldn't be a problem when the user is following along the examples;

- Any production app probably involves a build step;

- We should still add a note at the beginning, something along the lines of "if your app targets IE and you are not using a build step, you should probably stick to ES5 only features..."

nickmessing:
@yyx990803, are you sure that's a good idea in the context of IE9 support? I'd vote for it to be done after release of Vue 3.

yyx990803:
@nickmessing that's just for documentation. I don't think the fact that we support IE9 should lock all documentation code to ES5.

nickmessing:
Yeah, makes sense

chrisvfritz:
@yyx990803 I think the multiline strings look ugly too, but they're a pretty small price to pay for supporting progressive adoption IMO. I think jumping fully into ES6 in the docs would have far more downsides.

> Any production app probably involves a build step;

I don't think this is actually true and in the cases where there _is_ a build step, I think it's usually just for concatenation and minification - not transpilation. I think there's only this perception because we're in an isolated bubble. 

For example, combining the market share of Vue, React, Angular, and Ember - it reaches a total of about 0.9% of websites according to [w3techs](https://w3techs.com/technologies), whereas [96.2%](https://w3techs.com/technologies/details/js-jquery/all/all) are using jQuery. In the jQuery pool, 91.4% of those are still on jQuery 1 - meaning they haven't updated jQuery, never mind their build process, in at least 4 years.

If this is an indication of JavaScript expertise, then I think it's safe to say that a move to ES6 would increase the learning curve by orders of magnitude for at least 80% of potential newcomers. And even providing a warning about needing to use a build step, this warning would really only be targeted at users who:

- have to support IE (which is still ~6% of traffic in the USA, so probably the vast majority),
- can't already distinguish ES6 from ES5 features, and
- aren't already using a build step that supports transpilation.

So now before using Vue for any production purpose, they'll have to either:

- probably learn what ES6 and ES5 means, then learn all the ES6 features so that they can be recognized, as well as find ES5 alternatives for each one they encounter in the docs, or
- start using a build step, including how to incorporate it into their backend.

These are non-trivial obstacles to just getting started and suddenly, Vue is much less "The Progressive Framework" it formerly was, alienating all but the elite. I'm just not sure whether making a few examples a little prettier is worth it. 😕

PR #996, opened 2017-07-07
Update transitioning-state.md

Alex-Sokolov:
There are another 4 examples with same code as changed here: https://github.com/vuejs/vuejs.org/pull/921


chrisvfritz:
Well, that's an item off my todo list. 😄  Thanks!

PR #997, opened 2017-07-08
Update page.styl

dear-lizhihua:
The content section is wider.


chrisvfritz:
The content is actually deliberately this size. I think 800px is a little too wide for optimal readability.

PR #999, opened 2017-07-09
Update list.md

Alex-Sokolov:
Forget to fix example, while fixing live-example here: https://github.com/vuejs/vuejs.org/pull/998


kazupon:
Thanks!

PR #1000, opened 2017-07-10
Removed all unnecessary colons in stylus files.

Jinjiang:
ref: https://github.com/vuejs/vuejs.org/commit/75d9525535ac4d6947d7bd2401e0dfd84e0d78f3#commitcomment-23002926

### DIFF
--- a/themes/vue/source/css/_team.styl
+++ b/themes/vue/source/css/_team.styl
@@ -2,135 +2,135 @@
 
 #team-members
   .sort-by-distance-button
-    display: inline-block
-    padding: .4em .7em .45em
-    font-weight: bold
-    font-size: .5em
-    text-transform: uppercase
-    line-height: 1
-    border: none
-    background: $medium
-    color: white
-    border-radius: 3px
-    position: relative
-    cursor: pointer
-    float: right
-    margin-top: .3em
+    display inline-block
+    padding .4em .7em .45em
+    font-weight bold
+    font-size .5em
+    text-transform uppercase
+    line-height 1
+    border none
+    background $medium
+    color white
+    border-radius 3px
+    position relative
+    cursor pointer
+    float right
+    margin-top .3em
     i
-      margin-right: .25em
+      margin-right .25em
       &:last-child
-        margin-right: 0
+        margin-right 0
     &[disabled]
-      opacity: .7
-      cursor: default
+      opacity .7
+      cursor default
   .vuer
-    display: flex
-    padding: 25px 0
-    border-bottom: 1px dotted #ddd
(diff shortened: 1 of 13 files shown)


themes/vue/source/css/_team.styl
```diff
@@ -2,135 +2,135 @@
...
+          letter-spacing .3px
+          padding 2px 5px
+          margin-left 10px
+          color transparentify($dark, #f9f7f5, .6)
+          background #f9f7f5
+          border-radius 5px
       // NOTE: Removing the little signal icon for now
       // .distance
-      //   position: relative
+      //   position relative
```
nickmessing: Removed even from comments? :smiley:

Jinjiang: Yes, they are also stylus codes. Maybe some day they would be enabled again. :-)
(Is that OK?)

nickmessing: Yeah, ofc :+1:

chrisvfritz:
My preference would actually be always colons. 😅 There are some cases in Stylus where the colon is necessary for the value to be parsed correctly, so removing all could actually be dangerous. For me personally, the colons also work better with some of my editor features.

We're also inconsistent in Stylus files between 2 spaces and 4 spaces. I much prefer 2 spaces, if we're already taking a standardization pass

Any thoughts?

nickmessing:
@chrisvfritz, I vote for 2 spaces and no colons unless necessary (like javascript semicolons rule) but I moved from stylus to postcss so I guess my opinion is not really worth a lot :smile:

Jinjiang:
I vote for 2 spaces.

For colons or not, I accept any of them but not both of them together. Go all stylus with colons is OK for me if it is much safer and more friendly to editors.

chrisvfritz:
Then since I'm working with the code the most, let's go on with full colons. 😅

I don't see them as similar to semicolons in JS really, as the contexts in which they're necessary are a bit more complicated and it's not obvious why they're needed, whereas in JS it's much more obvious. Also, auto-completion means I'm never actually typing a colon, while auto-completion in JS will never be able to tell you when you've actually ended a statement. Just my thoughts. 🙂

nickmessing:
@chrisvfritz, Totally agree it's not like semicolons, more like rule of "dont't type unless necessary" is similar, but I totally agree that it's your call since you're the main person to work with this codebase, still, it should be enforced and have same code style over entire codebase.

Jinjiang:
Great. I will create another PR instead of this one to fix colons and spaces in all stylus files.
Thanks.

PR #1006, opened 2017-07-11
Fix issue w/ todo list example not working

smmccabe:
The todo-list part doesn't run, it throws an error on the $emit('remove') piece, I updated it to call a method with the emit instead of directly, as i "think" that is the way, but I've been doing vue for a total of 2 hrs, which is basically everything before this in the guide.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -262,10 +262,15 @@ Vue.component('todo-item', {
   template: '\
     <li>\
       {{ title }}\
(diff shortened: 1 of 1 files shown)


javoski:
Thanks for the PR, actually adding slashes before `'` can solve this issue, like
```html
'<button v-on:click="$emit(\'remove\')">X</button>'
```

chrisvfritz:
Fixed!

PR #1007, opened 2017-07-11
Escaped single quote in string in list demo.

Jinjiang:
The single quote in the string isn't escaped which causes a JS runtime error.


chrisvfritz:
Already fixed this actually - thanks though!

PR #1016, opened 2017-07-13
Change cast for coerce word

Haeresis:
They are lot of topics about « type » in JavaScript (strong type vs. weak type, static type vs. dynamic type) etc.

I thing is a good thing if we use correct word in our JS documentation. We do not use « casting » in JavaScript but « coercion ».

So I propose to use the « coerce » word in replacement of « cast » word but I don't know if is a correct english word.

— « In computer science, type conversion, type casting, and type coercion are different ways of changing an entity of one data type into another. » JavaScript use coercion, not casting : https://en.wikipedia.org/wiki/Type_conversion

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1888,7 +1888,7 @@ All lifecycle hooks automatically have their `this` context bound to the instanc
 
 - **Modifiers:**
   - [`.lazy`](../guide/forms.html#lazy) - listen to `change` events instead of `input`
-  - [`.number`](../guide/forms.html#number) - cast input string to numbers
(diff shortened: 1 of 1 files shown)


chrisvfritz:
In this particular case, I believe cast would be correct. My understanding is that type casting is _explicit_, while coercion is _implicit_. So for example, `1 == '1'` is coercion, because it's happening automatically, but `1 === Number('1')` is casting, because we're explicitly converting `'1'` into a number. 

In this case, Vue's API uses an explicit casting with `.number`. Does that make sense?

Haeresis:
Indeed, that's make sense. Thx for your explaination, I share the same thoughs.

PR #1021, opened 2017-07-14
Update team.md

IsraelOrtuno:


src/v2/guide/team.md
```diff
@@ -704,6 +705,21 @@ order: 31
       ]
     },
     {
+      name: 'Israel Ortuño',
+      title: 'Vue Promoter and VueJobs Guy',
```
nickmessing: Maybe something like "The supreme leader of VueJobsLand"? That's kinda supposed to be funny, it'll be just an easter egg hidden by default.

src/v2/guide/team.md
```diff
@@ -704,6 +705,21 @@ order: 31
       ]
     },
     {
+      name: 'Israel Ortuño',
+      title: 'VueJobs buccaneer',
```
IsraelOrtuno: @nickmessing that may sound better ;)

PR #1026, opened 2017-07-17
Rename instance method param to instanceOptions

james-brndwgn:
It's unclear that the options parameter for the plugin has no relation to the options parameter for the instance method. Using the single name "options" is confusing.


chrisvfritz:
Agreed it could cause confusion - just updated to `methodOptions`, to be a little more clear.

PR #1027, opened 2017-07-17
Add hreflang for language URLs

lucpotage:
Hello!

It seems Google crawls the different languages as different pages. I added few lines to fix this.
By the way, do you think ISO codes should be used for language subdomains as well (ex: **ja** instead of **jp**)?

Goggle issue with different languages


chrisvfritz:
Thanks! For subdomains, we use domain suffixes instead of ISO codes, as they're often more familiar in a web context.

lucpotage:
Hello @chrisvfritz 

It seems the alternate URLs *.vuejs.org don't include the **hreflang** tags, do you know why? Without them, it can't be crawled correctly.

chrisvfritz:
@lucpotage They will - there'll just be a little lag as translators merge recent commits into their translated forks.

PR #1031, opened 2017-07-18
Update class-and-style.md

sbaidon:
I know this a really small thing but I actually find it a bit confusing.

I do realize `v-bind:class="{ active: isActive, 'text-danger': hasError }"` is valid syntax.

But if you actually try `v-bind:class="{ 'active': isActive, text-danger: hasError }"`

You will get this error: invalid expression: `v-bind:class="{ 'active': isActive, text-danger: hasError }"`

So I guess just to make a bit more consistent with the other examples and avoid that error, I would propose having both as strings :).

### DIFF
--- a/src/v2/guide/class-and-style.md
+++ b/src/v2/guide/class-and-style.md
@@ -22,7 +22,7 @@ You can have multiple classes toggled by having more fields in the object. In ad
 
 ``` html
 <div class="static"
-     v-bind:class="{ active: isActive, 'text-danger': hasError }">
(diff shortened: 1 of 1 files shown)


chrisvfritz:
It's a generally common convention in JavaScript to only surround an object key with quotes when necessary, so we're following that convention. The fact that it's only done for the key with a hyphen is a clue to less savvy JavaScripters that there's something different about it, that makes the quotes necessary.

PR #1033, opened 2017-07-19
Changed the `var` keyword to `const` in src/v2/guide/index.md

jztsaurabh:
As most of the browsers are giving support to **ES6 synatax** and **ECMA** recommends to use `const` and  `let` instead of `var`. so I have updated  `src/v2/guide/index.md` a.k.a Getting started guide to use `const` and `let` instead of `var`

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -32,7 +32,7 @@ At the core of Vue.js is a system that enables us to declaratively render data t
 </div>
 ```
 ``` js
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Vue supports IE9+ and IE9 and IE10 don't support `const`/`let`. We don't like to assume a build step in the core docs, so we stick to ES5.

jztsaurabh:
OK

PR #1043, opened 2017-07-24
Add "Bringing Designs to Life" section to state transitions guide

chrisvfritz:
@sdras What do you think?


sdras:
I'm honored! 💙 Thank you so much for including it.

PR #1047, opened 2017-07-25
Add id to todo-item component

john-cheesman:
Give each todo item an id property to use as the key in the v-for loop. As discussed in [#4825](https://github.com/vuejs/vue/issues/4825).


chrisvfritz:
The id generation needed a slight tweak to prevent duplicates after removing some todos, then adding more - but otherwise looks good. 👍

john-cheesman:
great, thanks for reviewing it

PR #1054, opened 2017-07-31
Add computed property into `v-for` example.

Haeresis:
This is a proposition, feel free to not accept that if you think it's not the correct place or if this kind of example is findable somewhere else.

I have not find this type of example and it's a basic case to avoid some code like this : 

```htm
<ul id="repeat-computed" class="demo">
  <li v-for="timestamp in timestamps">
    {{ (new Date(timestamp)).getHours() + ':' + (new Date(timestamp)).getMinutes() + ':' + (new Date(timestamp)).getSeconds()  }}
  </li>
</ul>
```

So if you can change my word for a correct english sentence, I think it's an interesting case to add into discover Guide.

What is your thought ?

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -197,7 +197,7 @@ Sometime, you deal with an array and you want use a computed property based on e
 ``` html
 <ul id="repeat-computed" class="demo">
   <li v-for="(timestamp, index) in timestamps">
-    {{ formatedHours[index] }}
+    {{ formattedHours[index] }}
   </li>
 </ul>
 ```
@@ -213,7 +213,7 @@ new Vue({
     ]
   },
   computed: {
-    formatedHours: function () {
+    formattedHours: function () {
       return this.timestamps.map(function (timestamp) {
           var now = new Date(timestamp)
           return now.getHours() + ':' + now.getMinutes() + ':' + now.getSeconds()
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Hmm, I think in this case, I'd prefer just to iterate over the computed property, like this:

``` vue
<li v-for="date in formattedDates">
  {{ date }}
</li>
```

It accomplishes the same thing, but is more straightforward. 

I'm not convinced we need an example for this, as computed properties or even the results of methods can be used with `v-for` transparently. You don't need to do anything special.

Haeresis:
> I'm not convinced we need an example for this.

No problem, now, I know the answers and the « straightforward » answer! **So just close this issue ;)**

It was because I ask myself how do that on forum after months of Vue usage. I'm not really convainced myself by the « place » of this example. But I'm sure it's a current usage. Maybe in a sort of F.A.Q.? We'll see later.

Haeresis:
I see a lot of informations explaining this usage here : https://vuejs.org/v2/guide/migration.html#Filters.

From point of view of Vue 1, it's a « migartion document »

but from point of view of beginner with Vue 2, this become an extra guide.

« Vue 2 is poor because there is no filters! I cannot use easily filters » <= No, it's easy and well explained here : https://vuejs.org/v2/guide/migration.html#Filters.

Maybe some reference to this document in the Advanced section could be resolve the point. I will think more about that.

PR #1056, opened 2017-08-01
there are two ways of mixins

dear-lizhihua:


chrisvfritz:
I'm not sure I like how the headings break up the flow here. However, I'm actually rewriting this  page into a new Composition page, so it will still be improved! I'll ping you for review when the PR is ready. 🙂

PR #1058, opened 2017-08-02
Update plugins.md

dear-lizhihua:
https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/mixins.md#custom-option-merge-strategies
Here, say 'A more advanced example can be found on Vuex's 1.x merging strategy'
And, vuex 2.x not use `optionMergeStrategies`.
So, i think 'Add some component options by global mixin. e.g. vuex' in [plugin.html](https://vuejs.org/v2/guide/plugins.html), means vuex 1.x


chrisvfritz:
Good find! I updated to Vue Router, so that we could still reference the latest version of a library, since we still add options to components there.

PR #1060, opened 2017-08-04
Prevent language name from sliding with the code

### LINKED ISSUE #936: Language name slides with code
At mobile view of vue.js documentation, when the code area is so long that it can be slided left/right, the language mark (such as HTML or JS) goes left/right with the code, instead of keeping at the top-right of the area.

fwolanski:
Resolved #936


chrisvfritz:
Looks great!

PR #1062, opened 2017-08-04
extends option text fix

fffixed:
FIX: This is similar to 'mixins', the component's own options ALSO takes higher priority than the source component being extended.

And add full mixins and extends exsamples.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -892,14 +892,22 @@ if (version === 2) {
 
   ``` js
   var mixin = {
-    created: function () { console.log(1) }
+    created: function () { console.log(1) },
+    methods: {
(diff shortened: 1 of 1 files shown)


chrisvfritz:
> This is similar to `mixins`, the difference being that the component's own options takes higher priority than the source component being extended.

This is actually accurate. With `extends`, the component you reference in the `extends` option is the extended component, while with mixins, the current component is the extended component (while the mixin is not even a complete component).

Our descriptions _are_ a little confusing though, which is why I'm actually finishing up a rewrite of the Mixins page into a more general Composition page, comparing composition strategies using `mixins`, `extends`, factory functions, and render functions. This includes a rewrite of the API entries too.

When that's done, would you mind if I pinged you for a review?

PR #1064, opened 2017-08-05
remove false statement

rayrutjes:
We here wrongly informed users that we will know more about use cases for the `key` prop, but it is no longer existing.


chrisvfritz:
We talk about it [here](https://vuejs.org/v2/guide/transitions.html#Transitioning-Between-Elements):

> When toggling between elements that have the same tag name, you must tell Vue that they are distinct elements by giving them unique key attributes. Otherwise, Vue’s compiler will only replace the content of the element for efficiency. Even when technically unnecessary though, it’s considered good practice to always key multiple items within a <transition> component.

rayrutjes:
I didn't realize that "guide" !== "page"

PR #1065, opened 2017-08-05
[API] add valid "number" type for "key"

rayrutjes:
Not sure if I read this correctly, but it looks like key can work fine with numbers.

https://github.com/vuejs/vue/blob/5981529b497f5ef064b123aeb58eb9655172d6eb/src/core/instance/render-helpers/render-list.js#L12

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1975,7 +1975,7 @@ if (version === 2) {
 
 ### key
 
-- **Expects:** `string`
+- **Expects:** `number | string`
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
That's actually referring to a different `key` - the key of a property of an object used with `v-for`, but you're incidentally correct that this key is [also typed to accept numbers](https://github.com/vuejs/vue/blob/dev/types/vnode.d.ts#L18). 😄

rayrutjes:
🎉  That is fun 🎉

PR #1067, opened 2017-08-08
Add a tip about using ESNext in template expressions

Justineo:
... to clarify situations discussed in vuejs/vue#6307.

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -80,6 +80,8 @@ These expressions will be evaluated as JavaScript in the data scope of the owner
 
 <p class="tip">Template expressions are sandboxed and only have access to a whitelist of globals such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
 
+<p class="tip">ESNext features are available in template expressions with the caveat that the feature set is limited by the **Node.js** version you use.</p>
+
 ## Directives
 
 Directives are special attributes with the `v-` prefix. Directive attribute values are expected to be **a single JavaScript expression** (with the exception for `v-for`, which will be discussed later). A directive's job is to reactively apply side effects to the DOM when the value of its expression changes. Let's review the example we saw in the introduction:


src/v2/guide/syntax.md
```diff
@@ -80,6 +80,8 @@ These expressions will be evaluated as JavaScript in the data scope of the owner
 
 <p class="tip">Template expressions are sandboxed and only have access to a whitelist of globals such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
 
+<p class="tip">ESNext features are available in template expressions with the caveat that the feature set is limited by the **Node.js** version you use.</p>
```
rpkilby: A minor nitpick, but this supposes that rendering occurs on the server, not on the client.

chrisvfritz:
This isn't really the right place for a note like this, since it's specific to using Babel and `.vue` files. The [vue-template-compiler](https://www.npmjs.com/package/vue-template-compiler) docs would probably be more appropriate, but I think still inadequate, since it's extremely unlikely that anyone will find that note when they encounter an error.

I'm continuing the conversation in the original issue to try to find a better solution. This is definitely unexpected behavior for a user, so I think either a warning when an error occurs, or finding some way to work around the problem is needed.

Justineo:
Thanks Chris. Obviously it's the best if the original issue can be solved.

src/v2/guide/syntax.md
nickmessing: @rpkilby, template validation occurs on the server.

nickmessing:
@Justineo, opened a PR for that (thank you @chrisvfritz for the suggestion). vuejs/vue#6328

Justineo:
Should we still address the difference between precompiled template and runtime template? Like runtime template won't be transformed by Babel or other tools.

chrisvfritz:
@Justineo Good question! My current thinking is that if someone is using SFCs with the runtime-only build, it makes sense for them to expect that the JavaScript they use in a `script` will always work in a `template` as well - so that's the main problem I'm hoping to solve. 

Since Babel is a tool only used in the development environment, rather than bundled with client code, my _hope_ is that it will be intuitive to users that templates parsed on the client won't have access to Babel. Do you think that makes sense?

Justineo:
@chrisvfritz Yes that makes sense.

PR #1070, opened 2017-08-08
Replace toBe with to.equal for typeof tests

bilougit:
Having this undefined is not a function (evaluating 'expect(_MyComponent.default.data).toBe('function')')
when using toBe. No more error when using to.equal


chrisvfritz:
Note this comment:

> // Here are some Jasmine 2.0 tests, though you can
> // use any test runner / assertion library combo you prefer

Since you're not using Jasmine, but another assertion library with a different syntax, you have to adjust the examples to use the assertion syntax supported by that library.

PR #1071, opened 2017-08-08
Adjust example for circular component

LinusBorg:
Since vue-loader 13.*, we have to use `require('./component.vue').default` instead of `require('./component.vue')`

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -1288,7 +1288,7 @@ In our case, I'll make that point the `tree-folder` component. We know the child
 
 ``` js
 beforeCreate: function () {
-  this.$options.components.TreeFolderContents = require('./tree-folder-contents.vue')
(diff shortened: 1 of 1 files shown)


Jinjiang:
Sorry for a small question: Is it better to use `import` instead? And I've found there are other 3 code snippets using `import` in the same file like https://github.com/vuejs/vuejs.org/pull/1071/files#diff-044f0befe8adfe59792a31b1763e3cc4R1139.
Thanks.

LinusBorg:
Using `import()` there would "work", but not be the same - `import()` creates an async module, so wouldn't be included in the main bundle.

Jinjiang:
Got this and synced the update into cn docs. :-)
Thanks.

jbayona89:
https://vuejs.org/v2/guide/components.html#Circular-References-Between-Components seems to be outdated still

chrisvfritz:
@jbayona89 Just fixed it. 🙂

PR #1074, opened 2017-08-10
[Doc EN]: API more consistency

Haeresis:
We have correct lot of consistency detail while we have translate this document in french.

Signed-off-by: Bruno Lesieur <bruno.lesieur@gmail.com>


chrisvfritz:
_Magnifique!_ 🎉 😄

PR #1075, opened 2017-08-10
Keep-alive 2.2 hooks included...

rayfoss:
I was googling for this and tried every hook listed here, they don't run when I need them to as I use keep-alive.


chrisvfritz:
As examples, this section is not trying to provide a complete list of hooks (and even with what you've added, it is still not complete). You can see the complete list in the sidebar of [Lifecycle Hooks](https://vuejs.org/v2/api/#Options-Lifecycle-Hooks).

PR #1081, opened 2017-08-15
Missing letter

whoan:


chrisvfritz:
It was actually correct with `separate`, as `separated` would imply that they were combined at some point. For clarification, I've updated your PR to change to the word to "different", which has the same meaning here, but is a more common word in English. 🙂

PR #1082, opened 2017-08-16
Fix Lodash not loading from unpkg

ridewn:
For some reason, unpkg is not loading lodash.min.js, which is breaking the question example for watchers on the computed properties page. I've checked the .min links for axios and vue on unpkg, and it looks like it is just an issue with lodash. For example https://unpkg.com/lodash@4.13.1/lodash.min.js will return an error saying

> Cannot find file "/lodash.min.js" in package lodash@4.13.1

Going to https://unpkg.com/lodash@4.17.4/lodash.min.js results in the same error but with the 4.17.4 tag.

I have checked the directories on unpkg and they are listed as being there. Someone may want to get in contact with unpkg to let them know this, as for now I've suggested to use cdnjs for the lodash@4.13.1 min.js file which fixes the error the computed properties page was getting (_ is not defined).

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -203,7 +203,7 @@ For example:
 <!-- is able to remain small by not reinventing them. This also   -->
 <!-- gives you the freedom to just use what you're familiar with. -->
 <script src="https://unpkg.com/axios@0.12.0/dist/axios.min.js"></script>
-<script src="https://unpkg.com/lodash@4.13.1/lodash.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.13.1/lodash.min.js"></script>
 <script>
 var watchExampleVM = new Vue({
   el: '#watch-example',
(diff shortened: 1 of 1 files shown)


jimaek:
How about using jsDelivr instead? It can also pull from npm and can offer better reliability because it uses multiple CDNs.

https://cdn.jsdelivr.net/npm/lodash@4.17.4/lodash.min.js
https://cdn.jsdelivr.net/npm/lodash@4.17/lodash.min.js

chrisvfritz:
We actually used to use jsDelivr for everything, but moved to unpkg to be able to use specific builds. Now that jsDeliver can do that too, I've moved all non-Vue CDN links back there and may soon update the Vue links as well.

ridewn:
Thanks Chris, the translated sites also use unpkg, so you may want to change those over to jsDelivr!

jimaek:
Awesome. Let me know if you need anything, I will be glad to help out.

PR #1084, opened 2017-08-16
Update index.md

gurix:
lodash didn't load properly in your example I updated it in a forked fiddle


chrisvfritz:
Thanks! It looks like unpkg.com has fixed this issue now, so the original example should work again. In the future, we'll probably move our CDN links back to jsDelivr now that they support NPM builds.

PR #1086, opened 2017-08-18
Reorganize sidebar and some content; cannibalize reactivity page

chrisvfritz:
[__Browse the PR live here.__](https://vue-advanced-reorg.surge.sh/v2/guide)

### What I did

- Split _Advanced_ sidebar category into smaller, more meaningful categories.
- Renamed some page titles.
- Removed _Reactivity_ page, as I think a lot of its content may actually be too advanced for end users, but I've kept it hidden for future adaptation into a detailed Contributor's Guide to Vue's internals. Though removed, the essential information has been spread out throughout the Guide and API.
  - API entries for `mounted` and `updated` now cover `vm.$nextTick`.
  - Covered change detection caveats for objects (including `Vue.set`/`vm.$set`) under _List Rendering_, where the caveats for arrays were already listed.
  - Reorganized _List Rendering_, most notably putting the component section at the end. I discovered that many people were skipping the rest of the page once they got to components, since they hadn't learned components yet. Then they never came back. The result is that many people never even read about array change detection and some of the more niche `v-for` features. 😬 
  - Rewrote _The Vue Instance_, as many beginner and intermediate users were glazing over at terms like constructor, proxy, etc, leading to many assuming that the reactivity system is more complicated than it is (and not absorbing essential tips). I also tried to make some essential information more clear, e.g. data that starts out empty or non-existant still has to be registered with a starting value.
  - Added note about the devtools on the _Installation_ page
- I noticed we were using a combination of `->` and `=>` to represent expression return values, so I arbitrarily standardized on `=>`.
- Moved _Filters_ to its own page, removing it from _Syntax_, because it's not really an essential feature. 

### Questions

- Are the new sidebar categories meaningful to you? When you see what's inside them, are they the kinds of pages you'd expect to see? Is there anything that feels out of place?
- Do the new page titles make sense to you?
- How is [this intro paragraph](https://vue-advanced-reorg.surge.sh/v2/guide/installation.html#Vue-Devtools) of the devtools? Reading this as a user who has just decided to install Vue, do you feel it will entice people enough to click through to learn more?
- In the rewrite of [_The Vue Instance_](https://vue-advanced-reorg.surge.sh/v2/guide/instance.html), I've vastly simplified the explanation of the reactivity system. In that simplification, are there any details I've left out that you feel are essential for users to build successful applications, avoiding common mistakes?
- With the reorganization of [_List Rendering_](https://vue-advanced-reorg.surge.sh/v2/guide/list.html), do you feel the page still flows well? Do you feel the sections are generally ordered by most to least essential?


src/v2/guide/instance.md
```diff
@@ -16,46 +16,68 @@ var vm = new Vue({
 
 Although not strictly associated with the [MVVM pattern](https://en.wikipedia.org/wiki/Model_View_ViewModel), Vue's design was partly inspired by it. As a convention, we often use the variable `vm` (short for ViewModel) to refer to our Vue instance.
 
-When you instantiate a Vue instance, you need to pass in an **options object** which can contain options for data, template, element to mount on, methods, lifecycle callbacks, and more. The full list of options can be found in the [API reference](../api/#Options-Data).
+When you create a Vue instance, you pass in an **options object**. The majority of this guide describes how you can use these options to create your desired behavior. For reference, you can also browse the full list of options can in the [API reference](../api/#Options-Data).
```
phanan: `full list of options in`, not `can in`.

src/v2/guide/instance.md
```diff
@@ -16,46 +16,68 @@ var vm = new Vue({
...
+   |- TodoItem
+      |- DeleteTodoButton
+      |- EditTodoButton
+   |- TodoListFooter
+      |- ClearTodosButton
+      |- TodoListStatistics
 ```
 
-Although it is possible to create extended instances imperatively, most of the time it is recommended to compose them declaratively in templates as custom elements. We will talk about [the component system](components.html) in detail later. For now, you just need to know that all Vue components are essentially extended Vue instances.
+We'll talk about [the component system](components.html) in detail later. For now, just know know that all Vue components are also Vue instances, and so accept the same options object (except for a few root-specific options).
```
phanan: Duplicate `know`.

src/v2/guide/instance.md
```diff
@@ -16,46 +16,68 @@ var vm = new Vue({
...
-data.a // -> 2
+data.a // => 2
 
 // ... and vice-versa
 data.a = 3
-vm.a // -> 3
+vm.a // => 3
+```
+
+When this data changes, the view will re-render. It should be noted that properties in `data` are only **reactive** if they existed when the instance was created. That means if you add a new property, like:
```
phanan: Isn't "data" plural?

src/v2/guide/instance.md
```diff
@@ -16,46 +16,68 @@ var vm = new Vue({
...
-It should be noted that only these proxied properties are **reactive**, meaning any changes to their values will trigger the view to re-render. If you attach a new property to the instance after it has been created, it will not trigger any view updates. We will discuss the reactivity system in detail later.
+Then changes to `b` will not trigger any view updates. If you know you'll need a property later, but it starts out empty or non-existent, you'll just need to set some initial value. For example:
+
+``` js
+data: {
+  newTodoText: '',
+  visitCount: 0,
+  hideCompletedTodos: false,
+  todos: [],
+  errorMessage: null
```
phanan: I'd suggest `error` instead of `errorMessage`, as a "message" would imply a string type.

src/v2/guide/instance.md
```diff
@@ -64,25 +86,25 @@ var vm = new Vue({
...
-
-Consult the [API reference](../api/#Instance-Properties) for the full list of instance properties and methods.
+In the future, you can consult the [API reference](../api/#Instance-Properties) for a full list of instance properties and methods.
 
 ## Instance Lifecycle Hooks
 
-Each Vue instance goes through a series of initialization steps when it is created - for example, it needs to set up data observation, compile the template, mount the instance to the DOM, and update the DOM when data changes. Along the way, it will also invoke some **lifecycle hooks**, which give us the opportunity to execute custom logic. For example, the [`created`](../api/#created) hook is called after the instance is created:
+Each Vue instance goes through a series of initialization steps when it's created - for example, it needs to set up data observation, compile the template, mount the instance to the DOM, and update the DOM when data changes. Along the way, it also runs functions called **lifecycle hooks**, giving users the opportunity to add their own code at specific stages.
+
+For example, the [`created`](../api/#created) hook can be used to run code after an instance is created:
```
phanan: You mean `can be use to run after`?

src/v2/guide/reactivity.md.hidden
```diff
@@ -1,16 +1,67 @@
...
 type: guide
-order: 12
+order: -1
 ---
 
-We've covered most of the basics - now it's time to take a deep dive! One of Vue's most distinct features is the unobtrusive reactivity system. Models are just plain JavaScript objects. When you modify them, the view updates. It makes state management very simple and intuitive, but it's also important to understand how it works to avoid some common gotchas. In this section, we are going to dig into some of the lower-level details of Vue's reactivity system.
+!!!!!!!!!!!!!!!!!!!!!!!!!!
+!! NOTE FOR TRANSLATORS !!
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+!! Don't both translating changes to this page yet, !!
```
phanan: Don't _bother_?

src/v2/guide/instance.md
chrisvfritz: Technically, yes, but in practical use I think most people treat it as singular, e.g. "My data is backed up."

src/v2/guide/instance.md
chrisvfritz: Will do, though in this case, I'd still actually still have `errorMessage` be `null` when it'll be a string later. The reason is a blank error message can still mean something different than a non-existent one.

src/v2/guide/instance.md
chrisvfritz: In this case, `can be used to run code after` does actually say what I mean. What meaning does it have for you?

chrisvfritz:
@phanan Thanks for your excellent review. 😻  Great attention to detail!

src/v2/guide/instance.md
phanan: Oh, I misread it then. Yeah, it makes perfect sense to me now.

src/v2/guide/filters.md
```diff
@@ -0,0 +1,48 @@
...
+
+``` html
+<!-- in mustaches -->
+{{ message | capitalize }}
+
+<!-- in v-bind -->
+<div v-bind:id="rawId | formatId"></div>
+```
+
+<p class="tip">Vue 2.x filters can only be used inside mustache interpolations and `v-bind` expressions (the latter supported in 2.1.0+), because filters are primarily designed for text transformation purposes. For more complex data transforms in other directives, you should use [Computed properties](computed.html) instead.</p>
```
hootlex: This paragraph repeats the way of using filters.
The part "(the latter supported in 2.1.0+)" can be moved to the intro paragraph, so this can be: "Filters are primarily designed for text transformation purposes. For more complex data transforms....."

src/v2/guide/instance.md
```diff
@@ -16,46 +16,68 @@ var vm = new Vue({
...
-var myComponentInstance = new MyComponent()
+```
+Root Instance
+|- TodoList
+   |- TodoItem
+      |- DeleteTodoButton
+      |- EditTodoButton
+   |- TodoListFooter
+      |- ClearTodosButton
+      |- TodoListStatistics
```
hootlex: I am not sure if it's a good idea to show an example component tree at this point. My concern is that even though it's just an example, people with little experience may get confused or try to follow it as a rule.

If you are going to keep it, maybe it's safer to make it more general. For example:
```
Root Instance
    |- TodoList   
        |- TodoItem
            |- CreateTodoItem
            |- ShowTodoItem
    ...
```

chrisvfritz:
Thanks for the review @hootlex! 😄 

> Maybe the Transitions & Animation can go above Reusability & Composition

I might be misunderstanding, but isn't it already above _Reusability & Composition_? If you're suggesting the opposite of what I'm thinking though, then I put Transition & Animation first because it allows you to do something new, whereas Reusability & Composition just allow you to build the same kinds of apps and components, just organized or optimized differently. Does that make sense?

> why Single File Components page is under tooling. Wouldn't it fit better in the Components category?

Because they're not actually a feature of Vue _core_, but of vue-loader/vueify/etc. They're not possible without build tools.

> a small screenshot of DevTools in action would also look nice

I actually experimented with adding a screenshot there, but I worried that it took up too much vertical space. Once people have the devtools, I'd like it to be easy to skip over this section to the actual installing of Vue. Once people click through, they'll see a nice screenshot. 🙂 

> the key section to note that it's required when iterating components? 

We do currently have it stand out in a version note and users also receive a warning in the console when using the dev build, so I feel like we might have this well-covered already.

> I'd also like to have a warning in the array caveats section, just to draw attention.

This is something a lot of people have missed in the past, because it was below the section about components that we encouraged people to skip. 😅  My hope now is that with the reordering of this page, many more people will see it. 

It is important though, so if we still notice a problem with new users, this is a measure I'll consider. I'm just not sure which part we'd turn into a warning, as _the entire section_ is really a sort of warning - and hopefully, the fact that "Caveats" exists in the title will draw attention to it.

src/v2/guide/filters.md
chrisvfritz: Nice catch! I just removed the redundancy. 🙂

src/v2/guide/instance.md
chrisvfritz: Thanks for bringing this to my attention! I'll do a little more user testing on this part.

src/v2/guide/instance.md
chrisvfritz: @phanan I'm still open to changing it though. What do you think?

hootlex:
@chrisvfritz it seems that you have thought everything 😄

Regarding the "Animations & Transitions", I wanted to say the opposite. Hadn't thought of it this way. It makes sense to show more features before diving into specific needs. 👍🏻

About "Array Caveats", hopefully, people will notice it with the new order. Otherwise, it's hard to decide in which part to emphasize. It's like the whole section should be a warning.

chrisvfritz:
We got a couple reviews, so I'm going ahead and merging this. I at least feel confident there's nothing _disasterous_ here. 😄

cwill:
Just weighing in, providing info about reactivity throughout the guide as it's relevant seems pretty helpful, but I've found the 'Reactivity In Depth' page an extremely useful reference, especially when working with vuex mutations & dealing with some harder-to-diagnose bugs resulting from setting fields incorrectly. It'd be nice if a consolidated summary of this behavior could stick around somewhere - I needed to reference it today and had to pull it up from archive.

chrisvfritz:
@cwill The page is also still [here in source](https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/reactivity.md.hidden), partially rewritten. The plan is to expand it into a Contributor's Guide that's even more detailed than the current document.

As for the use cases you're thinking of, if the information isn't easily searchable either here or in the Vuex docs, opening an issue would be very welcome. In the API doc in particular, we've been adding more of these kinds of warnings.

chrisvfritz:
@cwill Thanks! 😄

yyx990803:
Maybe a bit late, but I still think the Reactivity Page should be preserved somewhere, especially for those who wish to have a go-to place when they need to point co-workers to a "source" where explains the whole thing in one place.

I'm thinking maybe we should one extra section titled "Advanced / Internals" that hosts the reactivity chapter and the render function chapter, as both of these somewhat relies on knowledge of how Vue actually works under the hood. IMO these are not only valuable for contributors but also for normal users.

rpkilby:
As a normal user, the reactivity docs have been invaluable in the past. Definitely agree that they should live in a separate section.

chrisvfritz:
@yyx990803 Sounds good! I'll have it back up in a new section soon.

PR #1092, opened 2017-08-25
Update class-and-style.md

dear-lizhihua:
add end-tag `</div>`

### DIFF
--- a/src/v2/guide/class-and-style.md
+++ b/src/v2/guide/class-and-style.md
@@ -82,7 +82,7 @@ computed: {
 We can pass an array to `v-bind:class` to apply a list of classes:
 
 ``` html
-<div v-bind:class="[activeClass, errorClass]">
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Hmm, I actually left off a closing tag intentionally. My reasoning was this was only the opening tag was necessary for a demo. With the closing tag, it implies a complete code example, which seems strange to me, as these classes don't really make sense on an empty div.

Does that make sense? Obviously it was confusing to some people, so I'm open to discussion. Is there another alternative that would be less confusing for everyone?

kazupon:
I could understand it why there was no closing tag.
However, considering consistency with other examples, I think it should be closed with a closing tag.

If we don't close with a closing tag, I think there are the following options:

```html
<div> ... </div>
```

```html
<div> something contents ...</div>
```

PR #1097, opened 2017-08-29
Including info about NativeScript in Native Rendering

rdlauer:
Just added a paragraph about NativeScript and its Vue integration in the
"Native Rendering" section of the guide


chrisvfritz:
Thanks! I shortened it slightly and made it more clear that the plugin is still relatively early in development.

PR #1101, opened 2017-09-01
Move reactivity & render function docs into "Advanced / Internals" section

rpkilby:
Per this [suggestion](https://github.com/vuejs/vuejs.org/pull/1086#issuecomment-326604000), added an "Advanced / Internals" section and moved the reactivity and render function docs to it.

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -1,7 +1,7 @@
 ---
 title: Comparison with Other Frameworks
 type: guide
-order: 701
+order: 801
 ---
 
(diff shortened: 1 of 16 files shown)


chrisvfritz:
I'm partway through some updates to the Reactivity page that I want to finish before republishing, but I'll continue that work in this PR. Thanks for getting it started!

chrisvfritz:
There were enough changes I wanted to make that it was easier to apply them outside of this PR, rather than building on top of it, but the Reactivity page is live again! 🙂  I've delayed rewriting that page for now, as there are other tasks I'd like to get done first.

Alex-Sokolov:
@chrisvfritz is it need to restore links to Reactivity page (for example deleted from API page)?

rpkilby:
> rather than building on top of it

Of course - I assumed as much. Thanks @chrisvfritz!

chrisvfritz:
@Alex-Sokolov Good catch! Just reverted.

PR #1102, opened 2017-09-02
Update index.md

Runrioter:
tweaks data source to make it work well

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -53,7 +53,7 @@ var app = new Vue({
 </script>
 {% endraw %}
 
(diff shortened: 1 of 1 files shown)


nickmessing:
@Runrioter, `app.message` and `app2.message` work well while `app.__vue__` is undefined, where did you get that from?

nickmessing:
Screenshot

Runrioter:
@nickmessing   I set up a project with vue-cli. I did as what documents said.
image

Runrioter:
@nickmessing  I got it,  if use UMD, it works well as documents said. If use with a bundler like webpack, it does't.

PR #1110, opened 2017-09-05
Typo: "let's cheat" should be "let's chat"

msaun008:
I believe the intention was for the phrase to be `let's chat` instead of `let's cheat`.  Full sentence comparison below.

```
BEFORE:
It’s good to understand why the rules exist though, so let’s cheat.

AFTER:
It’s good to understand why the rules exist though, so let’s chat.
```

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -132,7 +132,7 @@ Vue.component('my-component', {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
"let's cheat" is actually correct in this context. We're breaking the rule that Vue has set to understand why it exists.

msaun008:
Thanks for clarifying!

PR #1111, opened 2017-09-06
Update Component-Scoped CSS section of React comp

rawrmonstar:
Hi! I rewrote the Component-Scoped CSS section of the React comparison to hopefully provide a more neutral opinion about CSS-in-JS styling solutions--both React and Vue support them after all!

I'm definitely open to feedback about rewording anything y'all have suggestions about 😄

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -68,9 +68,9 @@ On a higher level, we can divide components into two categories: presentational 
 
 #### Component-Scoped CSS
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I like this a lot! There are a couple very minor tweaks I'd like to make, but I should probably wait until tomorrow when it's not 2:30am. 😄

chrisvfritz:
@rawrmonstar Thanks again! 🎉

PR #1112, opened 2017-09-06
fix blog page layout

sabrinaluo:
https://vuejs.org/2015/10/28/why-no-template-url/

## Before
Desktop
image

Mobile
image

## After
Desktop
image

Mobile
image


chrisvfritz:
Thanks! 😄

PR #1116, opened 2017-09-08
Add v-bind example to components

sebastiandedeyne:
There's no example of using `v-bind` for multiple props in the guide, only in the API references. This PR adds a quick example in the Components section.


chrisvfritz:
Thanks! I tweaked it slightly, but I think this is a great place for a comment like this. 🙂 👍

PR #1118, opened 2017-09-09
Explain components dependencies

zcyzcy88:


chrisvfritz:
Thanks for the suggestion, but the purpose of this section isn't really to help people understand more about different kinds of components, but simply highlight some scenarios where optimization is different than it is in React.

zcyzcy88:
Not, I add this **premise** to help people better understand: what decide render output, and why React need re-render lots and why Vue not.

PR #1122, opened 2017-09-11
Update list.md

alexsales:
Add clarifying statement on the significance for using the special is="" attribute. This may be useful for newbies new to web development and/or front-end frameworks. The is="" attribute isn't a well-known html attribute outside of the web components community.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -488,6 +488,8 @@ Here's a complete example of a simple todo list:
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


alexsales:
2nd commit: Add clarifying statement on use of $emit method since first time being used up until this point in the docs.

chrisvfritz:
Thanks! I made a couple changes:

- Tweaked the wording slightly.
- Removed the note about `$emit`, as it's not a feature with specific relevance within the context of list rendering.

PR #1124, opened 2017-09-12
[Doc EN]: Vue-Router to Vue Router

Haeresis:
If you see the https://vuejs.org/v2/guide/migration-vue-router.html document and you check into Ecosystem Tab in the main menu of the website, you will see the name for Vue Router is « Vue Router ».

So vue-router (without capital) refer to the package or some URL but the product is Vue Router. So some mixed term like Vue-Router should be transformed into vue-router or Vue Router depending of the context.

What are your throughs about that @chrisvfritz?

More explanation in comment


chrisvfritz:
Great catches!

PR #1125, opened 2017-09-12
[Website]: New item `Vue Server Renderer` in main menu

Haeresis:
Because we have a great full documentation, and because we have now a Scalling Up section that speaks about :

— Routing (Vue Router),
— State Management (Vuex), and
— Server-Side Rendering (Vue Server Renderer)

I think it should be a good thing to propose it directly from main Ecosystem menu.

What are your throughs about this @chrisvfritz?


Haeresis:
What about a vue-loader article in Scaling Up section ?

chrisvfritz:
@Haeresis I like the addition! As for vue-loader, we already cover it under Tooling and I was thinking of keeping Scaling Up to dependencies, while Tooling would be devDependencies. What do you think?

Haeresis:
Oh that's make sence! I will have a better understanding about vue-loader because we will start the french translation soon. Thx !

Issue #1120, opened 2017-09-11
Revise language that assumes ease ('simply', 'obviously', etc.)


robwierzbowski:
Vue **is** more easy, simple, and fun to use than other reactive frameworks 🎆.

But pointing out how 'simple', 'obvious', or 'easy' something is can have the opposite-of-intended effect. If a person doesn't understand the 'simple' code, the project can seem inaccessible. "If I don't understand this, what chance do I have to understand the rest?" There are plenty of articles on this effect, and I've observed it personally when working with junior co-workers.

To make the Vue documentation more friendly, I'd like to remove/revise some words that assume ease from the docs. Here's a list of potentially problematic words: https://css-tricks.com/words-avoid-educational-writing/. I can find more reference articles if you'd like. 

In most cases removing the word in question makes a sentence flow better; in some cases replacing the word with a more explicit one is a good option ('simple module' with 'small module', for example).

Would you be interested in a PR that revises one or two pages? If it looks good I could continue with the rest of the site.

chrisvfritz:
@robwierzbowski That sounds great. PR more than welcome! 👍

robwierzbowski:
All right! I'll give it a shot.

wujekbogdan:
@robwierzbowski 
Obviously, you're right. Let's just remove these additional words as they don't add any value. However, everyone knows that a PR would be welcome. So, basically, make a PR and it will be merged. Easy!

PR #1126, opened 2017-09-12
Reduce instances of words that shouldn't be used in educational writing.

robwierzbowski:
Fixes https://github.com/vuejs/vuejs.org/issues/1120.

I tried not to revise too much, mostly removing instances of just, simply, obviously, easily. In most cases this made the language more direct and concrete, which I think aides understanding. 

There are places where these words are useful, so this isn't a 100% find/replace/delete. I used my judgement — comments are welcome. 

Around half of the removals relate to my original issue (don't assume ease), and the other half improve the flow of reading by removing words when they're unnecessary for understanding. 

I avoided revising any blog posts, except common-gotchas which I thought might still get a lot of views.


src/v2/guide/join.md
```diff
@@ -1,51 +1,51 @@
...
-- **Watch a repo you care about.** This will send you notifications whenever there's activity in that repository, giving you insider knowledge about ongoing discussions and upcoming features. It's a fantastic way to build expertise so that you're eventually able to help address issues and pull requests.
-
-### Translate Docs
-
-Vue has already spread across the globe, with even the core team in at least half a dozen timezones. [The forum](https://forum.vuejs.org/) includes 7 languages and counting and many of our docs have [actively-maintained translations](https://github.com/vuejs?utf8=%E2%9C%93&query=vuejs.org). We're very proud of Vue's international reach, but we can do even better.
-
-I hope that right now, you're reading this sentence in your preferred language. If not, would you like to help us get there?
-
-If so, please feel free to fork the repo for [these docs](https://github.com/vuejs/vuejs.org/) or for any other officially maintained documentation, then start translating. Once you've made some progress, open an issue or pull request in the main repo and we'll put out a call for more contributors to help you out.
```
robwierzbowski: Not sure why this is a whole file diff. I removed a single `just`.

src/v2/cookbook/adding-instance-properties.md
```diff
@@ -46,7 +46,7 @@ Then what would you expect to be logged below?
 new Vue({
   data: {
     // Uh oh - appName is *also* the name of the
-    // instance property we just defined!
```
rpkilby: "just" implies proximity, rather than conceptual simplicity

src/v2/api/index.md
```diff
@@ -1357,7 +1357,7 @@ type: api
 
 - **Usage:**
 
-  Watch an expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The expression only accepts simple dot-delimited paths. For more complex expression, use a function instead.
```
rpkilby: "simple" emphasizes the lack of complexity of the expression, rather than conceptual simplicity

src/v2/cookbook/adding-instance-properties.md
robwierzbowski: It does. I think the "recently" is implied, but happy to revert. 

I definitely made some judgement calls, so comments welcome.

src/v2/guide/join.md
rpkilby: Maybe windows vs unix line endings?

src/v2/api/index.md
robwierzbowski: IMO it's superfluous, and reads more simply (ha!) without it.

robwierzbowski:
Thanks for starting a review @rpkilby. Happy to revert anything, or continue revisions. Just to be clear, I did remove more than just conceptual simplicity instances of these words, noted in the PR description.

rpkilby:
Yep, just demonstrating a couple of the places where the removed words were not referencing conceptual accessibility/simplicity.

Either way, +1 on the changes. It's kind of interesting - never really thought about this before.

chrisvfritz:
Thank you @robwierzbowski! It might take me a few days to review with my schedule this week, but I'm really looking forward to it. 😄

znck:
👍 Nice work @robwierzbowski.

codebryo:
Wow, cool PR. Thanks for that. <3

chrisvfritz:
@robwierzbowski Sorry it took me a bit longer to get through this! I didn't see anything that wasn't at least neutral to me and most were definite improvements. 😄  Thanks again and I look forward to more PRs in the future!

robwierzbowski:
Thanks! I might submit a PR with some potential simplifications I noted while writing this one.

I'm trying to get some projects I'm involved with on Vue, and I think the documentation is a huge plus. Happy to give back.

rpkilby:
Hi @robwierzbowski - out of curiosity, did you use any linting tools for this PR, or did you do this by hand?

robwierzbowski:
@rpkilby Just find and replace 🙂

robwierzbowski:
I️ also read over pretty much every line in the documentation. So, find and replace +

rpkilby:
Thanks. I also just noticed the article you originally referenced.

PR #1128, opened 2017-09-13
Update index.md

nishadunnisa:
Grocery list was confusing. It makes more sense. I and my team members who are newbies got confused with `item`


chrisvfritz:
Thanks for pointing this out! I'm not sure the examples still make sense with a todo list, so to avoid having to change too much, do you think `shoppingList` would be easier to understand? To replace `item`, what do you call the things in your shopping list? Is there a more common word we could use in this case?

chrisvfritz:
@nishadunnisa Closing due to inactivity, but I'm happy to reopen again if you'd like. 🙂

PR #1134, opened 2017-09-15
Change recommended links to jsDelivr

LukasDrgon:
[jsDelivr](https://www.jsdelivr.com/) is the [fastest opensource cdn](https://www.cdnperf.com/) available and can now serve any project from npm with zero config. This PR changes the links from unpkg [back to jsDelivr](https://github.com/vuejs/vuejs.org/pull/1082#issuecomment-322774489). vuejs/vue#6618 should be merged first to make sure that the short links work correctly.

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -111,7 +111,7 @@ React is renowned for its steep learning curve. Before you can really get starte
 While Vue scales up just as well as, if not better than React, it also scales down just as well as jQuery. That's right - all you have to do is drop a single script tag into a page:
 
 ``` html
-<script src="https://unpkg.com/vue"></script>
+<script src="https://cdn.jsdelivr.net/npm/vue"></script>
 ```
 
 Then you can start writing Vue code and even ship the minified version to production without feeling guilty or having to worry about performance problems.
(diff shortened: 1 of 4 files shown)


chrisvfritz:
@LukasDrgon I'm OK with this. We've had nothing but good experiences with jsDelivr and some community members have mentioned that it's even faster in their part of the world. Some requests however:

- I notice we're using both `.net` and `.com` throughout our links. Let's standardize on `.com`.
- Let's always use the shortest version of the link possible. For example, `https://www.jsdelivr.com/npm/vue/` instead of `https://www.jsdelivr.com/package/npm/vue/`.
- Let's still mention unkpg.com as one of the options, similar to cdnjs.
- Let's maintain the existing formatting. For example, there's a missing space after `Recommended:`.

LukasDrgon:
@chrisvfritz I fixed the formatting and added unpkg as an alternative.

Regarding `.com` vs `.net` - `.net` is the CDN and `.com` is the website (that's also the reason why there's `/package/` in one link). I could replace the link to https://www.jsdelivr.com/package/npm/vue with https://cdn.jsdelivr.net/npm/vue/ if you want, but it's a different page with different content, so let me know which one you prefer.

chrisvfritz:
Ah, thanks for the explanation! I feel like it's easier to remember to just add an ending slash to `https://cdn.jsdelivr.net/npm/vue`, so let's go with that one. After that, this is looking good to merge as soon as vuejs/vue#6618 is. 👍

PR #1135, opened 2017-09-16
[Doc EN]: kb to kB

Haeresis:
We just see the « ko » for « kilo-octet » become « kB » for « kilo-byte » in english.

 « kb » means « kilobit » that it's not correct.
— Source : https://en.wiktionary.org/wiki/kB
— Source : https://fr.wiktionary.org/wiki/kB

That's cause an issue from french documentation where someone said us « hey, you'll should not translate “kb” by “ko” because it's height time more bigger ! »


chrisvfritz:
I believe we're actually measuring sizes in `KB` (1024 bytes) as opposed to `kB` (1000 bytes), but I agree `kb` (all lowercase) is technically wrong, as I think it implies _bits_. 🙂  What would you think about standardizing on `KB`?

Haeresis:
Thx! It's a new thing I know!

It's just because for french people use « ko (kB) » (And I just see there is a « Kio (KB) » equivalent I never see before).

It's you know the best usage!

_20 480KB or 20 000kB will always done “20kb min+gzip Runtime” so I don't care ;)_

KB or kB no matter for me :)

Haeresis:
Oh my bad, I forgot to change kB in KB (to be more precise, I thinking you'll doing that).

PR #1136, opened 2017-09-16
Minor edits to improve readability

rlneumiller:


chrisvfritz:
Looks great! Thanks. 🙂  The only part I disagreed with was the use of `however`, so just reverted that.

PR #1137, opened 2017-09-16
Add description of how to use a .vue component

arve0:
Hi!
It was not immediate clear to me how one should register a component (I'm totally new to Vue). This suggestion adds a description.

### DIFF
--- a/src/v2/guide/single-file-components.md
+++ b/src/v2/guide/single-file-components.md
@@ -21,6 +21,15 @@ Here's a simple example of a file we'll call `Hello.vue`:
 
 <img src="/images/vue-component.png" style="display: block; margin: 30px auto;">
 
+To take use of this component, import and register it like this:
+
+```js
+import Vue from 'vue';
+import Hello from './Hello.vue';
+
+Vue.component('hello', Hello);
+```
+
 Now we get:
 
 - [Complete syntax highlighting](https://github.com/vuejs/awesome-vue#source-code-editing)


chrisvfritz:
Thanks for the PR! I agree we can provide more information here, but I'd like to add to this in the coming week to provide a fuller picture, a using `Vue.component` isn't really the "normal" way of using `.vue` components. It can happen sometimes, but normally single-file components are registered inside other single-file components with something like:

``` js
import Hello from './Hello.vue'
 
export default {
  components: {
    Hello
  }
}
```

dlgoodchild:
I think clearer distinction between local and global components would help, and take it further by explaining why one would want a global component (`Vue.component`) vs a local component (`component: { ... }`).

arve0:
Adding some context; I was actually trying to mount my .vue component, so maybe add note about render function?

```js
let vm = new Vue({
  el: '#app',
  render: h => h(App)
})
```

sdras:
Hi all! 
In order to provide clarity, maybe this should be broken up into two sections: 
1) all the ways you could register a component, and when we recommend using what
2) Single File Components and their value

I think Single File Components are one of the most compelling things for developers about Vue, so it seems pertinent that we keep this section very clear and standalone without getting into the weeds of implementation details, aside from links to learn more. Particularly because it's also an entry point in the docs- people sometimes want to know what they are and are not yet ready to learn how to use them.

We already kind of show the [ways to register a component here](https://vuejs.org/v2/guide/components.html), but this page is very long and maybe it's worth having a dedicated page for it.

Thoughts?

chrisvfritz:
@sdras One of the next big tasks I have planned is to break out the components page, since it's so huge now - and better cover some concepts like component registration. 

But in the meantime, I like the idea of covering it briefly here and actually don't mind some duplication in the guide, especially for extremely stable features such as how components are registered.

sdras:
Sounds good! Let me know if you need help breaking it out

sdras:
Since we're breaking the component page out into sections, I'm going to go ahead and close this out, since it will probably create a conflict with the change. I'll try to keep tabs and make sure we include this way of registering a component in the improved section, though. Feel free to open an issue if you'd like to track it and make sure it's addressed. Thanks!

PR #1144, opened 2017-09-20
Bug in select2 wrapper component example (on options changed)

mrmorcs:
I found when using this - if the options change after rendering, the new set gets appended to the existing set rather than replacing them. Adding the empty() call in the options watch solved this.

I'm not sure how to update the jsfiddle so I linked to my fork of it, happy to change as needed!


chrisvfritz:
Thanks! I just forked your fork, as we like to keep the fiddles under member accounts for security reasons. 🙂

md5:
This approach ends up throwing out any additional options that are passed in via the `<slot>`, unfortunately.

PR #1146, opened 2017-09-22
Fix standalone vue.js cdn url

gabor-farkas:
The other one results in `vue.common.js:245 Uncaught ReferenceError: process is not defined` while loading


MartinKolarik:
@chrisvfritz I think the problem here is you didn't publish a new version after vuejs/vue#6618 so jsDelivr still uses `main`

MartinKolarik:
@gabor-farkas right now the correct link is https://cdn.jsdelivr.net/npm/vue@2.4.4/dist/vue.js and the short link should be fixed shorty.

PR #1147, opened 2017-09-22
Update components.md

tomwang1013:
To fix circular reference in webpack, change `require('./tree-folder-contents.vue').default` to `require('./tree-folder-contents.vue')`
my env:
```
nodejs: 6.9.1
vuejs: 2.3.4
webpack: 2.7.1
```


chrisvfritz:
Hmm, it hasn't been necessary in my projects to use `.default` with `.vue` files for some time, but I don't remember which piece was responsible for that change (e.g. upgraded Webpack, vue-loader, or a config change). I'll keep this open for now, then add a note when we find out.

PR #1148, opened 2017-09-22
style guide - missing no

rDr4g0n:
Missing No. Ya know, from Pokemon. Also the word "no" is missing in a sentence that should definitely have an extra "no".


chrisvfritz:
Thanks, but I already have this fixed locally actually. These docs are also not done yet, so not ready for contributions at thi point. _Very much_ a work-in-progress so there will still be many fixed mistakes and revisions before release. 🙂

rDr4g0n:
Ah, got it. I was aware that its a WIP, but not that it is very much still in progress.

PR #1151, opened 2017-09-25
Formatted the new styl code.

Jinjiang:
Just as the former discussion in #1000 & #1129 :-)


chrisvfritz:
Thanks. 😄  I had just come from a different project with _almost_ the same conventions. We really need to set up a linter here!

PR #1156, opened 2017-09-27
Add link to modifiers list

etnolover:
Add link to detailed info about modifiers


chrisvfritz:
Revised it a bit, with links to both features' modifiers.

PR #1159, opened 2017-09-29
Added links for AOT and tree-shaking

Jinjiang:
As @Justineo said https://github.com/vuejs/cn.vuejs.org/pull/531#discussion_r141079928 we are afraid some new guys don't know what AOT and tree-shaking is. So I added some links for explanation.

Thanks.


chrisvfritz:
Great idea! 🎉

PR #1160, opened 2017-09-30
Style Guide

chrisvfritz:
It's done(ish)! 😄 🎉  Here's the [hosted demo](https://vue-style-guide-dev.surge.sh/v2/style-guide/).

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -50,7 +50,7 @@ Some features of Vue exist to accommodate rare edge cases or smoother migrations
 
 **Component names should always be multi-word, except for root `App` components.**
 
-This prevents conflicts with existing and future HTML elements, since all HTML elements are a single word.
+This [prevents conflicts](http://w3c.github.io/webcomponents/spec/custom/#valid-custom-element-name) with existing and future HTML elements, since all HTML elements are a single word.
 
 {% raw %}<div class="style-example example-bad">{% endraw %}
 #### Bad
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -244,7 +244,7 @@ props: {
 
 **Always use `key` with `v-for`.**
 
-It's _always_ required on components, in order to maintain internal component state down the subtree. Even for elements though, it's a good practice to maintain predictable behavior, such as [object constancy](https://bost.ocks.org/mike/constancy/) in animations.
+`key` with `v-for` is _always_ required on components, in order to maintain internal component state down the subtree. Even for elements though, it's a good practice to maintain predictable behavior, such as [object constancy](https://bost.ocks.org/mike/constancy/) in animations.
 
 {% raw %}
 <details>
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -48,7 +48,7 @@ Some features of Vue exist to accommodate rare edge cases or smoother migrations
 
 ### Multi-word component names <sup data-p="a">essential</sup>
 
-**Component names should always be multi-word.**
+**Component names should always be multi-word, except for root `App` components.**
 
 This prevents conflicts with existing and future HTML elements, since all HTML elements are a single word.
 
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -773,7 +773,7 @@ components/
 
 Since editors typically organize files alphabetically, all the important relationships between components are now evident at a glance.
 
-You might be tempted to solve this problem differently, nesting all the search components under a "search" directory, then all the settings components under a "settings" directory. We only recommend this approach in very large apps of 100+ components, for these reasons:
+You might be tempted to solve this problem differently, nesting all the search components under a "search" directory, then all the settings components under a "settings" directory. We only recommend considering this approach in very large apps (e.g. 100+ components), for these reasons:
 
 - It generally takes more time to navigate through nested sub-directories, than scrolling through a single `components` directory.
 - Name conflicts (e.g. multiple `ButtonDelete.vue` components) make it more difficult to quickly navigate to a specific component in a code editor.
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -2,7 +2,7 @@
 type: style-guide
 ---
 
-# Style Guide
+# Style Guide <sup class="beta">beta</sup>
 
 This is the official style guide for Vue-specific code. If you use Vue in a project, it's a great reference to avoid errors, bikeshedding, and anti-patterns. However, we don't believe that any style guide is ideal for all teams or projects, so mindful deviations are encouraged based on past experience, the surrounding tech stack, and personal values.
 
@@ -28,7 +28,7 @@ These rules have been found to vastly improve readability in most projects. If y
 
 ### Priority C: Recommended
 
-Where multiple, equally good options exist, an arbitrary choice can be made to ensure consistency. In these rules, we describe each acceptable option and then suggest a default choice. That means you can feel free to make a different choice in your own codebase, as long as you're consistent and have a good reason. Please do have a good reason though! By adapting to the community standard, you will:
+Where multiple, equally good options exist, an arbitrary choice can be made to ensure consistency. In these rules, we describe each acceptable option and suggest a default choice. That means you can feel free to make a different choice in your own codebase, as long as you're consistent and have a good reason. Please do have a good reason though! By adapting to the community standard, you will:
 
 1. train your brain to more easily parse most of the community code you encounter
 2. be able to copy and paste most community code examples without modification
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -112,7 +112,7 @@ data: {
 
 We might want to reuse this component, allowing users to maintain multiple lists (e.g. for shopping, wishlists, daily chores, etc). There's a problem though. Since every instance of the component references the same data object, changing the title of one list will also change the title of every other list. The same is true for adding/editing/deleting a todo.
 
-Instead, we want each component instance to only manage its own data. For that to happen,each instance must generate a unique data object. In JavaScript, this can be accomplished by returning the object in a function:
+Instead, we want each component instance to only manage its own data. For that to happen, each instance must generate a unique data object. In JavaScript, this can be accomplished by returning the object in a function:
 
 ``` js
 data: function () {
@@ -287,7 +287,7 @@ In our experience, it's better to _always_ add a unique key, so that you and you
 <ul>
   <li v-for="todo in todos">
     {{ todo.text }}
-  </td>
+  </li>
 </ul>
 ```
 {% raw %}</div>{% endraw %}
@@ -302,7 +302,7 @@ In our experience, it's better to _always_ add a unique key, so that you and you
     :key="todo.id"
   >
     {{ todo.text }}
-  </td>
+  </li>
 </ul>
 ```
 {% raw %}</div>{% endraw %}
@@ -582,16 +582,13 @@ components/
 </summary>
 {% endraw %}
 
-These components lay the foundation for consistent styling and behavior in your application. They can only contain:
+These components lay the foundation for consistent styling and behavior in your application. They may **only** contain:
 
 - HTML elements,
 - other `Base`-prefixed components, and
 - 3rd-party UI components.
 
-They'll never contain:
-
-- global state (e.g. from a Vuex store)
--
+But they'll **never** contain global state (e.g. from a Vuex store).
 
 Their names often include the name of an element they wrap (e.g. `BaseButton`, `BaseTable`), unless no element exists for their specific purpose (e.g. `BaseIcon`). If you build similar components for a more specific context, they will almost always consume these components (e.g. `BaseButton` may be used in `ButtonSubmit`).
 
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1,8 +1,8 @@
 ---
-title: Style Guide
-type: guide
-order: 402
+type: style-guide
 ---
+
+# Style Guide
 
 This is the official style guide for Vue-specific code. If you use Vue in a project, it's a great reference to avoid errors, bikeshedding, and anti-patterns. However, we don't believe that any style guide is ideal for all teams or projects, so mindful deviations are encouraged based on past experience, the surrounding tech stack, and personal values.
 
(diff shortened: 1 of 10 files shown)


chrisvfritz:
Just doing a final code review, then I'll publish officially.

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+``` js
+data: {
+  listTitle: '',
+  todos: []
+}
+```
+
+We might want to reuse this component, allowing users to maintain multiple lists (e.g. for shopping, wishlists, daily chores, etc). There's a problem though. Since every instance of the component references the same data object, changing the title of one list will also change the title of every other list. The same is true for adding/editing/deleting a todo.
+
+Instead, we want each component instance to only manage its own data. For that to happen,each instance must generate a unique data object. In JavaScript, this can be accomplished by returning the object in a function:
```
phanan: Missing a space between `happen,` and `each`.

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+  data () {
+    return {
+      foo: 'bar'
+    }
+  }
+}
+```
+
+``` js
+// It's OK to use an object directly in a root
```
phanan: When it's "OK" to use an object in a root instance, I'd argue that it's still better to use a function for consistency and one less thing to remember. So maybe move this out of the "Good" section, for example as a note?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+<details>
+<summary>
+  <h4>Detailed Explanation</h4>
+</summary>
+{% endraw %}
+
+Detailed [prop definitions](https://vuejs.org/v2/guide/components.html#Prop-Validation) have two advantages:
+
+- They document the API of the component, so that it's easy to see how the component is meant to be used.
+- In development, Vue will warn you if a component is ever provided incorrectly formatted props, helping you catch potential sources of error.
```
phanan: provided _with_ incorrectly formatted props?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+```
+{% raw %}</div>{% endraw %}
+
+
+
+### Keyed `v-for` <sup data-p="a">essential</sup>
+
+**Always use `key` with `v-for`.**
+
+It's _always_ required on components, in order to maintain internal component state down the subtree. Even for elements though, it's a good practice to maintain predictable behavior, such as [object constancy](https://bost.ocks.org/mike/constancy/) in animations.
```
phanan: This sentence reads a bit weird to me. What's always required on components again?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+{% raw %}</details>{% endraw %}
+
+{% raw %}<div class="style-example example-bad">{% endraw %}
+#### Bad
+
+``` html
+<ul>
+  <li v-for="todo in todos">
+    {{ todo.text }}
+  </td>
```
phanan: Why `</td>`?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+#### Good
+
+``` html
+<ul>
+  <li
+    v-for="todo in todos"
+    :key="todo.id"
+  >
+    {{ todo.text }}
+  </td>
```
phanan: Why `</td>`?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+These components lay the foundation for consistent styling and behavior in your application. They can only contain:
+
+- HTML elements,
+- other `Base`-prefixed components, and
+- 3rd-party UI components.
+
+They'll never contain:
+
+- global state (e.g. from a Vuex store)
+-
```
phanan: What's this hyphen for?

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+|- VTable.vue
+|- VIcon.vue
+```
+{% raw %}</div>{% endraw %}
+
+
+
+### Single-instance component names <sup data-p="b">strongly recommended</sup>
+
+**Components that should only ever have a single active instance should begin with the `The` prefix, to denote that there can be only one.**
```
phanan: :/ I don't know about this. `The` seems weird to me, especially when used as a tag. I often go with something like `App` or `Site` instead, e.g. `AppFooter` or `SiteHeader`.

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1805 @@
...
+|- SearchButtonRun.vue
+|- SearchInputExcludeGlob.vue
+|- SearchInputQuery.vue
+|- SettingsCheckboxLaunchOnStartup.vue
+|- SettingsCheckboxTerms.vue
+```
+
+Since editors typically organize files alphabetically, all the important relationships between components are now evident at a glance.
+
+You might be tempted to solve this problem differently, nesting all the search components under a "search" directory, then all the settings components under a "settings" directory. We only recommend this approach in very large apps of 100+ components, for these reasons:
```
phanan: IMO the number `100+` is pretty arbitrary. How about just mention "apps with a large number of components" instead?

themes/vue/source/css/_style-guide.styl
```diff
@@ -0,0 +1,82 @@
+$style-guide-bad-bg = lighten(desaturate($red, 80%), 80%)
+$style-guide-bad-text = darken(desaturate($red, 80%), 20%)
+$style-guide-good-bg = lighten(desaturate($green, 80%), 85%)
```
phanan: It's kind of subjective, but personally I don't think the current background color is a bit too dark to indicate good. Maybe try something like `#dbf1d9`?

src/v2/style-guide/index.md
chrisvfritz: It'd be pretty controversial to discourage this, since we've demonstrated the pattern in half the pages of the docs for a long time. 😕

src/v2/style-guide/index.md
chrisvfritz: The "with" would only be appropriate if we also used an indirect object without "to". [source](https://english.stackexchange.com/questions/137660/provide-vs-provide-with)

src/v2/style-guide/index.md
chrisvfritz: Always using `key` with `v-for` is always required on components.

src/v2/style-guide/index.md
chrisvfritz: Because I forgot to change it when modifying the example. 😅  Good catch!

src/v2/style-guide/index.md
chrisvfritz: Thanks! Fixing.

src/v2/style-guide/index.md
chrisvfritz: Same deal here.

src/v2/style-guide/index.md
chrisvfritz: For adding more items that I never needed to add. Removing this - thanks!

src/v2/style-guide/index.md
chrisvfritz: `The` is the only one that unequivocally implies a single instance though. I've seen the `App` prefix commonly used for simple element wrappers, for example, like `AppButton`.

src/v2/style-guide/index.md
chrisvfritz: It _is_ somewhat arbitrary, in the sense that we're using `100` rather than `99` or `101`, but I worry that if we say "apps with a large number of components" people will just ask, "How many is a large number?" 😄   

I've worked very comfortably with well over a 100 well-named components in a single directory, so I figured this was relatively low, safe number for people to start considering breaking things out. What do you think?

themes/vue/source/css/_style-guide.styl
chrisvfritz: We can tweak the colors after publishing, but I think this is good enough for now. The specific color you suggested looks very bright to me, like it's trying to highlight something rather than blending into the background. And since the rest of the page is using more subtle, desaturated colors, a more complete overhaul might be necessary to make something like that fit in.

src/v2/style-guide/index.md
phanan: I got it, but wouldn't it read more natural if we revise the section like this?

```md
**Always use `key` with `v-for`.**
 
`key` with `v-for` is _always_ required on components, in order to maintain internal component state down the subtree. Even for elements though, it's a good practice to maintain predictable behavior, such as [object constancy](https://bost.ocks.org/mike/constancy/) in animations.
```

Since the previous sentence is actually a heading, I guess we can/should repeat the subject in the explaining paragraph.

src/v2/style-guide/index.md
phanan: > "How many is a large number?" 😄

"But that's the whole point – we've no idea! An app can be pretty complex with 50-ish components, when another with 150 components can still be considered simple enough. It depends on what you're building, your skills, your team size etc. In other words, _you_ know when to organize components into directories."

phanan: I guess we can rewrite this in a way to emphasize that this is subjective and user-dependent?

src/v2/style-guide/index.md
phanan: We don't discourage it, right? We can say that in the note, e.g.

"Since there's only one root Vue instance, it's OK to use a `data` object directly – in fact, that's what the demos on this site are doing. However, if you want consistency blahblah…"

Does it make sense?

src/v2/style-guide/index.md
phanan: Hm… when I say "the avatar," I don't mean there's only one avatar in the whole app right? Maybe it's _the_ avatar on the user profile? I don't know, this recommendation isn't very convincing to me TBH (but again maybe it's just me, as I'm not a native speaker).

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1802 @@
...
+
+## Priority A Rules: Essential (Error Prevention)
+
+
+
+### Multi-word component names <sup data-p="a">essential</sup>
+
+**Component names should always be multi-word, except for root `App` components.**
+
+This prevents conflicts with existing and future HTML elements, since all HTML elements are a single word.
```
Jinjiang: Here is a ref may be useful: [Valid custom element name in W3C Custom Elements Spec](http://w3c.github.io/webcomponents/spec/custom/#valid-custom-element-name)
The same concern for the name of W3C custom elements.

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1802 @@
...
+```
+{% raw %}</div>{% endraw %}
+
+
+
+### Component name casing in templates <sup data-p="b">strongly recommended</sup>
+
+**Component names should always be PascalCase in [single-file components](single-file-components.html) and string templates - but kebab-case in DOM templates.**
+
+PascalCase in templates improves readability, because it's easier to tell at a glance what's an element and what's a component. Unfortunately, PascalCase is invalid for a custom element in HTML, so a DOM template must still use kebab-case.
```
Justineo: I think using kebab-case for component names is also quite clear to me (HTML tag names don't contain `-`) and we can always use kebab-case in template strings/SFCs/DOM templates, which might be even more consistent IMO. PascalCase component name smells a little bit like JSX... What about “kebab-case in templates, PascalCase in JSX”？

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1802 @@
...
+3. **Render Modifiers** (changes the way the element renders)
+  - `v-pre`
+  - `v-once`
+
+4. **Global Awareness** (requires knowledge beyond the component)
+  - `id`
+
+5. **Unique Attributes** (attributes that require unique values)
+  - `ref`
+  - `key`
```
Justineo: Should `key` be together with `v-for` because of their correlation？

src/v2/style-guide/index.md
```diff
@@ -0,0 +1,1802 @@
...
+#### Good
+
+``` js
+props: {
+  greetingText: String
+}
+```
+
+``` html
+<WelcomeMessage greeting-text="hi"/>
```
Jinjiang: This example is a little strange for me because the tag name is PascalCase but attribute name is kebab-case. Maybe I prefer kebab-case tag name which is more natural for me. The second option is both PascalCases. The here is the 3rd. 😂

src/v2/style-guide/index.md
chrisvfritz: That could be true, but we do have an international userbase, so that's not a reason to dismiss what you're saying. Is it the specific prefix or the idea of the convention in general that's not convincing you?

src/v2/style-guide/index.md
chrisvfritz: I think removing it from good would imply that it's _not_ good, right? Why move it to a note, when this section is literally for demoing what's acceptable?

src/v2/style-guide/index.md
chrisvfritz: Sure, I don't mind updating.

src/v2/style-guide/index.md
chrisvfritz: I guess part of my problem here is that I personally see nested component directories as an anti-pattern at virtually _every_ size, even though many others would disagree. The problem is people reach for folders to help manage their complexity instead of giving their components better names, which has many more benefits. So I worry about giving apps that are too small an excuse to do this.

src/v2/style-guide/index.md
chrisvfritz: The PascalCase provides an extra layer of meaning in your app, by differentiating components from elements visually. Additionally, having the difference in convention between the two contexts (in-DOM components vs others) carries even more information, acting as a warning that you're in the DOM, so some limitations apply.

Using kekab-case everywhere doesn't have similar advantages, so "smells like JSX" isn't a good enough reason to reject a convention with clear benefits. 🙂

src/v2/style-guide/index.md
chrisvfritz: No, `key` isn't specific to `v-for`, but is used in a lot of contexts where uniquely identifying the element is useful.

src/v2/style-guide/index.md
chrisvfritz: Good idea!

src/v2/style-guide/index.md
chrisvfritz: Hmm, PascalCase is a convention used to identify instances in JavaScript. It applies to components, because they're really both JavaScript and instances. I think PascalCase would be strange for attributes, because they _don't_ have instances. Does that make sense?

Jinjiang: Ah, make sense. No problem. 👌

src/v2/style-guide/index.md
chrisvfritz: OK, I just made a small change here from:

> We only recommend this approach in very large apps of 100+ components [...]

to:

> We only recommend considering this approach in very large apps (e.g. 100+ components) [...]

Gives people a little more leeway. 🙂

chrisvfritz:
@phanan @Justineo I'm merging now so that we can publish the beta on Vue 2's anniversary. 😄  That doesn't mean unresolved discussion has to end though. For anything you're still not satisfied with, please _do_ open up an issue and we can discuss further. We won't be removing the beta tag until about a month from now.

src/v2/style-guide/index.md
Justineo: OK that makes sense. I'm thinking of putting them together would make it easier to check for `key` after we created a `v-for`.

PR #1168, opened 2017-10-02
Style guide: Update abbreviation rule to note avoidance

mikestreety:


chrisvfritz:
The rule titles only include the topic, rather than the specific suggestion. 🙂

mikestreety:
Ah ok, thanks @chrisvfritz - I did get a little confused as it says:

> Abbreviations in component names **strongly recommended**

Took me a while to work it out! Is this something I could raise as an issue? Great style guide by the way, massive help!

chrisvfritz:
Ah, I can definitely see how that could be confusing. Thanks for the further explanation! Maybe we should update it to:

> Full-word component names

What do you think?

mikestreety:
Works for me! 👍  updated!

chrisvfritz:
Thanks! 🎉

PR #1169, opened 2017-10-02
Update Index.md (bad example repeated as good)

SegersIan:
Drop the 
``` html
<p v-if="error">
  Error: {{ error }}
</p>
<div v-else>
  {{ results }}
</div>
```
In the **Good** example, the exact same example was given as in the **bad** example for the rule : 
**It's usually best to use `key` with `v-if` + `v-else`, if they are the same element type (e.g. both `<div>` elements).**

See here https://github.com/vuejs/vuejs.org/blob/master/src/v2/style-guide/index.md#good-24

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1524,14 +1524,6 @@ By default, Vue updates the DOM as efficiently as possible. That means when swit
 </div>
 ```
 
-``` html
-<p v-if="error">
-  Error: {{ error }}
-</p>
-<div v-else>
-  {{ results }}
-</div>
-```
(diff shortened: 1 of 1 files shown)


chrisvfritz:
This example is different, as it uses a `p` and `div`, rather than two `div`s.

SegersIan:
Now it makes sense, I was staring at it for ages, checking if I was missing the difference. Didn't pick up the <p> and <div> difference. My bad.

chrisvfritz:
No worries! 😄

chrisvfritz:
Re-opening, as I want to add a note above the example to highlight the difference.

PR #1176, opened 2017-10-03
Update _common.styl

Alex-Sokolov:
Deleted duplicated class


chrisvfritz:
It's there for a reason actually, to give the class the specificity it needs. 🙂

Alex-Sokolov:
Ok :)

PR #1178, opened 2017-10-03
Style guide: Fix JSX and single-file components links

etiennellipse:
The JSX and single-file components linking to Vue.js documentation are broken online.

I fixed them, keeping them relative to current path. Let me know if that's alright.

(tested locally with Hexo, which is dead simple to setup! nice)


chrisvfritz:
Thanks! This page was originally part of the guide, so those were left-over. 😅

PR #1182, opened 2017-10-04
Add Additional Resources to ecosystem dropdown

fakefarm:
I visited _vuejs.org_ for the purpose of visiting vue-loader documentation and noticed it wasn't listed. I feel it is a crucial aspect of developing vue and would suggest adding it alongside the ecosystem dropdown.

These three links didn't seem to fit the pattern of the other headers so I created another header. However, if others feel they can be nested under an existing header or one named differently (eg. - 'Getting Started'), I would be happy to make that change. 

Thank you all for your great work around this project. I'm happy to help however I can.

Dave

### DIFF
--- a/themes/vue/layout/partials/ecosystem_dropdown.ejs
+++ b/themes/vue/layout/partials/ecosystem_dropdown.ejs
@@ -14,7 +14,7 @@
         <li><a href="https://vue-loader.vuejs.org" class="nav-link" target="_blank">Vue Loader</a></li>
       </ul>
     </li>
-    <li><h4>Core Plugins</h4></li>
+    <li><h4>Core Libraries</h4></li>
     <li><ul>
       <li><a href="https://router.vuejs.org/" class="nav-link" target="_blank">Vue Router</a></li>
       <li><a href="https://vuex.vuejs.org/" class="nav-link" target="_blank">Vuex</a></li>
--- a/themes/vue/layout/partials/ecosystem_dropdown.ejs
+++ b/themes/vue/layout/partials/ecosystem_dropdown.ejs
@@ -26,7 +26,13 @@
       <li><a href="https://curated.vuejs.org/" class="nav-link" target="_blank">Vue Curated</a></li>
       <li><a href="https://github.com/vuejs/awesome-vue" class="nav-link" target="_blank">Awesome Vue</a></li>
     </ul></li>
+    <li><h4>Additional Resources</h4></li>
+    <li>
+      <ul>
+        <li><a href="https://vue-loader.vuejs.org" class="nav-link" target="_blank">Vue Loader</a></li>
+        <li><a href="http://vuejs-templates.github.io/webpack/" class="nav-link" target="_blank">Webpack template</a></li>
+        <li><a href="https://github.com/vuejs/vue-cli" class="nav-link" target="_blank">Vue CLI</a></li>
+      </ul>
+    </li>
   </ul>
 </li>
-
-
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I agree that tooling could probably be more prominent, so I made a slight modification, adding these resources to a new `Tooling` section. I've also reorganized the dropdown to follow the order that I think people might need these resources in their journey to learn Vue. I just want a couple other team members to sign off on this one first.

phanan:
LGTM. My only concern is the entire missing of Vue-templates now, with nowhere to be found throughout the docs. Also, with our docs often praised for being very newcomer-friendly, replacing it with only the most complex template might be – I don't know – a bit weird?

chrisvfritz:
@phanan 

> My only concern is the entire missing of Vue-templates now, with nowhere to be found throughout the docs.

We do, just indirectly. That same information is available in the Vue CLI readme, which [we link to](https://vuejs.org/v2/guide/installation.html#CLI) on the installation page.

> Also, with our docs often praised for being very newcomer-friendly, replacing it with only the most complex template might be – I don't know – a bit weird?

While that template has the most complex build process, it's also (perhaps counter-intuitively) the one you're least likely to have issues with as a relative beginner, since it does so much for you out-of-the-box and is the best-documented.

My experience is that _absolute_ beginners will have a hard time with a build process, no matter what. If they've never used a build tool or module system, then probably every item in the Ecosystem dropdown below the `Devtools` will be confusing to them.

For people who are at least a _little_ more advanced, they often become frustrated with Browserify or the `webpack-simple` template. We simply don't have as robust support for Browserify, so there are more bugs and fewer features. And `webpack-simple` actually requires knowing more about Webpack, ironically. 😄  If they want to use a CSS preprocessor, for example, they have to do quite a lot of work and know more about the Webpack ecosystem. The `webpack` template, on the other hand, describes in detail how to work with CSS preprocessors and doesn't require them to change any configuration. They just install the appropriate loaders and peer dependencies, e.g. `sass-loader` and `node-sass`.

Does that make sense?

chrisvfritz:
Also a note for the future: I'm thinking we'll be able to replace the link to the Webpack template with one to Vue CLI 3, once that's out, since it will offer the same kind of simple dev experience, but hide away the stuff that looks complex unless you actually need it.

themes/vue/layout/partials/ecosystem_dropdown.ejs
```diff
@@ -1,25 +1,32 @@
...
     </ul></li>
+    <li><h4>Tooling</h4></li>
+    <li>
+      <ul>
+        <li><a href="https://github.com/vuejs/vue-devtools" class="nav-link" target="_blank">Devtools</a></li>
+        <li><a href="https://vuejs-templates.github.io/webpack" class="nav-link" target="_blank">Webpack Template</a></li>
+        <li><a href="https://vue-loader.vuejs.org" class="nav-link" target="_blank">Vue Loader</a></li>
+      </ul>
+    </li>
     <li><h4>Core Plugins</h4></li>
```
yyx990803: Also "Core Libraries" here sounds better since `vue-server-renderer` isn't a plugin

yyx990803:
LGTM

themes/vue/layout/partials/ecosystem_dropdown.ejs
chrisvfritz: Good catch! Updated.

themes/vue/layout/partials/ecosystem_dropdown.ejs
```diff
@@ -1,25 +1,32 @@
...
   <a class="nav-link">Ecosystem</a><span class="arrow"></span>
   <ul class="nav-dropdown">
-    <li><a href="https://github.com/vuejs/roadmap" class="nav-link" target="_blank">Roadmap</a></li>
     <li><h4>Help</h4></li>
     <li><ul>
       <li><a href="https://forum.vuejs.org/" class="nav-link" target="_blank">Forum</a></li>
       <li><a href="https://chat.vuejs.org/" class="nav-link" target="_blank">Chat</a></li>
-      <li><a href="https://github.com/vuejs-templates" class="nav-link" target="_blank">Templates</a></li>
+    </ul></li>
+    <li><h4>Tooling</h4></li>
```
fakefarm: I agree, _Tooling_ is a better header. Thanks for updating!

PR #1183, opened 2017-10-05
fix minor flaw

pierreneter:


chrisvfritz:
Computed properties are not actually called with `()`. The function definition is actually an automatically cached getter.

pierreneter:
@chrisvfritz are you sure?
Please check, https://codesandbox.io/s/2v2mzv899y

chrisvfritz:
I didn't see a computed property in the page you linked to. Perhaps you're thinking of methods, which _are_ callable functions?

pierreneter:
So does it automatically for `methods`?

chrisvfritz:
No, methods are not cached and must be called to be used. The only time you omit the parentheses is when you want the method to be called later, in response to an event. For example:

``` html
<button @click="myMethod">
  Do something
</button>
```

or:

``` html
document.addEventListener('keydown', this.myMethod)
```

PR #1185, opened 2017-10-05
Change complex expressions in templates example

daniiell3:
Change complex expressions in templates example using filters instead of computed properties

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1078,9 +1078,9 @@ In JavaScript, splitting objects with multiple properties over multiple lines is
 
 ### Complex expressions in templates <sup data-p="b">strongly recommended</sup>
 
-**Component templates should only include simple expressions, with more complex expressions refactored into computed properties or methods.**
+**Component templates should only include simple expressions, with more complex expressions refactored into filters or computed properties.**
 
-Complex expressions in your templates make them less declarative. We should strive to describe _what_ should appear, not _how_ we're computing that value. Computed properties and methods also allow the code to be reused.
+Complex expressions in your templates make them less declarative. We should strive to describe _what_ should appear, not _how_ we're computing that value. Filters and Computed properties also allow the code to be reused.
 
 {% raw %}<div class="style-example example-bad">{% endraw %}
 #### Bad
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks for the suggestion, but I don't really see value added here. We've just refactored a computed property to a filter, making the example more complex. Filters are generally just for simple text transformations that have nothing to do with the business logic of the app, so it's often not appropriate and generally more work to convert a complex template expression to a filter, rather than a computed property or method.

daniiell3:
I respect your opinion, but I disagree. Whenever we use filters, we follow one of the major principles of the functional paradigm, immutability. Whenever we use computed properties and methods, the chance of generating side effects in our app is too great. You said that filters are just for text transformations, exactly the context of the example. Thank you for your time.

chrisvfritz:
> Whenever we use filters, we follow one of the major principles of the functional paradigm, immutability. Whenever we use computed properties and methods, the chance of generating side effects in our app is too great.

A couple notes here:

- Immutability isn't actually a principle, but rather a _tool_ commonly used in service of a principle: predictability.

- Filters do not guard against side effects (including mutability). In the case of strings, which have no self-mutating methods, one is only protected from the side effects from an explicit assignment like:

  ``` js
  this.fullName = 'new string'
  ```

  I've never personally encountered a case where someone did this accidentally, but the soon-to-be-released `eslint-plugin-vue` will actually be able to flag _any_ side effects (including mutation) in computed properties, so that it really will be impossible if you're worried.

  When we move beyond strings to arrays or objects, functions are passed a reference rather than an immutable value in JS, so filters are actually able to mutate data properties just like anything else. And if you're really passionate about predictability, then you're very likely also using a central state management solution like Vuex. In this case, you could import the store anywhere in your app (including a filter) and call an action, triggering a mutation. It's not good practice, but once again, the filter won't protect you.

  The point is, if you've found there's a large chance of accidentally generating side effects when using computed properties or methods, the problem isn't computed properties or methods - it's developer education. By overusing filters, you only avoid one of the more obvious ways of generating side effects, ignoring the larger problem of training developers about what qualifies as a side effect, when to avoid them, and how.

> You said that filters are just for text transformations, exactly the context of the example.

Reading the rest of the sentence, you'll notice I said filters are for "text transformations **that have nothing to do with the business logic of the app**." This filter is specific to `fullName`, rather than strings in general, so it's business logic and best suited to a computed property. 

Computed properties are also automatically cached, since they can track their own dependencies, so will offer improved performance over filters in every scenario.

PR #1186, opened 2017-10-05
Fix: Demystified markdown logic

prog-rajkamal:
I have changes jsfiddle link with a similar jsfiddle. 

The difference between two jsfiddle is in javascript code, `marked()` has a comment mentioning its source as external library. 

This is important because it is easy to overlook that `marked()` comes from an external library instead of Vue, especially for a beginner.

### DIFF
--- a/src/v2/examples/index.md
+++ b/src/v2/examples/index.md
@@ -6,4 +6,4 @@ order: 0
 
 > Dead simple Markdown editor.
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Good idea! I made a slight modification: added the dependencies to the HTML, so that it would be more apparent that we're using external libraries here.

PR #1193, opened 2017-10-06
additional details and example of component data arrow functions

brianjorden:
I've noticed a couple tutorial sites using arrow functions for component data. Thought it would be good to add it in as an example to avoid for the style guide as it is already part of the API documentation. Long ago I was using arrow functions personally and didn't understand why I was having trouble with certain things. Hopefully, this ends up helping someone else.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -118,6 +118,9 @@ data: function () {
   }
 }
 ```
+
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Rather than appending to this rule, I think I might add a more general Priority D rule about using arrow functions for component option values. I'll open a new issue for that.

PR #1198, opened 2017-10-08
Update custom-directive.md

HeGangSZ:
I suggest add this params demo because i can not understand when i saw this part,i think other people will have the same confusion


chrisvfritz:
Sounds good to me. 🙂  Thanks for helping improve this!

PR #1199, opened 2017-10-09
Portuguese language

JoseBorges:
Better identification for the Brazillian language, since there are several major diferences between Portuguese (Portugal) from Portuguese (Brazil).

### DIFF
--- a/themes/vue/layout/partials/language_dropdown.ejs
+++ b/themes/vue/layout/partials/language_dropdown.ejs
@@ -5,7 +5,7 @@
     <li><a href="https://jp.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">日本語</a></li>
     <li><a href="https://ru.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Русский</a></li>
     <li><a href="https://kr.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">한국어</a></li>
-    <li><a href="https://br.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Português</a></li>
+    <li><a href="https://br.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Português (Brazil)</a></li>
     <li><a href="https://fr.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Français</a></li>
   </ul>
 </li>


chrisvfritz:
Should this be "Brasil" (with an "s")?

phanan:
The country name is Brazil I think.

themes/vue/layout/partials/language_dropdown.ejs
```diff
@@ -5,7 +5,7 @@
     <li><a href="https://jp.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">日本語</a></li>
     <li><a href="https://ru.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Русский</a></li>
     <li><a href="https://kr.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">한국어</a></li>
-    <li><a href="https://br.vuejs.org/<%- page.path %>" class="nav-link" target="_blank">Português</a></li>
```
ludo237: The official language is [Portugues](https://en.wikipedia.org/wiki/Brazil) so this specification is meaningless

JoseBorges: @ludo237 American English is the same as England English? It's not meaningless.

ludo237: I didn't want to be mean but it's not suitable to create a menu that groups every dialectical differences. Generally speaking the only two languages that have this treatment are English (UK and US) and Chinese (simplified and standard)

phanan: It's not meaningless, but I wouldn't say it's too necessary either. Similar to English, you don't specify "American English" in a dropdown _unless_ you have both American and British English.

chrisvfritz:
@phanan I thought the name was spelled with an "s" in Portuguese though, which might be more appropriate since the everything in that dropdown is in its native language.

@ludo237 It really depends on how different those dialects are - and I don't have that information. Would someone from Portugal who would otherwise be interested in a Portuguese translation decide to just read the English version if the dialect is Brazilian? 

Another advantage is that people may be less likely to submit PRs to that translation that target the wrong dialect.

themes/vue/layout/partials/language_dropdown.ejs
JoseBorges: We Portuguese/Brazilians are used to diferenciate our native Portuguese languages, that's why i did the PR. 

Just to make clear that it's Brazilian Portuguese, nothing more nothing less. :) 

As it seems this is not an issue i can live with it as it is.

JoseBorges:
I guess that due to project members reactions it's best to leave the language as it is.

ludo237:
@JoseBorges it's not `due to project members reactions` but it's about the impact of the pr on the project itself.

If we had allowed this we would have to create a new standard for the drop down, allowing every different dialect to be present, which is not possible. Don't get me wrong, I respect your differences in speaking but a Brazilian can understand a native Portuguese as well as I can understand an native speaker from United Kingdom

PR #1202, opened 2017-10-10
Add some comments for `import` function

Justineo:
...in case some readers are not familiar with it.


ludo237:
This PR does not bring anything useful to the documentation

Justineo:
@ludo237 `import` function is rather new so some readers might not be familiar with it. This piece of code is demonstrating “You can also return a `Promise` in the factory function”, I think it's useful for better understanding of the guide.

PR #1203, opened 2017-10-10
Update Riot comparison

MartinMuzatko:
* Animation - this is no longer true, they have https://github.com/riot/animore
* Routing - the router API has been extended to allow for subroutes and parameters just like Vue does.

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -222,9 +222,7 @@ It is also totally feasible to offer deeper integration between Vue with Web Com
 
 ## Riot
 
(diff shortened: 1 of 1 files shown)


MartinMuzatko:
I'm not sure if one can see the more loosely defined expressions in Riot as an advantage or disadvantage.
I however, found that it is a lot simpler to do bindings in riot. `<todo-list text={text}>` vs `<todo-list :text="text">`.

Riot is generally more flexible and requires less syntax to learn. Same for event handlers:
`<div onclick={action}></div>` vs `<div @click="action()"></div>`

However. One advantage to note, would be, that the event handling in Vue has more benefits (e.g. `@keyup.enter` or `@click.prevent`,  and even `@click="doSomething()"` etc) which is not possible in riot.

chrisvfritz:
Thanks! I removed the last advantage added for Vue, simply because there are many, many ways that Vue's API offers more convenience and I worry it can sound like nit-picking when we don't highlight small little features of Riot that may offer an advantage over Vue.

PR #1205, opened 2017-10-10
Style guide: Use Vue.delete instead of replacing state

timwis:
Unless there was some reason an immutable approach was used?


chrisvfritz:
This looks like an unnecessary optimization to me. `filter` works for this case and keeps the code simpler, so that users reading the example can focus on what's being demoed.

PR #1210, opened 2017-10-11
Documented how to use v-cloak together with slots

randohinn:
One needs to use v-cloak in the file that defines the slot contents. This has not been documented before.


chrisvfritz:
Thanks, but this slot behavior isn't actually specific to `v-cloak`. Other directives and attributes also have no effect on `slot`.

PR #1217, opened 2017-10-14
Updating computed.md. 

pranayaryal:
Gave a little bit more comment for the 'debounce' functionality so that a newbie understands it better.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -243,7 +243,8 @@ var watchExampleVM = new Vue({
           })
       },
       // This is the number of milliseconds we wait for the
-      // user to stop typing.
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks, but I worry this is actually more confusing. Debounce isn't a straight-forward "delay" - we really are waiting for the user to stop typing for a certain number of milliseconds. What missing piece of information are you trying to provide?

pranayaryal:
You are right Chris. I somehow overlooked the good explanation about debounce at the beginning of the method. You may ignore my commit.

chrisvfritz:
No worries! Thanks for the PR. 🙂

PR #1225, opened 2017-10-18
Clarify how to define filters

flybayer:
I was confused about how to do this, [as are others](https://forum.vuejs.org/t/possible-mistake-in-filters-documentation-failed-to-resolve-filter/14966).


Labradoodle-360:
Thanks for taking the time to submit a PR to update these docs. Since it had `new Vue({...` at the top, it was misleading and I tried to register as part of the main Vue instance as well.

chrisvfritz:
Thanks! I made a few tweaks, particularly so that the component example is not specific to a single-file component or specific module system, but this is a great change. 🙂

PR #1228, opened 2017-10-19
Update typescript.md

wangyi7099:
The typesrcipt error cann't be checked by Vuter until you assert the  Object type to ComponentOptions<Vue>.

### DIFF
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -64,9 +64,10 @@ Note that when using Vetur with SFCs, type inference will be automatically appli
 </template>
 
 <script lang="ts">
-export default {
-  // type inference enabled
-}
+import Vue, {ComponentOptions} from 'vue'
+  export default {
+    // type inference enabled
+  } as ComponentOptions<Vue>
 </script>
 ```
 


wangyi7099:
Now, type can be checked by typescript.

ktsn:
The casting should not be needed since vetur wraps the object literal with `new Vue` under the hood...

/cc @octref

HerringtonDarkholme:
@ktsn is right. 
Vetur wraps the object literal in `Vue.extend`. If you want to make external tools like ts-loader to check script in SFC, I would still suggest you use `Vue.extend` since it can infer more properties on `this`.

chrisvfritz:
@ktsn @HerringtonDarkholme I'll take whatever action you both recommend. 🙂

octref:
The original issue: https://github.com/vuejs/vetur/issues/473

image

@wangyi7099 Custom properties can also be used in Vue (like when using Apollo), so Vetur (or the underlying type definition) does not restrict its usage.

@chrisvfritz You can close it I believe.

chrisvfritz:
@octref OK, thanks for taking a look. 🙂

wangyi7099:
@octref   I'm  fool ...  Thank you  for your comment!

PR #1234, opened 2017-10-21
Add a note about `this` type inference

ktsn:
Although the recommended configuration has `strict: true`, I saw some people does not include it in their config and confused why `this` type is not inferred.

I guess the beginners of TypeScript may not understand what is `stricter inference for data properties on this`. So I added a more description about that.

### DIFF
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -31,6 +31,8 @@ We also plan to provide an option to scaffold a ready-to-go Vue + TypeScript pro
 }
 ```
 
+Note that you have to include `strict: true` (or at least `noImplicitThis: true` which is a part of `strict` flag) to leverage type checking of `this` in component methods otherwise it is always treated as `any` type.
+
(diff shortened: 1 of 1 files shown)


galenyuan:
What about add a mention that `export default Vue.extend({ ..ComponentOption })` is required?
I tried like this
image
but complier throw errors
image

ktsn:
@galenyuan It's already in docs...

> To let TypeScript properly infer types inside Vue component options, you need to define components with Vue.component or Vue.extend:

galenyuan:
@ktsn but there is

> Note that when using Vetur with SFCs, type inference will be automatically applied to the default export, so there’s no need to wrap it in Vue.extend:

in next paragraph, it's confused..

HerringtonDarkholme:
We have already removed that paragraph. Please patiently wait for deployment.

ktsn:
FYI #1226

galenyuan:
Aha, thanks guys~

PR #1244, opened 2017-10-31
Add Vietnamese option to language dropdown

phainv:
First of all, Thanks for having the document suport Vietnamese. [Link Repo](https://github.com/vuejs-vn/vuejs.org)
But. It does not display in the language selection. might it not yet perfected (beta version). But it is essential to understand it. 
Thanks!


chrisvfritz:
@phanan I'll leave this up to you. Merge if you think it's a good idea.

phanan:
Thanks both! Merged.
@phainv Would you mind creating a similar PR in the Vietnamese repo as well?

phanan:
Actually scratch that – we don't need a new menu item in Vi version.

phainv:
@phanan I thought so. Because by default its language is set to Vietnamese

PR #1254, opened 2017-11-04
Relocate myself to Munich

phanan:


eddyerburgh:
How's it going over there?

phanan:
Mixed feelings. I've not started work yet, still airbnb'ing and looking for an apartment.

PR #1256, opened 2017-11-06
Minor document changes

BitYog:


chrisvfritz:
I appreciate you taking the time, but this actually reads a bit more awkwardly to me. Specifically:

- I think the added commas break up the flow unnecessarily.
- `an MIT` is actually correct, because while `M` is a consonant, it begins with a vowel sound when pronounced as an acronym.
- `the project` or `this project` are both acceptable.

BitYog:
Alright.

PR #1259, opened 2017-11-07
Update directive documentation with tip relating to v-html and scoped styles

dlgoodchild:
Adding a tip relating to the fact that v-html content is unaffected by scoped styling, and explaining an example way around it.


chrisvfritz:
Thanks! I only tweaked the content slightly to include a note about CSS modules. 🙂

dlgoodchild:
Hah, cool. There's a super minor typo, where it says "In you want" instead of "If you want".

chrisvfritz:
Thanks! Just fixed it. 😅

PR #1260, opened 2017-11-07
docs: add CodePen as a supported social network

phanan:
This PR adds CodePen as a supported social network following @sdras's 66958937edbdc782c2161feb86768ef094bea6bb.


sdras:
thanks!

PR #1262, opened 2017-11-09
Fix code tags

Justineo:


chrisvfritz:
Great little refinements all around @Justineo. Thank you for these PRs! 😄

PR #1263, opened 2017-11-09
Fix header link

Justineo:
Fix the problem that headers are covered by the top menu and also keep them clickable.


themes/vue/source/js/common.js
```diff
@@ -337,8 +336,14 @@
...
-      link.parentNode.insertBefore(wrapper, link)
-      wrapper.appendChild(link)
+
+      var nodes = link.childNodes
+      for (var i = 0; i < nodes.length; i++) {
+        var node = nodes[i]
+        if (node !== wrapper) {
+          wrapper.appendChild(node)
+        }
+      }
```
Jinjiang: I found that just remove these additions here and all the modifications in `page.style` could work (in Safari 11 in my Mac), which means just keeping `setAttribute('data-scroll', '')` only in function `makeHeaderClickable` is OK.

themes/vue/source/css/_settings.styl
```diff
@@ -23,7 +23,7 @@ $header-inner-height = 41px
 $heading-padding-vertical = 10px
 $header-height = $header-inner-height + $heading-padding-vertical * 2
 $mobile-header-height = 40px
-$heading-link-padding-top = $content-padding-top
-$mobile-heading-link-padding-top = $content-padding-top
+$heading-link-padding-top = $header-height + $content-padding-top
```
Jinjiang: For the former mistake, I advice to add some comment here and `function makeHeaderClickable()` (whatever it will become finally).

Justineo:
@Jinjiang Done.

chrisvfritz:
Had to revert this unfortunately to fix #1273, but I'm happy to accept another similar PR.

Justineo:
@chrisvfritz Fixed that in #1280.

PR #1264, opened 2017-11-09
Add description about `scope` in API doc

### LINKED ISSUE #1223: The deprecation of `scope` prop should be addressed in guide
Since 2.5.0 we just replaced all `scope` with `slot-scope` in the guide, which may cause confusion among users. We should clearly address that `scope` is deprecated and `slot-scope` should be used instead.

Justineo:
This will probably resolve #1223.

Not sure if we should state this in the section of scoped slots in the guide.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -2121,15 +2121,25 @@ type: api
 
 ### slot-scope
 
+> New in 2.5.0+
+
 - **Expects:** `function argument expression`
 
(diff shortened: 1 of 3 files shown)


Jinjiang:
Seems you have also committed the changes of #1263 together. I suggest to extract them out first.
Thanks.

Justineo:
@Jinjiang done.

PR #1266, opened 2017-11-09
feat: add auto deployment with surge and circleCI

pi0:
Let's make deployments faster and hassle-free. With this PR every commit into master branch will be generated using hexo and deployed to surge. (Currently https://vuejs-fa.surge.sh but can be any custom domain too)

/ping @chrisvfritz What's your idea about this approach and do you see any problems we may encounter in future?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,38 @@
+version: 2
+jobs:
+  build:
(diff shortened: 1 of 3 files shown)


pi0:
Opps Sorry! Wrong target :))

PR #1268, opened 2017-11-09
Add in cookbook template

sdras:
Creating a template for common use and guidelines for contributing to The Cookbook. Most of this is a standardization of the Adding Instance Properties example from @chrisvfritz.

Some things to note, entirely up for debate:
- I detailed that some sections are required, and some are optional. I think smaller examples will not need as much detail but more advanced concepts do.
- The order is set to 1000 to make it last, but I'm unfamiliar with this ordering schema so I may need guidance there.

### DIFF
--- a/src/v2/cookbook/cookbook-contributions.md
+++ b/src/v2/cookbook/cookbook-contributions.md
@@ -1,0 +1,59 @@
+---
+title: Cookbook Contributions
+type: cookbook
+order: 1000
+---
+
+## What we're looking for
+
+The Cookbook gives developers examples to work off of that both cover common or interesting use cases, and also progressively explain more complex detail. Our goal is to move beyond a simple introductory example, and demonstrate concepts that are more widely applicable, as well as some caveats to the approach. If you're interested in contributing, please follow the template below as much as possible. Some sections are required, and some are optional.
+
+## Simple Example
+
+_required_
+
+- Articulate the problem in a sentence or two.
+- Explain the simplest possible solution in a sentence or two.
+- Show a small code sample.
+- Explain what this accomplishes in a sentence.
+
+## Details about the Value
+
+_required_
+
+- Address common questions that one might have while looking at the example. (Blockquotes are great for this)
+- Show examples of common missteps and how they can be avoided.
+- Show very simple code samples of good and bad patterns.
+- Discuss why this may be a compelling pattern. Links for reference are not required but encouraged.
+
+## Real-World Example
+
+_required_
+
+Demonstrate the code that would power a common or interesting use case, either by:
+- Walking through a few terse examples of setup, or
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/cookbook-contributions.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Alternative Patterns
+
+_optional, except when the section above is provided_
+
+This section is required when you've provided the section above about avoidance. It's important to explore other methods so that people told that something is an antipattern in certain situations are not left high and dry. In doing so, consider that the web is a big tent and that many people have different codebase structures and are solving different goals. Is the app large or small? Are they integrating Vue into an existing project, or are they building from scratch? Are their users only trying to achieve one goal or many? Is there a lot of asyncronous data? All of these concerns will impact alternative implementations. A good cookbook recipe gives developers this context.
```
chrisvfritz: I think "left high and dry" might be difficult for non-native speakers to understand. How about "left wondering"?

src/v2/cookbook/cookbook-contributions.md
```diff
@@ -0,0 +1,59 @@
...
+The Cookbook gives developers examples to work off of that both cover common or interesting use cases, and also progressively explain more complex detail. Our goal is to move beyond a simple introductory example, and demonstrate concepts that are more widely applicable, as well as some caveats to the approach. If you're interested in contributing, please follow the template below as much as possible. Some sections are required, and some are optional.
+
+## Simple Example
+
+_required_
+
+- Articulate the problem in a sentence or two.
+- Explain the simplest possible solution in a sentence or two.
+- Show a small code sample.
+- Explain what this accomplishes in a sentence.
```
chrisvfritz: Hm, sometimes I like to include a code example demonstrating the problem before I go into the solution.

Also, maybe these should be numbered to imply that they are all required and should generally be followed in order.

sdras: Hm, what if the flow of how to explain that code in particular doesn't work that way? Perhaps it's possible to do ol and say that the order is "strongly suggested, but not imperative". Actually imperative has other meanings in code haha, so maybe... "vital".

src/v2/cookbook/cookbook-contributions.md
sdras: Good call! I need to get in this mindset of internationalization

sdras:
I've updated the PR with some of the suggestions above, as well as some additional improvements:
- change the wording of the english slang
- update the `ul`s to `ol`s, but with a statement that the ordering is strongly suggested but not required (flexibility for times when the flow makes more sense another way)
- put in a requirement for filing an issue, under a tag I just created called _cookbook idea_
- small edits for clarity

Happy to update further! Let me know what you think.

Issue #1250, opened 2017-11-02
Clarify "falsy" values


Justineo:
https://vuejs.org/v2/guide/syntax.html#Attributes

> It also works for boolean attributes - the attribute will be removed if the condition evaluates to a falsy value

In fact, binding an attribute to `0` or `NaN` won't result in removing the attribute. It's not a normal case but the doc should be more precise on this IMO.

via https://github.com/vuejs/cn.vuejs.org/issues/586

chrisvfritz:
Interesting. I'm glad you brought this up, but I'm wondering if it's a documentation issue or a bug. @yyx990803 Is this intentional and if so, why the deviation?

yyx990803:
Yeah we should be more precise here - it only removes the value if it is `null`, `undefined` or `false`: https://github.com/vuejs/vue/blob/dev/src/platforms/web/util/attrs.js#L41-L43

chrisvfritz:
@yyx990803 Thanks! Can you share the reasoning for that? I'd like to provide a justification for the behavior in the more precise explanation.

yyx990803:
Well, if the value is 0, most likely the user does want to display `0`. The case for `NaN` is subject to debate, but usually it is a result of error and probably better to have that shown in the output so the user can actually spot it.

chrisvfritz:
I see. Thanks!

sdras:
Fixed!

PR #1271, opened 2017-11-10
Clarify falsy conditions for attribute description

sdras:
In #1250, Evan clarifies what values are considered falsy, so updating the attribute description to reflect that and provide a little more clarity.


chrisvfritz:
@sdras Thanks for initiating this! I made a minor modification to remove `falsy` altogether and tie the explanation to the example a little more. Feel free to merge if you agree with the change. 🙂

sdras:
[I see what you did there](http://cdn.smosh.com/sites/default/files/legacy.images/smosh-pit/032011/seewhat-deniro.jpg). Cool, I can dig it :)

Issue #1246, opened 2017-10-31
Styleguide: Template / Options consistancy


Haeresis:
Hi @chrisvfritz.

In the Styleguide, the rules about options order said the correct order in a normal case is: 

```js
module.exports = {
    ...,
    data: function () {
        return {}
    },
    ...,
    template: `<div>
        <div></div>
    </div>`
}
```

but, the rules about top-level element order in SFC said:

> [Single-file components](../guide/single-file-components.html) should always order `template`, `script`, and `style` tags consistently, with `<style>` last.

So the way this sentence is written suggest that order:

```htm
<template>
    <div>
        <div></div>
    </div>
</template>

<script>
module.exports = {
    data: function () {
        return {}
    }
}
</script>
```

This is exactly the inverse of previous rules.

**So I suggest changing this sentence by:**

> [Single-file components](../guide/single-file-components.html) should always order `script`, `template`, and `style` tags consistently, with `<style>` last.

**and change good example order place:**

``` html
<!-- ComponentA.vue -->
<script>/* ... */</script>
<template>...</template>
<style>/* ... */</style>

<!-- ComponentB.vue -->
<script>/* ... */</script>
<template>...</template>
<style>/* ... */</style>
```

example before:

``` html
<!-- ComponentA.vue -->
<template>...</template>
<script>/* ... */</script>
<style>/* ... */</style>

<!-- ComponentB.vue -->
<template>...</template>
<script>/* ... */</script>
<style>/* ... */</style>
```

That not said the others order is not possible, but that bring consistency in rules

What is your thought? (Yeah I know is not a horrible problem, but if it's just little better, why not?)

Haeresis:
I just read this

```
I go template -> style -> script
Because I often go back and forth between layout and styling
While the component definition itself rarely changes
```

chrisvfritz:
Hm, I can see the logic here. You've convinced me. 😄 If you submit a PR to the Style Guide removing the part that recommends a specific order, it will be accepted.

Issue #1245, opened 2017-10-31
Styleguide: Priority C: Attributes order


Haeresis:
Hi @chrisvfritz,

I read in the `style-guide/index.md` i am currently translated the following thing:

`v-for` should be before `v-if` in the order of document.

— https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended

When I parse myself an HTML Vue template, I check if the `v-if` expression of current element I read is `true`  **before** start to read more about this element and all its child elements.

For exemple, in this case:

```html
<li v-if="..." v-for="...">...</li>
<li v-else-if="..." v-for="...">...</li>
<li v-else-if="..." v-for="...">...</li>
<li v-else-if="..." v-for="...">...</li>
<li v-else-if="..." v-for="...">...</li>
<li v-else >...</li>
```

I want first know witch case I want **before** know more about what this element do (read `v-for` and continue to parse this subpart).

**So what is the reason of this order if it's not the reading flow?**

I see in the original Markdown document the two have `2.`  : 

```md
2. **Rendu de liste** ...
...
2. **Conditionals** ...
```

I think it's maybe because of I just explained. But, in the HTML rendering of documentation, there is no way to allows two points of an « ordered list (`<ol>`) » to have twice « 2. » just like `.md` file.

What's your throughs?

EDIT : maybe the good practice is never use `v-if` with `v-for` but I not still read this in styleguide and in my tested case, `v-if` work fine with `v-for` on the same tag.

Haeresis:
I just see this : 

![v-for before v-if](https://media.discordapp.net/attachments/325477692906536972/374891599928360971/unknown.png)

So if I use `v-if` on the same item as `v-for` it's more like a `filter` than really a `case` system so, we first read the loop information **before** read filter condition.

I have my answer :) I will transform my issue name like a question to allows others people find this answer.

chrisvfritz:
@Haeresis Thanks! Do you think we should make this its own style guide rule, to add that information you were missing?

Haeresis:
I don't know. If I would writing something, that could be something like that but I'm **believe** is better, I don't **know** if it better.

I need your advice @chrisvfritz ;)

```md
### `v-for` and `v-bind` on the same element <sup data-p="a">???</sup>

**Always prefer a computed property instead of usage of `v-if` as filter in a `v-for` loop that displays only one case.**

[...]

PR #1275, opened 2017-11-11
[Doc EN]: remove `Single-file component top-level element order` from  `style-guide/index.md`

Haeresis:
As discussed below with @chrisvfritz, we removed this part which really depends on prefered usage of `.vue` file.

> If you submit a PR to the Style Guide removing the part that recommends a specific order, it will be accepted.

— https://github.com/vuejs/vuejs.org/issues/1246#issuecomment-343671331


Haeresis:
I'm not sure you watch closed issue (and it's normal) so I highlight you an answer I do about an adding in Styleguide : 
— ...

chrisvfritz:
Thanks! I think I miscommunicated though. 😅 I actually meant only remove the part that dictates a _specific_ order. I still want to keep the rule suggesting a consistent order within the project, because I still think consistency is helpful. Does that make sense?

chrisvfritz:
Also, thanks for the ping about that closed issue. I'll take a closer look and give feedback.

Haeresis:
Absolutely! And I prefer just reorder it like I initially suggest! Cool! I
will change this PR!

Haeresis:
Ok @chrisvfritz, it done.

Just need your through about: https://github.com/vuejs/vuejs.org/issues/1245#issuecomment-342426331

PR #1276, opened 2017-11-12
improved #1269 with lorem ipsum

Jinjiang:
ref: #1269
I changed the content instead of lorem ipsum. So the text content will have full width of the box to be easy and simple enough to understand what `<transition name="bounce">` will do.
demo: https://jsfiddle.net/qcar0cqq/
cc @StarkL
Thanks.


StarkL:
Great!  Thanks!

PR #1277, opened 2017-11-12
docs: add mustaches vs v-html example

phanan:
When translating this part of the guide into Vietnamese I felt an example would explain the differences between double-mustache syntax and `v-html` directive much more clearly. If the PR is merged, we'll have something like this:

image

Does this make sense?


phanan:
Btw I notice we have _a lot_ of examples at the beginning and fewer and fewer of them (feels like we're "getting tired") deeper into the guide.

Issue #814, opened 2017-03-09
nextTick: Add note about long operations blocking UI refresh ins some browsers.


LinusBorg:
As discussed [here](https://github.com/vuejs/vue/issues/5136#issuecomment-285288733), there's a caveat with $nextTick (or rather, native Promises as a whole, which $nextTick relies on):

The callback is pushed into the browsers microtask queue rather then the event loops normal queue. In some browsers (i.e. Chrome), than means that UI updates (due to DOM Changes) will be blocked until the microtask queue is done. 

That means that heavy operations in a `$nextTick` callback can still block the UI - while an "old-school" `setTimeout(cb, 0)` will have UI updates run before the `cb` is picked up by the event loop.

So in short:  

* use $nextTick when you have to do some lightweight (or really async, like AJAX) operation after the DOM has been updated.
* don't use `$nextTick` or rather, combine it with a `setTimeout(cb, 0)`, when you have heavy, sync operations to do.

chrisvfritz:
Closing, since this is no longer an issue in 2.4+.

PR #1279, opened 2017-11-14
Add a note about nextTick and render blocking to Reactivity in Depth

sdras:
I'm not totally sold on the formatting for this one, but this takes the text from #814, and adds a note to the $nextTick section of Reactivity in Depth. I edited a bit for clarity, but I'm welcome to feedback.

### DIFF
--- a/src/v2/guide/reactivity.md
+++ b/src/v2/guide/reactivity.md
@@ -118,3 +118,9 @@ Vue.component('example', {
   }
 })
 ```
+
+<p class="tip">`$nextTick` relies on native Promises, which means that heavy operations in a `$nextTick` callback can be render blocking. This is because the callback is pushed into the browsers microtask queue rather then the event loops normal queue. 
+
+For this reason, when you have heavy synchronized tasks, it’s it’s wise to avoid using `$nextTick`, or rather, combine it with a `setTimeout()`. A `setTimeout()` will have UI updates run before the callback is picked up by the event loop.
+
+`$nextTick` is better suited for when you have to do some lightweight or async operation _after_ the DOM has been updated.</p>


src/v2/guide/reactivity.md
```diff
@@ -118,3 +118,7 @@ Vue.component('example', {
   }
 })
 ```
+
+<p class="tip">`$nextTick` relies on native Promises, so heavy operations in a `$nextTick` callback can be render blocking. This is because the callback is pushed into the browsers microtask queue rather then the event loops normal queue. For this reason, when you have heavy synchronized tasks it’s wise to avoid using `$nextTick`, or rather, combine it with a `setTimeout()`. A `setTimeout()` will run UI updates before the callback is picked up by the event loop.
```
chrisvfritz: > `$nextTick` relies on native Promises
> the callback is pushed into the browsers microtask queue

This isn't _quite_ accurate. In 2.4+, `$nextTick` now tries to avoid the microtask queue, currently trying to use:

- `setImmediate` if available, then
- `MessageChannel` if available, then
- `Promise.resolve` (which uses microtasks) if available, then
- `setTimeout`

From the source:

``` js
// An asynchronous deferring mechanism.
// In pre 2.4, we used to use microtasks (Promise/MutationObserver)
// but microtasks actually has too high a priority and fires in between
// supposedly sequential events (e.g. #4521, #6690) or even between
// bubbling of the same event (#6566). Technically setImmediate should be
// the ideal choice, but it's not available everywhere; and the only polyfill
// that consistently queues the callback after all DOM events triggered in the
// same loop is by using MessageChannel.
/* istanbul ignore if */
if (typeof setImmediate !== 'undefined' && isNative(setImmediate)) {
  timerFunc = function () {
    setImmediate(nextTickHandler);
  };
} else if (typeof MessageChannel !== 'undefined' && (
  isNative(MessageChannel) ||
  // PhantomJS
  MessageChannel.toString() === '[object MessageChannelConstructor]'
)) {
  var channel = new MessageChannel();
  var port = channel.port2;
  channel.port1.onmessage = nextTickHandler;
  timerFunc = function () {
    port.postMessage(1);
  };
} else
/* istanbul ignore next */
if (typeof Promise !== 'undefined' && isNative(Promise)) {
  // use microtask in non-DOM environments, e.g. Weex
  var p = Promise.resolve();
  timerFunc = function () {
    p.then(nextTickHandler);
  };
} else {
  // fallback to setTimeout
  timerFunc = function () {
    setTimeout(nextTickHandler, 0);
  };
}
```

sdras:
Oh interesting! @LinusBorg what do you think? Can you please provide a repro of what you're describing in the issue, and if this is no longer a problem, we can close out the issue and the PR?

Thank you!

LinusBorg:
@sarah I would require some sync process that takes about a second so it's noticable.  I have nothing ready but can see what I can come up with.

The undesirable behaviour I described would manifest like this:
```javascript
methods: {
  heaveOperation() {
    this.processing = true // flag to show a "Processing..." animation / loader
    this.$nextTick(() => {
      this.data = parsingAHugeCSVFileOrSomething(file)
      this.processing = false
    })
  }
}
```
You would expect that animation to show while the heave operation is running, but that was actually not the case, because `nextTick` squeezed its execution in between the change of the `processing` change and the async DOM update resulting from it.

chrisvfritz:
@LinusBorg So would you say that this kind of warning isn't necessary anymore?

sarah:
wrong sarah.

LinusBorg:
Ooops, sorry. ^^

@chrisvfritz Well I haven't been able to verify it in a real app, but technically there shouldn't be a possibility for it to happen now, so we should be fine to drop this.

sdras:
This is great, I'm happy Vue handles this more gracefully now. Cheers.

Issue #1273, opened 2017-11-11
Styleguide Priority D Styles are a little b0rked


sdras:
Small issue with one of the smaller headers in the styleguide, I'll fix sometime this week:

<img width="462" alt="screen shot 2017-11-11 at 7 48 42 am" src="https://user-images.githubusercontent.com/2281088/32690498-f9c790c6-c6b4-11e7-8817-1affe84abdf8.png">

printercu:
I'm not sure if this same issue with: 
![image](https://user-images.githubusercontent.com/1144890/32697276-0b4c27be-c79e-11e7-83f3-46e59c153da0.png)

May be it's related to https://github.com/vuejs/vuejs.org/pull/1263

chrisvfritz:
Just fixed this by reverting #1263.

sdras:
Thanks Chris!

PR #1280, opened 2017-11-15
Fix header link

Justineo:
Should convert `NodeList` to `Array` to prevent #1273.


Justineo:
See it in action at [CN site](https://cn.vuejs.org/v2/style-guide/#没有在-v-if-v-if-else-v-else-中使用-key-谨慎使用). It should work well now.

/ping @chrisvfritz

PR #1281, opened 2017-11-15
`v-if` on same element as `v-for`

Haeresis:
Following the discussion form this issue with @chrisvfritz:
— https://github.com/vuejs/vuejs.org/issues/1245#issuecomment-341794344

I suggest this new style guide rules:

Please help me to do something more clear ;)


Haeresis:
Hi @chrisvfritz, do you need some clarification about this?

chrisvfritz:
@Haeresis No, I just need a good chunk of time and energy to review it properly. 😅  This week is looking good.

Haeresis:
I agree. Time... Time...

chrisvfritz:
I just made some time to look this over and also moved it to Priority A, since thinking about it more, the possibility of extremely poor performance really is more in the spirit of Error Prevention. Just like _Keyed `v-for`_ above, it likely won't result in any actual "errors", but it could definitely result in behavior that could be considered a bug.

Let me know what you think. 🙂

src/v2/style-guide/index.md
```diff
@@ -305,6 +305,164 @@ In our experience, it's better to _always_ add a unique key, so that you and you
...
+  >
+    {{ user.name }}
+  <li>
+</ul>
+```
+
+We get the following benefits:
+
+- The filtered list will _only_ be re-evaluated if there are relevant changes to the `users` array, making filtering much more efficient.
+- Using `v-for="user in activeUsers"`, we _only_ iterate over active users during render, making rendering much more efficient.
```
phanan: One more benefit I can think of: Logic is now decoupled from the presentation layer, making maintenance (change/extension of logic) much easier.

phanan: E.g. if later we decide to consider a user active only if `isActive` is `true` AND `lastLoggedIn` is greater than 1 year ago.

Haeresis:
👏 👏 I understand better why @chrisvfritz need some time to review this PR. 👏 👏 

Really great job, is more clearer in this way. I agree with priority A. Thx for your review @posva and @phanan

src/v2/style-guide/index.md
```diff
@@ -305,6 +305,164 @@ In our experience, it's better to _always_ add a unique key, so that you and you
...
+
+We get similar benefits from updating:
+
+``` html
+<ul>
+  <li
+    v-if="shouldShowUsers"
+    v-for="user in users"
```
znck: If it's not written intentionally as above, `v-for` can be placed before `v-if` to avoid confusion.

chrisvfritz:
Thanks everyone for all your contributions to this! I'm really happy with it. 🙂

src/v2/style-guide/index.md
```diff
@@ -305,6 +305,165 @@ In our experience, it's better to _always_ add a unique key, so that you and you
...
 
+### Avoid `v-if` with `v-for` <sup data-p="a">essential</sup>
+
+**Never use `v-if` on the same element as `v-for`.**
+
+There are two common cases where this can be tempting:
+
+- To filter items in a list (e.g. `v-for="user in users" v-if="user.isActive"`). In these cases, replace `users` with a new computed property that returns your filtered list (e.g. `activeUsers`).
+
+- To avoid rendering a list if it should be hidden (e.g. `v-for="user in users" v-if="shouldShowUsers"). In these cases, move the `v-if` to a container element (e.g. `ul`, `ol`).
```
Haeresis: There is a type error here just after @chrisvfritz 

```
`v-for="user in users" v-if="shouldShowUsers" <== need a tick `
```

PR #1285, opened 2017-11-18
Make id example with Html

723gt:
If you have not specified to write Html before JavaScript
Since the result is false, you should specify the ID in html

``` js
VM. $ el === document.getElementById ('example') // => false
```


chrisvfritz:
Hmm, I thought about it and started removing the `el` option altogether to avoid complicating the example, but the more I think about it, I actually think it's OK not to display the implied element with `id="example"`. I'm open to revisiting this in the future though, if it comes up again.

PR #1287, opened 2017-11-18
docs: add a filter example

phanan:
Currently, we only have the code of the example, not the demo example itself. This PR would add something like this into the page:

image

which helps the page to be a bit more visual. What do you think?

### DIFF
--- a/src/v2/guide/filters.md
+++ b/src/v2/guide/filters.md
@@ -36,6 +36,32 @@ Vue.filter('capitalize', function (value) {
 })
 ```
 
+Below is an example of our `capitalize` filter being used:
+
+{% raw %}
+<div id="example_1" class="demo">
+  <input type="text" v-model="message">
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Sure. 🙂

src/v2/guide/filters.md
```diff
@@ -36,6 +36,32 @@ Vue.filter('capitalize', function (value) {
 })
 ```
 
+Below is an example of our `capitalize` filter being used:
+
+{% raw %}
+<div id="example_1" class="demo">
```
Jinjiang: Sorry for a late question (just found that when do translation), but is it better to use hyphenated id (`example-1`) as other doc pages used?
Thanks :-)

chrisvfritz: @Jinjiang Yeah, good catch!

PR #1289, opened 2017-11-21
Clarify checkbox value bindings

jordanlev:
Add a "tip" box to the section on checkbox value bindings that explains how true-value and false-value only apply to the Vue instance and are *not* submitted with form data to the server.

### DIFF
--- a/src/v2/guide/forms.md
+++ b/src/v2/guide/forms.md
@@ -309,18 +309,19 @@ But sometimes we may want to bind the value to a dynamic property on the Vue ins
 <input
   type="checkbox"
   v-model="toggle"
-  v-bind:true-value="a"
-  v-bind:false-value="b"
+  true-value="yes"
+  false-value="no"
 >
 ```
 
 ``` js
 // when checked:
-vm.toggle === vm.a
+vm.toggle === 'yes'
 // when unchecked:
-vm.toggle === vm.b
+vm.toggle === 'no'
 ```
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks @jordanlev! I've updated the wording slightly and also pinged @sdras for review, to double check the clarity of the new explanation and examples. Sarah, please merge if it looks good to you. 🙂

jordanlev:
@chrisvfritz I like your clarification in the wording of the "tip" (although there is another way to send both true and false values without having to use radio buttons -- by having a hidden input before the checkbox input with the same name and the false value... but perhaps that's too complicated to get into in this doc).

The code you changed, though -- I wonder if it's making things more confusing, because above that code it's talking about binding the values, but you've changed it so `true-value` and `false-value` are *not* binding to anything but instead have hard-coded strings.

chrisvfritz:
>The code you changed, though -- I wonder if it's making things more confusing, because above that code it's talking about binding the values, but you've changed it so true-value and false-value are not binding to anything but instead have hard-coded strings.

@jordanlev The "binding of values" is still happening with `v-model`. The `true-value` and `false-value` do not affect the binding of the value, but how rather how Vue stores the value. These attributes will also very rarely be dynamic in practice, so I think it makes sense for them not to be hard-coded strings.

chrisvfritz:
> I like your clarification in the wording of the "tip" (although there is another way to send both true and false values without having to use radio buttons -- by having a hidden input before the checkbox input with the same name and the false value

You _can_ do that, but I wouldn't recommend it. 😅  I'd argue that when you want to guarantee one of two different values submitted to a form, a checkbox is semantically the wrong element. The reason is that the presence of a value implies user input. An unchecked box doesn't necessarily mean the user saw a field and chose not to leave it unchecked - they could have just missed it. Radio inputs _do_ guarantee user input.

chrisvfritz:
I'm going to go ahead and merge this. I _think_ it's OK and people will let me know if it's not. 😄

PR #1292, opened 2017-11-23
changes for v-enter-to&&v-leave-to

luanyujia:
Changed the explanation of the “v-enter-to” and “v-leave-to”


sdras:
Thanks for the PR! I'm not super into this change because it gives less information about what's really going on in the logic of how the component is added and how that interplay is connected to the animation states. I'm going to close this out, but I'm open to more discussion around why you think the information given doesn't suit the purpose of this explanation, because perhaps there's another addition we could make that could help with that.

PR #1294, opened 2017-11-24
change var to const when approriate in Guide

p-adams:
@chrisvfritz Is this worth the effort? If so, I would like to continue updating instances of var to const or let as approriate.


chrisvfritz:
Thanks for the contribution, but we're actually keeping it `var` intentionally so that we don't assume any knowledge of ES6+ in our core docs. This makes it a bit more friendly to beginners. 🙂

PR #1295, opened 2017-11-24
components are also Vue instances

indexlongrun:
I think this is important info, especially for someone who reaches this page directly (e.g. from search) and has missed the comment in the Vue instance page.


chrisvfritz:
I like it, thanks! 😄

PR #1301, opened 2017-12-01
Change Ctor to Constructor in unit testing portion of guide

p-adams:


chrisvfritz:
I'm happy to merge this if you revert const back to var (I'm guessing it was included in this PR by mistake). 🙂

p-adams:
const is actually being used here in the docs https://vuejs.org/v2/guide/unit-testing.html

chrisvfritz:
Yes, the `const`s in the unit testing part of the guide were actually good, since unit tests will typically be run in a Node environment. But you've changed `var`s to `const`s in the Essentials of the guide (`src/v2/guide/index.md`) and that's the part I'd like to change back..

p-adams:
Yes, that was a mistake. Thanks for pointing it out!

chrisvfritz:
Looks great! 🙂 🎉

PR #1302, opened 2017-12-01
Update syntax.md

shaiker:
Added missing "mustache" operator call in v-html example


shaiker:
My bad... Wrong understanding

PR #1303, opened 2017-12-01
Clarify valid use of exact with no system modifiers

mokkabonna:
The wording currently seem to only allow exact in combination with system modifiers.


chrisvfritz:
Thanks! I updated the wording a little to keep the description brief.

PR #1306, opened 2017-12-04
Update forms.md 

LongKissGoodBye:
I think there should be  '='  rather than '==='


chrisvfritz:
As Alex mentioned, this is correct. We're just pointing out that these two things are equal.

PR #1309, opened 2017-12-05
Scoped slots: slot-scope must be on a descendant

adam-lynch:


chrisvfritz:
I might misunderstanding what you're trying to say, but `slot-scope` can also be used on the root slot element/component (as seen in the example directly below that sentence), so it's not true that it must be on a child.

adam-lynch:
@chrisvfritz I expected to be able to do this:

```javascript
<list-view :defaultCells="defaultCells" :fetchParams="fetchParams" mainEntityType="activities" slot-scope="props">
    <p>Activities ({{props.data.length}}/?)</p>
</list-view>
```

In an earlier sentence in the doc, it says:

> In a **child** component, pass data into a slot as if you are passing props to a component:

I must've missed this. Adding "child" where I suggested would be clearer in my opinion but I'd understand if you were to say it should be obvious at that point.

chrisvfritz:
Ah, I see! Thanks for explaining. I can definitely where the misunderstanding came from now, since `slot-scope` _can_ be used on the root element of named slots (e.g. `slot="mySlotName" slot-scope="props"`) and since default slots don't have a root element, it could be easy to assume that maybe the component itself _is_ the root element of the default slot.

I could see an argument for `slot-scope` being able to work like this, so you're welcome to open an issue on [the core library](https://github.com/vuejs/vue) to see what other team members think. 

In the meantime, I've opened #1312 as an attempt to make this whole section a little clearer and I'd love your feedback on it. 🙂  I'll close this, as #1312 is a sort of spiritual successor to this one.

adam-lynch:
OK thanks. I'll have a look at your PR & potentially a PR to the core (eek maybe) tomorrow / sunday.

PR #1310, opened 2017-12-06
Add Props and Events Diagram

MatheusRich:
I was helping with the [pt-BR translation of Vue.js](https://github.com/vuejs-br/br.vuejs.org) when I noticed that there's no `.ai` file of [this diagram](https://vuejs.org/v2/guide/components.html#Composing-Components) (image bellow) in the `assets` directory. 

So, I tried to remake (_as close as I could_) the diagram. Here's what I've got. I hope this helps others translations around the world. 

**Original**

</img>

**My Version**
</img>


sdras:
That font is wrong but I'm probably the only one who will notice :)

@chrisvfritz can you tell?

MatheusRich:
@sdras I've noticed that too, but I don't know which one is...

sdras:
Looks like Source Sans Pro to me but I could be wrong.

chrisvfritz:
@MatheusRich I love it! 😍  I didn't notice the font (at least when out of context), but if it stood out to you @sdras, I think it'll stand out to a lot of other other people with similar a typography spideysense. If it doesn't turn out to be Source Sans Pro, I can find out what it is.

MatheusRich:
I got the font from another diagram (`state.ai` i think), so maybe in some other we can find the right one.

chrisvfritz:
@MatheusRich Let me know if you can't find it by digging around in the other `.ai` files - if it's not in there, I'll launch a formal investigation. 😄

yyx990803:
I actually like the new font - it looks fairly close to another font used across the diagrams ([Montserrat Bold](https://fonts.google.com/specimen/Montserrat)).

PR #1311, opened 2017-12-07
Add Bugsnag integration to errorHandler docs

bengourley:
Hi there 👋

I work on the JavaScript platforms at @bugsnag. We just released a complete rewrite of our [JS notifier](https://github.com/bugsnag/bugsnag-js), and along with it added first-class Vue support via our official [bugsnag-vue plugin](https://github.com/bugsnag/bugsnag-vue).

It would be great for our support to be listed here where you already mention Sentry's, and potentially beneficial for your users who may already be Bugsnag users that don't know about the new integration yet.

I reworked the wording of the section in the way I thought best, but happy for it to go a different way if desired.

Let me know what you think! Cheers

_**Before**_

image

_**After**_

image

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -83,7 +83,7 @@ type: api
 
   > In 2.4.0+ this hook also captures errors thrown inside Vue custom event handlers.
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Looks great! 🙂

PR #1312, opened 2017-12-08
Improve explanation of scoped slots

chrisvfritz:
I've noticed a lot of people get very confused about scoped slots, having to re-read the section many times before starting to feel comfortable. There have also been some misunderstandings like the one in #1309. I think scoped slots are inherently one of the more difficult concepts to grasp, but this is an attempt to overhaul the section to make it a little clearer.

@Haeresis @adam-lynch GitHub won't let me request a review from you, but I'm particularly interested in your feedback. 🙂

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -963,67 +963,56 @@ The content distribution API is a very useful mechanism when designing component
 
 > New in 2.1.0+
 
-A scoped slot is a special type of slot that functions as a reusable template (that can be passed data to) instead of already-rendered-elements.
-
-In a child component, pass data into a slot as if you are passing props to a component:
-
-``` html
-<div class="child">
-  <slot text="hello from child"></slot>
-</div>
-```
-
-In the parent, a `<template>` element with a special attribute `slot-scope` must exist, indicating that it is a template for a scoped slot. The value of `slot-scope` will be used as the name of a temporary variable that holds the props object passed from the child:
-
-``` html
-<div class="parent">
-  <child>
-    <template slot-scope="props">
-      <span>hello from parent</span>
-      <span>{{ props.text }}</span>
-    </template>
-  </child>
-</div>
-```
-
-If we render the above, the output will be:
-
-``` html
-<div class="parent">
-  <div class="child">
-    <span>hello from parent</span>
-    <span>hello from child</span>
-  </div>
-</div>
-```
-
-> In 2.5.0+, `slot-scope` is no longer limited to `<template>` and can be used on any element or component.
-
-A more typical use case for scoped slots would be a list component that allows the component consumer to customize how each item in the list should be rendered:
-
-``` html
-<my-awesome-list :items="items">
-  <!-- scoped slot can be named too -->
-  <li
-    slot="item"
-    slot-scope="props"
-    class="my-fancy-item">
-    {{ props.text }}
-  </li>
-</my-awesome-list>
-```
-
-And the template for the list component:
+Scoped slots allow the definition of dynamic slot templates, using data passed from a child component. For example, in this hypothetical `<review-list>` component:
 
 ``` html
 <ul>
-  <slot name="item"
-    v-for="item in items"
-    :text="item.text">
-    <!-- fallback content here -->
-  </slot>
+  <li v-for="review in reviews">
+    <slot :review="review"></slot>
+  </li>
 </ul>
 ```
+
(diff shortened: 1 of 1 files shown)


src/v2/guide/components.md
```diff
@@ -963,68 +963,57 @@ The content distribution API is a very useful mechanism when designing component
...
+    <h3>So many choices!</h3>
+    <p class="byline">
+      Samwise Gamgee
+    </p>
+    <p>Normally, only one or two things look good to me, but here, everything is delicious!</p>
+  </li>
 </ul>
 ```
 
+> Before 2.5.0, the `slot-scope` attribute was limited to `<template>` elements, but can now be used with any element or component, on or inside a slot.
```
Haeresis: > **on** or inside a slot.

I don't fully understand this. That means the following code work?

```html
<review-list
  :reviews="restaurantReviews"
  slot-scope="slotProps">
  <h3>{{ slotProps.review.title }}</h3>
  <p class="byline">
    {{ slotProps.review.author }}
  </p>
  <p>{{ slotProps.review.content }}</p>
</review-list>
```

chrisvfritz: No, because `<review-list>` is not on or inside a slot (it's not inside itself). Instead, the default slot starts _inside_ the component. This is one of the areas of confusion I was trying to make clearer and have apparently failed. 😅  What do you think could make it clearer?

src/v2/guide/components.md
```diff
@@ -963,68 +963,57 @@ The content distribution API is a very useful mechanism when designing component
...
-    <template slot-scope="props">
-      <span>hello from parent</span>
-      <span>{{ props.text }}</span>
-    </template>
-  </child>
-</div>
-```
-
-If we render the above, the output will be:
+Scoped slots allow the definition of dynamic slot templates, using data passed from a child component. For example, in this hypothetical `<review-list>` component:
```
adam-lynch: I prefer the old text but the new code example

src/v2/guide/components.md
```diff
@@ -963,68 +963,57 @@ The content distribution API is a very useful mechanism when designing component
...
+    <slot :review="review"></slot>
+  </li>
+</ul>
 ```
 
-> In 2.5.0+, `slot-scope` is no longer limited to `<template>` and can be used on any element or component.
+We're passing `review` to the slot, just like props to a component.
 
-A more typical use case for scoped slots would be a list component that allows the component consumer to customize how each item in the list should be rendered:
+Now when we use this component, we can use a `slot-scope` attribute to access these props, binding them to an object called `slotProps` in the example below:
```
adam-lynch: Maybe name the component instead of `this component`?

src/v2/guide/components.md
adam-lynch: If the doc contained this, I still would have misunderstood. Like I said though, I missed the mention of "child elements" earlier in the doc so maybe that's just my own fault. "on or inside a slot" doesn't clarify it further from my point of view. If there's only one slot, I would've thought the wrapping tags were an acceptable place (i.e. on the slot).

Maybe making it "child element or component" would be best? You could go further and a HTML comment to the example?

znck: [L1015](https://github.com/vuejs/vuejs.org/pull/1312/files?diff=split#diff-044f0befe8adfe59792a31b1763e3cc4R1015) is confusing. Maybe providing an inline example would help. 

@chrisvfritz this PR should be run by someone new to vue.

src/v2/guide/components.md
yyx990803: We could also use a bit more explanation on the problem:

> Imagine we have a `<review-list>` component that renders a list of reviews. It would be easy if the component simply renders the same markup for each review all the time. However, sometimes we may want to allow other developers using the component to customize the template for each review item.

> We know that we can use slots to inject content from parent to the child, but in this case, each review item relies on data that is only available in the child component, and normal slots only have access to parent scope data.

> This is where scoped slots come into play: scoped slots allow you to inject a block of reusable template into the child component that can receive data provided by the child.

sdras:
Hey @chrisvfritz this PR points to the components page which has now gone through all of that restructuring. Do you want to fix it up or should we close this one out and start fresh? Thanks!

chrisvfritz:
I'll close it out. It's also less necessary now, since we made some improvements to the explanation in the big components refactor.

Haeresis:
Yeah! Great refactoring!

Issue #1298, opened 2017-11-29
Add `<template functional>` syntax to section about functional components


lmiller1990:
Functional components are discussed [here](https://vuejs.org/v2/guide/render-function.html#Functional-Components) and according to [this](https://medium.com/the-vue-point/vue-2-5-released-14bd65bf030burl) announcement we can now do

`<template functional>` instead of `functional: true`.

I think the docs should include this. I can make this PR later this week if the team is happy with this idea.

chrisvfritz:
A small note regarding SFCs on that page linking to [the vue-loader docs](https://vue-loader.vuejs.org/en/features/functional.html) would be welcome. 🙂

lmiller1990:
Thanks Chris, I will do it in the next day or two!

lmiller1990:
I looked into this, it feels really out of place on [this](https://vuejs.org/v2/guide/single-file-components.html) page. I feel like it fits better in [this section[(https://vuejs.org/v2/guide/render-function.html#Functional-Components) that talks about how to make a file `functional`. 

Is there any reason you felt it fits better in the SFC area @chrisvfritz ? The reason I think it goes best in the section section is because when I was making some functional components, I was referring to that area mostly, as opposed to the SFC page or vue-loader documentation, which I rarely use.

chrisvfritz:
Just a misunderstanding. 🙂  we're on the same page actually (no pun intended). When I said "that page" I was referring to the one you linked to in your original post, which is the render functions page.

lmiller1990:
Sure! Great to hear. I'll make the PR this weekend.

Thanks for your quick response, no wonder Vue has the best documentation of any framework I've used.

lmiller1990:
It's done, closing

PR #1313, opened 2017-12-08
Add new functional syntax

lmiller1990:
#1298 

Let me know if you think this should be in a different place or worded differently.

Thanks!


chrisvfritz:
Made a couple tiny tweaks, but looks good. Thanks! 😄

PR #1315, opened 2017-12-11
multi-attribute elements with directive

Buzut:
Add a multi-attribute element exemple with directives grouped at the top. I find it to be the most readable solution.


chrisvfritz:
The example isn't really related to the number of attributes per line, unfortunately. Also, it violates our [recommended attribute order](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended). Attribute order is only a Priority C rule though, so I think you're fine doing what you're doing if you find it more readable. 🙂

PR #1320, opened 2017-12-13
fix a typo in code example

MinimalistYing:
The previous code example is not correct(lost brace).It seems like the code even can't complie by vue-loader in `.vue` component.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -618,7 +618,7 @@ The following
 is expanded into:
 
 ``` html
-<comp :foo="bar" @update:foo="val => bar = val"></comp>
+<comp :foo="bar" @update:foo="val => { bar = val }"></comp>
 ```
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
This is actually correct code that I've just confirmed is compiled by vue-loader without error. Can you provide more information about the error you're seeing?

MinimalistYing:
Sorry~I tried again, the code is indeed correct, and it can be compiled successfully.
But there is still a little problem without brace.
I am using Sublime with Vue Syntax Highlight(3.2.3 the latest version).
This code without brace will cause the Syntax Highlight can't highlight the code after this correctly.

vue-highlight

Thanks for your reply~ If you think this problem doesn't matter,I will close this PR.

chrisvfritz:
I do think it matters - it's just a bug with a different repo, rather than a documentation issue. It looks like there's already [an open issue for the problem here](https://github.com/vuejs/vue-syntax-highlight/issues/101).

Issue #1314, opened 2017-12-11
Form validation demo


cfjedimaster:
I'd like to create a Cookbook entry that demonstrates form validation with Vue. This would be based heavily on the blog post I did here: https://www.raymondcamden.com/2017/12/07/an-example-of-form-validation-with-vuejs/

sdras:
Hi Raymond! This seems like a really great entry for the cookbook. The only thing I might add would be a mention of vuelidate as an alternative pattern, as it's usage is fairly ubiquitous in Vue. What do you think @chrisvfritz?

chrisvfritz:
@sdras @cfjedimaster I like the topic! vee-validate is another popular library I think we should mention as an alternative, but I agree focusing on building form validations from scratch is the right focus. It'll probably be the most educational and also easier to keep up-to-date. 😄

cfjedimaster:
What is the next step? I assume my blog entry can't be used as is - that first paragraph would need to change. But how should I get a draft to yall?

sdras:
As I mentioned previously on twitter, please use [this template](https://vuejs.org/v2/cookbook/cookbook-contributions.html) as a guide. you can look at "adding instance properties" as an example as well. You'd submit a PR here and we'd review it. You'd probably have to rewrite the post to fit our format, but there's a lot of content there that's reusable with some eye to formatting. Thanks

cfjedimaster:
Ah sorry - forgot that earlier tweet. Doing too much at once. :) Thank you for the reminder.

sdras:
No worries! Thanks

cfjedimaster:
Are codepen's ok? I love that users can see the code and run it directly on the page, but having it tied to my account may not be safe.

sdras:
I like codepens. We've also used JSFiddle and CodeSandbox, it's probably no less safe than those- I know codepen does a lot of security work.

cfjedimaster:
I guess my issue is - are you (the official Vue you ;) ok with "my" codepens being on the site? I don't mind for sure, but wanted to make sure that was ok.

sdras:
I'm not sure why you think the answer is any different? We're using CodePens for the cookbook.

cfjedimaster:
Sorry - I guess I was just worried about there being a "policy" against using someone's pen (even the author of the entry) on vuejs.org. I'm over thinking it - I'll just go for it.

sdras:
[...]

PR #1324, opened 2017-12-15
cookbook entry on form validation

cfjedimaster:
Addresses issue #1314 - a cookbook version of my blog entry. I modified the text to switch from "I" to "we" and made a few other minor edits.

### DIFF
--- a/src/v2/cookbook/form-validation.md
+++ b/src/v2/cookbook/form-validation.md
@@ -1,0 +1,322 @@
+---
+title: Form Validation
+type: cookbook
+order: 1.2
+---
+
+## Simple Example
+
+Form validation is natively supported by the browser, but sometimes different browsers will handle things in a manner which makes relying on it a bit tricky. Even when validation is supported perfectly, there may be times when custom validations are needed and a more manual, Vue-based solution may be more appropriate. Let's begin with a simple example.
+
+For my first example, let's create something as simple as possible. Given a form of three fields, make two required. Let's look at the HTML first:
+
+``` html
+<form id="app" @submit="checkForm" action="/something" method="post">
+  
+  <p v-if="errors.length">
+    <b>Please correct the following error(s):</b>
+    <ul>
+      <li v-for="error in errors">{{ error }}</li>
+    </ul>
+  </p>
+  
+  <p>
+  <label for="name">Name<label>
+  <input type="text" name="name" id="name" v-model="name">
+  </p>
+
+  <p>
+  <label for="age">Age<label>
+  <input type="number" name="age" id="age" v-model="age">
+  </p>
+
+  <p>
+  <label for="movie">Favorite Movie<label>
+  <select name="movie" id="movie" v-model="movie">
+    <option>Star Wars</option>
+    <option>Vanilla Sky</option>
+    <option>Atomic Blonde</option>
+  </select>
+  </p>
+
+  <p>
+  <input type="submit" value="Submit">  
+  </p>
+
+</form>
+```
+
+Let's cover it from the top. The form tag has an ID that we'll be using for the Vue component. There's a submit handler that you'll see in a bit, and the action is a fake URL that would point to something real on a server someplace (where you have backup server-side validation of course).
+
+Beneath that there isa paragraph that shows or hides itself based on an error state. This is a personal preference of mine when building forms. I like a nice simple list of errors on top of the form. You may like error handling by the fields themselves. Use what works. Also note I'm going to fire my validation on submit rather than as every field is modified. Again, this is a personal preference.
+
+The final thing to note is that each of the three fields has a corresponding v-model to connect them to values we will work with in the JavaScript. Now let's look at that.
+
+``` js
+const app = new Vue({
+  el:'#app',
+  data:{
+    errors:[],
+    name:null,
+    age:null,
+    movie:null
+  },
+  methods:{
+    checkForm:function(e) {
+      if(this.name && this.age) return true;
+      this.errors = [];
+      if(!this.name) this.errors.push("Name required.");
+      if(!this.age) this.errors.push("Age required.");
+      e.preventDefault();
+    }
+  }
+})
+```
+
+Fairly short and simple. We default an array to hold errors and set null values for the three form fields. The checkForm logic (which is run on submit remember) checks for name and age only as movie is optional. If they are empty we check each and set a specific error for each. And that's really it. You can run the demo below. Don't forget that on a successful submission it's going to POST to a non-existent URL.
+
+<p data-height="265" data-theme-id="0" data-slug-hash="GObpZM" data-default-tab="html,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="form validation 1" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/GObpZM/">form validation 1</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://production-assets.codepen.io/assets/embed/ei.js"></script>
+
+## Using Custom Validation
+
+For the second example, the second text field (age) was switched to email which will be validated with a bit of custom logic. The code is taken from the StackOverflow question, [How to validate email address in JavaScript?](https://stackoverflow.com/questions/46155/how-to-validate-email-address-in-javascript). This is an awesome question because it makes your most intense Facebook political/religious argument look like a slight disagreement over who makes the best beer. Seriously - it's insane. Here is the HTML, even though it's really close to the first example.
+
+``` html
+<form id="app" @submit="checkForm" action="/something" method="post" novalidate="true">
+  
+  <p v-if="errors.length">
+    <b>Please correct the following error(s):</b>
+    <ul>
+      <li v-for="error in errors">{{ error }}</li>
+    </ul>
+  </p>
+  
+  <p>
+  <label for="name">Name<label>
+  <input type="text" name="name" id="name" v-model="name">
+  </p>
+
+  <p>
+  <label for="email">Email<label>
+  <input type="email" name="email" id="email" v-model="email">
+  </p>
+
+  <p>
+  <label for="movie">Favorite Movie<label>
+  <select name="movie" id="movie" v-model="movie">
+    <option>Star Wars</option>
+    <option>Vanilla Sky</option>
+    <option>Atomic Blonde</option>
+  </select>
+  </p>
+
+  <p>
+  <input type="submit" value="Submit">  
+  </p>
+
+</form>
+```
+
+While the change here is small, note the `novalidate="true"` on top. This is important because the browser will attempt to validate the email address in the field when type="email". Frankly it may make more sense to trust the browser in this case, but as we wanted an example with custom validation, we're disabling it. Here's the updated JavaScript.
+
+``` js
+const app = new Vue({
+  el:'#app',
+  data:{
+    errors:[],
+    name:null,
+    email:null,
+    movie:null
+  },
+  methods:{
+    checkForm:function(e) {
+      this.errors = [];
+      if(!this.name) this.errors.push("Name required.");
+      if(!this.email) {
+        this.errors.push("Email required.");
+      } else if(!this.validEmail(this.email)) {
+        this.errors.push("Valid email required.");        
+      }
+      if(!this.errors.length) return true;
+      e.preventDefault();
+    },
+    validEmail:function(email) {
+      var re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    return re.test(email);
+    }
+  }
+})
+```
+
+As you can see, we've added `validEmail` as a new method and it is simply called from `checkForm`. Nothing too crazy, but a good example I think. You can play with this example here:
+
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+title: Form Validation
+type: cookbook
+order: 1.2
+---
+
+## Simple Example
+
+Form validation is natively supported by the browser, but sometimes different browsers will handle things in a manner which makes relying on it a bit tricky. Even when validation is supported perfectly, there may be times when custom validations are needed and a more manual, Vue-based solution may be more appropriate. Let's begin with a simple example.
+
+For my first example, let's create something as simple as possible. Given a form of three fields, make two required. Let's look at the HTML first:
```
sdras: This is a bit redundant as you said, "let's begin with a simple example" above. We try to use "we" instead of "my" in teaching as well. I'd suggest beginning the second paragraph with "Given a form..."

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+  
+  <p v-if="errors.length">
+    <b>Please correct the following error(s):</b>
+    <ul>
+      <li v-for="error in errors">{{ error }}</li>
+    </ul>
+  </p>
+  
+  <p>
+  <label for="name">Name<label>
```
sdras: Please indent the labels and inputs within the `p` tags, thanks!

sdras:
Thanks Raymond! I'll have some time this weekend to look it over and provide feedback.

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+  <p>
+  <input type="submit" value="Submit">  
+  </p>
+
+</form>
+```
+
+Let's cover it from the top. The form tag has an ID that we'll be using for the Vue component. There's a submit handler that you'll see in a bit, and the action is a fake URL that would point to something real on a server someplace (where you have backup server-side validation of course).
+
+Beneath that there isa paragraph that shows or hides itself based on an error state. This is a personal preference of mine when building forms. I like a nice simple list of errors on top of the form. You may like error handling by the fields themselves. Use what works. Also note I'm going to fire my validation on submit rather than as every field is modified. Again, this is a personal preference.
```
sdras: Separate "isa" to "is a". Please remove the references to personal preferences, this is a cookbook, so this is implied. Please switch the "I" statements to "we".

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+    },
+    validEmail:function(email) {
+      var re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+    return re.test(email);
+    }
+  }
+})
+```
+
+As you can see, we've added `validEmail` as a new method and it is simply called from `checkForm`. Nothing too crazy, but a good example I think. You can play with this example here:
```
sdras: You don't need the "nothing too crazy" sentence here

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+
+<p data-height="265" data-theme-id="0" data-slug-hash="vWqNXZ" data-default-tab="html,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="form validation 2" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/vWqNXZ/">form validation 2</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://production-assets.codepen.io/assets/embed/ei.js"></script>
+
+## Another Example of Custom Validation
+
+For the third example, we've built something you've probably seen in survey apps. The user is asked to spend a "budget" for a set of features for a new Star Destroyer model. The total must equal 100. First, the HTML.
```
sdras: This is well stated, thanks

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+  </p>
+
+  <p>
+  <input type="submit" value="Submit">  
+  </p>
+
+</form>
+```
+
+Note the set of inputs covering the five different features. Note the addition of .number to the v-model attibute. This tells Vue to cast the value to a number when you use it. However, there is a bug with this feature such that when the value is blank, it turns back into a string. You'll see the workaround below. To make it a bit easier for the user, we also added a current total right below so they can see, in real time, what their total is. Now let's look at the JavaScript.
```
sdras: I appreciate the attention to detail in this example. Really nice.

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+```
+
+We set up the total value as a computed value, and outside of that bug I ran into, it was simple enough to setup. My checkForm method now just needs to see if the total is 100 and that's it. You can play with this here:
+
+<p data-height="265" data-theme-id="0" data-slug-hash="vWqGoy" data-default-tab="html,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="form validation 3" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/vWqGoy/">form validation 3</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://production-assets.codepen.io/assets/embed/ei.js"></script>
+
+## Server-side Validation
+
+In my final example, we built something that makes use of Ajax to validate at the server. The form will ask you to name a new product and will then check to ensure that the name is unique. We wrote a quick OpenWhisk serverless action to do the validation. While it isn't terribly important, here is the logic:
```
sdras: Please add a link to OpenWhisk in case people aren't familiar with it

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+<form id="app" @submit="checkForm" method="post">
+  
+  <p v-if="errors.length">
+    <b>Please correct the following error(s):</b>
+    <ul>
+      <li v-for="error in errors">{{ error }}</li>
+    </ul>
+  </p>
+
+  <p>
```
sdras: Again, it's a small thing but please format the HTML in tandem with the DOM :) thanks

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+  }
+})
+```
+
+We start off with a variable representing the URL of the API that is running on OpenWhisk. Now look at `checkForm`. In this version, we always prevent the form from submitting (which, by the way, could be done in the HTML with Vue as well). You can see a basic check on this.name being empty, and then we hit the API. If it's bad, we add an error as before. If it's good, right now we do nothing (just an alert), but you could navigate the user to a new page with the product name in the URL, or do other actions as well. You can run this demo below:
+
+<p data-height="265" data-theme-id="0" data-slug-hash="BmgzeM" data-default-tab="js,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="form validation 4" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/BmgzeM/">form validation 4</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://production-assets.codepen.io/assets/embed/ei.js"></script>
+
+## Vue Form Validation Libraries
```
sdras: I'd appreciate a little more consideration here. Thanks for including it at our request, but we'd call this section "Alternative Patterns" and I'd love it if it addressed _why_ you might use a library in place of what you've written above and what the tradeoffs are (one is larger, but it might cover more use cases for different types of forms if you know you're building out more forms later. They will probably stay up to date with browser support based on community improvements etc)

Thanks!

cfjedimaster:
Thank you for all this! I've got a busy weekend so I may not get the fixes till Monday.

src/v2/cookbook/form-validation.md
cfjedimaster: FYI, after I make this (and the other edits), I'll be sure to update the codepen's too.

cfjedimaster:
Edits have been applied. Updating the Codepen's now. Thank you again for the review!

sdras:
Great! Give me a day or so and I'll take a look. Thanks!

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+  methods:{
+    checkForm:function(e) {
+      this.errors = [];
+      if(this.total != 100) this.errors.push("Total must be 100!");
+      if(!this.errors.length) return true;
+      e.preventDefault();
+    }
+  }
+})
+```
```
sdras: This seems like it needs to be kept from going below 0. The negative numbers don't make much sense if the user clicks down instead of up

cfjedimaster: One could see it as a strong vote against. :)

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+      this.errors = [];
+      if(!this.name) this.errors.push("Name required.");
+      if(!this.age) this.errors.push("Age required.");
+      e.preventDefault();
+    }
+  }
+})
+```
+
+Fairly short and simple. We default an array to hold errors and set null values for the three form fields. The checkForm logic (which is run on submit remember) checks for name and age only as movie is optional. If they are empty we check each and set a specific error for each. And that's really it. You can run the demo below. Don't forget that on a successful submission it's going to POST to a non-existent URL.
```
sdras: The weird part about this demo is that it posts to nothing, so playing with the demo will give the user a blank page. You kind of address this in the last sentence, but I would either add a success to the demo, or make the last sentence a warning, and say specifically that the POST URL will go to success page that they will have to build. (nitpicky, I know, but right now the demo feels a little funny.

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+            // redirect to a new url, or do something on success
+            alert('ok!');
+          }
+        });
+      }
+    }
+  }
+})
+```
+
```
sdras: this.name here should be in backticks for code :)

src/v2/cookbook/form-validation.md
sdras: This goes for the first example with the age (or maybe built into the second one) as well. Thanks!

src/v2/cookbook/form-validation.md
```diff
@@ -0,0 +1,322 @@
...
+<form id="app" @submit="checkForm" action="/something" method="post" novalidate="true">
+  
+  <p v-if="errors.length">
+    <b>Please correct the following error(s):</b>
+    <ul>
+      <li v-for="error in errors">{{ error }}</li>
+    </ul>
+  </p>
+
+  <p>
```
sdras: I quite like this example, but I think the text here could make it clearer that you have to select exactly 100. Maybe something like "Given a budget of 100 dollars, indicate how much you would spend on the following features for the next generation Star Destroyer by making the Current Total of features add up to 100:"

sdras:
This is looking really great! I started moving the examples over to CodePen. Once these things are fixed, I think it will be ready to get in :) Great work. I think it will be really helpful.

src/v2/cookbook/form-validation.md
sdras: @cfjedimaster good point! Maybe we leave this example as is but change the first one with the age. Pretty sure you can't be negative years old :)

cfjedimaster:
Thanks again for this - my rough ETA is Tuesday. Note - for cases where the codepen needs to be updated, I'll be sure to leave a note here so you can edit the official Vue copy.

sdras:
Awesome, thank you so much Raymond!

src/v2/cookbook/form-validation.md
cfjedimaster: I've added `min="0"` to the article (not yet in code), please add to the CodePen. Here is the complete line for reference:

    <input type="number" name="age" id="age" v-model="age" min="0">

src/v2/cookbook/form-validation.md
cfjedimaster: I'm a bit unsure how to handle this. I worry that if I post to a real URL, like Vue's home page, that would be confusing as well. Could I make the last sentence stronger? Like maybe just bold it? And perhaps add an exclamation mark? And then maybe we can see if reader's get confused by itself and send us feedback?

src/v2/cookbook/form-validation.md
cfjedimaster: I went with: " Your total must sum up to 100.".

cfjedimaster:
All changes pushed. Only open question is the warning about form submissions.

src/v2/cookbook/form-validation.md
sdras: It will only update within the iframe of the embed, so that would be ok. Can you please try it out and if that doesn't work we can go back and do bold or something.

sdras:
Looking great! Answered the form submission query inline above, that's the last piece and then I'll merge :)

src/v2/cookbook/form-validation.md
cfjedimaster: Do we want to change the embeds and text? If both, I'll update, if just the embeds, you can do that on your end, right?

cfjedimaster: Disregard previous comment.

cfjedimaster:
Ok, last thing changed.

sdras:
Congrats! Looks awesome!!! 🎉 Thanks for your contribution! I'll be following up with codepen embeds from vue's account later tonight.

PR #1329, opened 2017-12-19
Updated Vue style guide

dulin666:
I had a small misunderstanding with some lines of code in my team. I think this changes will provide a better approach for development based on Vue style guide.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -295,8 +295,7 @@ In our experience, it's better to _always_ add a unique key, so that you and you
 <ul>
   <li
     v-for="todo in todos"
-    :key="todo.id"
-  >
+    :key="todo.id">
     {{ todo.text }}
   </li>
 </ul>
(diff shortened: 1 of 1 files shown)


chrisvfritz:
We actually don't recommend having `>` on the same line as the last attribute, instead choosing to treat HTML the same way we treat JS. For example, you probably wouldn't write:

``` js
if (
  abc &&
  def ) {
  test
}
```

But rather:

``` js
if (
  abc &&
  def 
) {
  test
}
```

This also makes it easier to rearrange attributes and create cleaner diffs when you add or remove attributes.

dulin666:
Thanks. Sounds good.

dulin666:
I found google HTML / CSS style guide. What do you think about that?
https://google.github.io/styleguide/htmlcssguide.html#HTML_Line-Wrapping

chrisvfritz:
It doesn't change my thoughts, if that's what you're asking. Their more permissive style doesn't offer any advantages I'm aware of.

sdras:
Thanks for the PR! Since we're not going to change this, I'm going to go ahead and close it out. The research here is appreciated, though.

PR #1333, opened 2017-12-20
Update events.md

wangyi7099:
When using capture mode , parent is being handled before the children is being handled


chrisvfritz:
Thanks! This was changed a little while ago.

PR #1334, opened 2017-12-20
Updated style guide

dulin666:
Updated style guide with one more thing in HTML based on Multi-attribute elements.


chrisvfritz:
Thanks!

PR #1339, opened 2017-12-21
Updated Vue style guide with a new rule

dulin666:
Updated Vue style guide with a new rule. HTML and Vue attributes/directives in tag must be ordered.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1276,6 +1276,64 @@ In JavaScript, splitting objects with multiple properties over multiple lines is
 
 
 
+### HTML and Vue attributes/directives in tag <sup data-p="b">strongly recommended</sup>
+
+**HTML and Vue attributes/directives in tag must be ordered.**
+
+Mixed describing attributes in your templates make them less readable. Writing Vue attributes/directives in tag or component need to be after native HTML attributes.
+
+{% raw %}<div class="style-example example-bad">{% endraw %}
+#### Bad
+
+``` html
+<img
+  @click="toggle"
+  src="https://vuejs.org/images/logo.png"
+  alt="Vue Logo"
+>
+```
+
+``` html
+<CustomComponent
+  v-if="true"
+  class="custom-class"
+/>
+```
+{% raw %}</div>{% endraw %}
+
+{% raw %}<div class="style-example example-good">{% endraw %}
+#### Good
+
+``` html
+<img
+  class="vue-logo"
+  src="https://vuejs.org/images/logo.png"
+  alt="Vue Logo"
+  @click="toggle"
+>
+```
+
+``` html
+<CustomComponent
+  class="custom-class"
+  v-if="true"
+/>
+```
+
+``` html
+<button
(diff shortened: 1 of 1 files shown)


sdras:
Thanks @dulin666 for your contribution. Though it's well worded, I'm not sure that I agree with this. I write them in the other order because typically when I'm reading through the code to understand the logic, it helps me skim much faster. The part that I do agree with you strongly on is that they should be consistent- in other words, if you decide put the native attributes first, they should always be first, or vice versa. Curious what @chrisvfritz thinks as owner of the styleguide.

chrisvfritz:
There's actually [already a rule](https://vuejs.org/v2/style-guide/#Element-attribute-order-recommended) for this.

dulin666:
@chrisvfritz yes, but in this rule, I see requirements only for Vue directives/attributes and nothing about native HTML attributes. I guess we need more examples of that (for example we can add good/bad part of Element attribute order rule).
@sdras thanks, what do think about that?

chrisvfritz:
The rule does mention `id` explicitly, but for most element attributes, I'm not sure there needs to be an explicit order unless they fit into one of the other categories. And either way, it seems like a concern that's not really Vue-specific. Do you disagree?

As for good/bad examples, I didn't add them in this case because the possible combinations are nearly endless and I thought it would be self-explanatory as is. What are the details or situations you're thinking examples would clarify?

dulin666:
- In my opinion, if ```id``` section for most element attributes then we need to change it because it looks weird. We recommend writing attributes like this:
```
<custom-item
  v-for="todo in todos"
  id="single"
  class="custom-todo"
  key="todo.id"
  @whatever="update"
>
</custom-item>
```
or with button
```
<button
  v-if="true"
  class="btn"
  type="button"
  key="search-btn"
  @click="update"
>
</div>
```
It looks mixed and hard to maintain for markup developers. I guess we need provide a rule to avoid this.

- I think in this part of rules we can recommend using explicit order for attributes/directives. It will be more readable and easy to maintain for other members of developers team.

- I disagree because Vue provides us custom components for using in HTML so we need to add a recommendation for using attributes in components, it's like a props section for the component on the HTML side. 

- I agree with you about possible combinations, but my point is that we should single out HTML attributes.

chrisvfritz:
> In my opinion, if `id` section for most element attributes then we need to change it because it looks weird.

The `id` section is _not_ for most element attributes. If you read the rule I linked to, there's a very explicit section for these: **Other Attributes (all unspecified bound & unbound attributes)**.

> we should single out HTML attributes

If you mean distinguishing props from element attributes, that's actually not a concern consumers of components should be thinking about. Whether information passed to a component is treated as a prop or attribute is intentionally an internal implementation detail, rather than part of a component's interface. Props are really just an abstraction to extend the attribute interface and also to place additional constraints on them.

dulin666:
You are right Chris. Thanks for the response.

chrisvfritz:
Thanks for taking the time to discuss it and help improve the docs. 🙂

PR #1340, opened 2017-12-22
Updated style guide with prop definitions

dulin666:
Updated style guide with Prop definitions. I guess it also should be described as the bad part of this rule.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -206,6 +206,17 @@ props: ['status']
 ```
 {% raw %}</div>{% endraw %}
 
+{% raw %}<div class="style-example example-bad">{% endraw %}
+#### Bad
+
+``` js
+props: {
+   value: { type: String, required: true },
+   isVisible: { type: Boolean, default: false }
+}
+```
+{% raw %}</div>{% endraw %}
+
 {% raw %}<div class="style-example example-good">{% endraw %}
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I'm not sure I understand what this is adding. Why are those prop definitions bad?

dulin666:
I suppose we should always write props of the prop data in each new row. It will be more detailed and readable. Isn't it?
Like this:
```
props: {
   value: { 
      type: String,
      required: true
   },
   isVisible: { 
      type: Boolean, 
      default: false
  }
}
```

chrisvfritz:
There's technically the same amount of detail, but I personally agree that it's more readable. Unfortunately, the reasons for it being more readable aren't really specific to Vue, but instead more general to JavaScript objects. In these cases, we try to stay unopinionated because there are already a lot of projects for enforcing general code style (e.g. Prettier, StandardJS, popular ESLint configs). See this from the intro to the style guide:

> For the most part, we also avoid suggestions about JavaScript or HTML in general. We don’t mind whether you use semicolons or trailing commas. We don’t mind whether your HTML uses single-quotes or double-quotes for attribute values. Some exceptions will exist however, where we’ve found that a particular pattern is helpful in the context of Vue.

dulin666:
Got the picture, Chris. Thanks for the response.

PR #1348, opened 2017-12-26
fix wrong anchor link in sidebar in API detail page

uleec:
**Bug Description**

Navigation Bar in [API Detail Page](https://vuejs.org/v2/api/), `href` attribute are set to wrong value in some anchor link, when click, the corresponding content in the content section can not scroll to viewport. Check the links bellow `Instance Methods / Data` and `Instance Methods / Events`.

**Why it happened**

The anchor link `href ` value in navigation bar and id value for API `<h3>` element in content section have different value source.

id for `<h3>` element in the content section has two value source:

- if regular expression `/^([^(]+)\(/` can not match any substring in `textContent` in `<h3>`, then id comes from id value in `<h3>` in HTML template.

- if `/^([^(]+)\(/` do match some substring in `textContent`, then id comes from processed `textContent` value.

While the `href ` value in anchor link just comes from id value of `<h3>` in HTML template. So, `href` value in anchor link and id value in content section can be different.

**Solution**

- Unite the source of id value in `<h3>` in content section: All come from `textContent` of `<h3>` elements. 

- Unite the source of anchor link `href ` value in navigation bar and id value of  `<h3>`  elements in content: All come from `textContent` of `<h3>` elements.

The solution has been tested.
Wish I have explain the problem clearly.

### DIFF
--- a/themes/vue/source/js/common.js
+++ b/themes/vue/source/js/common.js
@@ -14,8 +14,9 @@
       var apiTitles = [].slice.call(apiContent.querySelectorAll('h3'))
       apiTitles.forEach(function (titleNode) {
         var methodMatch = titleNode.textContent.match(/^([^(]+)\(/)
+        var idWithoutArguments = slugize(titleNode.textContent)
         if (methodMatch) {
-          var idWithoutArguments = slugize(methodMatch[1])
+          idWithoutArguments = slugize(methodMatch[1])
           titleNode.setAttribute('id', idWithoutArguments)
           titleNode.querySelector('a').setAttribute('href', '#' + idWithoutArguments)
         }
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks! I just [merged this](https://github.com/vuejs/vuejs.org/pull/1366) to solve the same issue.

Alex-Sokolov:
@chrisvfritz looks like the problem still exist without this. See https://github.com/vuejs/vuejs.org/issues/1352

rekateka:
Hey @Alex-Sokolov, I believe a (manual?) deployment is still pending because https://vuejs.org/js/common.js and https://github.com/vuejs/vuejs.org/blob/master/themes/vue/source/js/common.js are different.

PR #1349, opened 2017-12-28
Add description and tip about `.passive` modifier

re-fort:
`.passive` modifier was added in Vue 2.3.0, but I couldn't find its detailed description in the doc, so I added it.


sdras:
This looks good to me! I like the tip especially.

kazupon:
@re-fort  good job!  👍

PR #1350, opened 2017-12-28
Retrieve tweeningValue from onUpdate callback in documentation

Boydbueno:
The tweeningValue no longer seems to be available in the tween object itself. Instead, the tweeningValue is available in the tweened object that passed as a parameter to the onUpdate callback.


ypnos-web:
With the current version of Tween.js, this patch is actually a regression. See #1574.

PR #1363, opened 2018-01-02
[Doc EN]: `event.md` add space to new part

Haeresis:
Hi @chrisvfritz and @posva !

Nothing too complicated, just presentation files rules fix ;)


posva:
I think there should be no line break between those paragraphs (it's good the way it is), they are talking about the same thing

Haeresis:
I hesitate in first place :) It's done @posva!

posva:
I meant in the final rendered result 😆 
I couldn't find that in the actual docs and assumed a new line was used to make it easier to edit (since markdown bring them back together) but we will wait for Chris input

Haeresis:
Let me know what is your choice. 

From my observation accross all documentation I have translated and corrected when something should be written in multiple paragraphs the correct syntax was: 

```
A

B

C
```

and when something should be written in one paragraph, the correct syntax was:

```
A B C
```

That allows a consistent generation across some parser.

The following syntax is currently not used:

```
A
B
C
```

because it is less predictive (parser will create one paragraph or tree ?)

PR #1366, opened 2018-01-04
fix: Correct sidebar anchor targets

rekateka:
Change function execution order to correct sidebar anchor targets; more concise fix to https://github.com/vuejs/vuejs.org/pull/1348.


rekateka:
Thanks, @chrisvfritz - I assume the file remains to be manually deployed in the future, right? Or should this have taken place automatically?

chrisvfritz:
We recently updated this repo to automatically deploy master. :-)

Issue #1344, opened 2017-12-23
How/what to test when testing vue components


lmiller1990:
Hi,

I would like to contribute to the cookbook, regarding how and what to test when testing Vue components. 

I wrote [this](https://github.com/vuejs/vue-test-utils/pull/282) for the `vue-test-utils` documentation recently. I also write a lot about testing Vue components [here](https://medium.com/@lachlanmiller_52885), so maybe something based on those articles.

The main things I want to convey are

- what is in the scope of unit tests
- when to mock/stub
- common pitfalls

Edit: I do not know how to assign a label.

sdras:
Sounds like a great addition! We'd love to have something like this in the cookbook. You probably already found it, but there's [a template](https://vuejs.org/v2/cookbook/cookbook-contributions.html) we ask people to work from in the cookbook section. Let me know if you need anything else to get started. 🎉

lmiller1990:
Thanks @sdras . I will be working [here](https://github.com/lmiller1990/vuejs.org/tree/1344-cookbook-what-to-test), and ping you with a PR for feedback. Thanks!

sdras:
You did it! Congrats! Thank you for your contribution 🙏

PR #1369, opened 2018-01-05
[Cookbook] Unit Testing components

lmiller1990:
For #1344 

First draft. Some high level feedback would be great.

- am I heading the right direction?
- are the examples too simple/complex?
- general structure/content


sdras:
Excellent! I'm slammed at work today but can't wait to look it over this weekend. Thanks!

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
+---
+title: Unit Testing Vue Components
+type: cookbook
+order: 1.2
+---
+
+## Simple Example
+
+Unit testing is a fundamental part of software developmenet. The underlying idea is tests the smallest units of code, in isolation from others. This makes it easy to refactor, add new features, or track down bugs. Vue's [single-file components](./single-file-components.html), it is straight forward to write unit tests for components in isolation. This lets you develop new features with confidence you are not breaking existing ones, and helps other people to understand what your component does.
```
sdras: Great opening, a couple of things:
1) s/developmenet/development
2) Second sentence isn't grammatically correct, how about something like: Unit tests evaluate the smallest units of code in isolation, in order to track down bugs, or make it easier to track down bugs or refactor.
3) No comma necessary after link to sfc- so it would be "Vue's single file components make it straightforward..."

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+      'Lachlan'
+    }
+  })
+
+  // assert the error has gone away
+  expect(wrapper.find('error').exists()).toBe(false)
+})
+```
+
+This simple example shown how to test whether an error message is rendered based on the length of the username. It demonstrates the general idea of unit testing Vue components: render the component, and make assertions to check if the markup matches the state of the component.
```
sdras: You said simple example earlier, how about "This shows how to ..."

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+- Facilitate refactoring
+
+The benefits of automated testing are great and it what allows large teams of developers to maintain complex codebases. 
+
+#### Getting started
+
+[vue-test-utils](https://github.com/vuejs/vue-test-utils) is the official library for unit testing Vue components. The (vue-cli)[https://github.com/vuejs/vue-cli] webpack template comes with either Karma or Jest, both well supported test runners, and there are some (guides)[https://vue-test-utils.vuejs.org/en/guides/] in the `vue-test-utils` documentation.
+
+
+4. Discuss why this may be a compelling pattern. Links for reference are not required but encouraged.
```
sdras: This is from the template, should be removed :)

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+  })
+
+  // assert the error has gone away
+  expect(wrapper.find('error').exists()).toBe(false)
+})
+```
+
+This simple example shown how to test whether an error message is rendered based on the length of the username. It demonstrates the general idea of unit testing Vue components: render the component, and make assertions to check if the markup matches the state of the component.
+
+## Details about the Value
```
sdras: I don't think you need to literally say "details about the value, I think you could change the h2 directly to "Why Test"

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+
+
+4. Discuss why this may be a compelling pattern. Links for reference are not required but encouraged.
+
+## Real-World Example
+
+Unit tests should be
+- Fast to run
+- Easy to understand
+- Only test a _single unit of work_
```
sdras: Perfect. Straightforward, easy to reproduce, wonderful list and explanation.

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+There are some problems with the above:
+- a single test is making assertions about different things
+- difficult to tell the different states the component can be in, and what should be rendered
+
+The below example improves the test by:
+- only making one assertion per `it` block
+- having short, clear test descriptions
+- providing only the minimum data requires for the test
```
sdras: Great, again, good job being clear and concise. The demo illustrates exactly this as well.

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+
+## Additional Context
+
+Thee above test is fairly simple, but in practise Vue components often have other behaviors you want to test, such as:
+
+- making API calls
+- committing or dispatching mutations or actions with a `Vuex` store
+- testing interaction
+
+`vue-test-utils` and the enormous JavaScript ecosystem provides plenty of tooling to facilitate almost 100% test coverage. Unit tests are only one part of the testing pyramid, though. Some other types of tests include e2e (end to end) tests, and snapshot tests. Unit tests are the smallest and most simple of tests - they make assertions on the smallest units of work, isolating each part of a single component.
```
sdras: It might be worth mentioned what e2e and snapshot tests are in contrast to unit tests (only a very short description is needed)

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,234 @@
...
+
+Unit tests are most useful during development, either to help a developer think about how to design a component, or refactor an existing component, and are often run every time code is changed. 
+
+Higher level tests, such as end to end tests, run much slower. These usually run pre-deploy, to ensure the everything is still working together correctly.
+
+More information about testing Vue components can be found in [Testing Vue.js Applications](https://www.manning.com/books/testing-vuejs-applications) by core team member [Edd Yerburgh](https://eddyerburgh.me/).
+
+## When To Avoid This Pattern
+
+Unit testing is an important part of any serious application. At first, when the vision of an application is not clear, unit testing might slow down development, but once a vision is established and real users will be interacting with the application, unit tests (and other types of automated tests) are absolutely essential to ensure the codebase is maintainable and scalable.
```
sdras: This conclusion is perfect. Well done.

sdras:
This is excellent! It flows really nicely and is very easy to follow. I put in some minor grammar changes that need to be fixed, as well as some clean up, but we're close. I'd like to see if Codebryo can do a quick review as unit testing is his domain. I think you can get the grammar changes done now, though. Thank you!

lmiller1990:
Thanks a lot @sdras . I updated the PR based on your feedback, and make some small changes, mainly making things more concise and basic grammar/wording.

codebryo:
Hi @lmiller1990 ,

Sarah asked for a review, so here it is :)
Upfront, thanks for the hard work you are putting into this and I am super excited that this gets made :)

[The Simple Example](https://github.com/lmiller1990/vuejs.org/blob/95885fff0b21df83dd406043d5614cddfe45023c/src/v2/cookbook/unit-testing-vue-components.md#simple-example)

In the current form this code example can’t work properly

*Original Code*

```js
test('Foo', () => {
  // render the component
  const wrapper = shallow(Hello)

  // assert the error is rendered
  expect(wrapper.find('error').exists()).toBe(true)

  // update the name to be long enough
  wrapper.setData({
    username: {
      'Lachlan'
    }
  })

  // assert the error has gone away
  expect(wrapper.find('error').exists()).toBe(false)
})
```

`error` is a class, so it should be `find('.error')`  
If it’s preferred to validate the value of `error` it should be done using `wrapper.vm.error`.
Evaluating that the element exists on the page seems good as it verifies that something is on the page while also being sure the computed prop is tested.

The usage of `.error` seems to be correct further down in the document.
In general though I would use this first example to already educate users in thinking critical while writing specs.
For example I would update the computed property to the following:

```js
computed: {
    error () {
      return this.username.trim().length < 7
        ? 'Please enter a longer username'
        : ''
    }
  }

```

The spec then could look like this already and catch a easy bug right away.

```js
test('Foo', () => {
  // render the component
  const wrapper = shallow(Hello)

 wrapper.setData({
    username: {
      '       ' // 7 spaces
    }
  })


  // assert the error is rendered
  expect(wrapper.find('error').exists()).toBe(true)

  // update the name to be long enough
  wrapper.setData({
    username: {
      'Lachlan'
    }
  })

  // assert the error has gone away
  expect(wrapper.find('error').exists()).toBe(false)
})
```

When writing unit tests it's not only about "Writing a spec" but also really testing that piece of code and be sure it's not breaking.
Teach the user to think about this in advance.

[The Real life example](https://github.com/lmiller1990/vuejs.org/blob/95885fff0b21df83dd406043d5614cddfe45023c/src/v2/cookbook/unit-testing-vue-components.md#real-world-example)

Latest at this point in the document I would account for the `trim()` method as well.
Going forward, when the tests are cleaned up and get presented in a "this is how to do it" fashion 
it instantly throws a lot of repetition at the user.
Again I would like to educate users in a good way when it comes to writing multiple specs. 
So it's a perfect opportunity to introduce a factory function.

So I would change this:

```js
import { shallow } from 'vue-test-utils'

describe('Foo', () => {
  it('renders a message', () => {
    const wrapper = shallow(Foo, {
      data: {
        message: 'Hello World'
      }
    })

    expect(wrapper.find('.message').text()).toEqual('Hello World')
  })

  it('renders error when username is less than 7 characters', () => {
    const wrapper = shallow(Foo, {
      data: {
        username: ''
      }
    })

    expect(wrapper.find('.error').exists()).toBeTruthy()
  })

  it('does not render error when username is 7 characters or more', () => {
    const wrapper = shallow(Foo, {
      data: {
        username: 'Lachlan'
      }
    })

    expect(wrapper.find('.error').exists()).toBeFalsey()
  })
})
```

To this:

```js
import { shallow } from 'vue-test-utils'

const factory = (values = {}) => {
  return shallow(Foo, {
    data: { ...values }
  })
}

describe('Foo', () => {
  it('renders a message', () => {
    const wrapper = factory({ message: 'Hello World' })

    expect(wrapper.find('.message').text()).toEqual('Hello World')
  })

  it('renders error when username is less than 7 characters', () => {
    const wrapper = factory({ username: '' })

    expect(wrapper.find('.error').exists()).toBeTruthy()
  })

  it('does not render error when username is 7 characters or more', () => {
    const wrapper = factory({ username: 'Lachlan' })

    expect(wrapper.find('.error').exists()).toBeFalsey()
  })
})
```

lmiller1990:
@codebryo thanks for the detailed feedback. I'll update me PR using your advice by the end of the week.

lmiller1990:
@codebryo  thanks for the feedback. @sdras if you have a moment maybe you can also take a look.

I made these changes:

Simple Example:
- updated first simple test to include `trim` and test for the case of whitespace

Real world example:
- Updated to use factory function. Also include brief explanation of factory function.

Other:
- added link to `vue-test-utils` for more guides
- some grammar and spelling errors
- actually run all the tests in a real project to make sure they all work.

Please let me know if I should change anything else.

sdras:
This is looking really good! I'm going to go investigate what's going on with the checks and give it one final detailed read-through, but I think we're close to merging. Great work! I love this contribution.

sdras:
Hi @lmiller1990!
I think I know why this is failing- you designated the order of this cookbook entry as the same as one that already exists- form validation. Can you please update to 1.3 or 1.4? Thanks!

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,247 @@
...
+
+```js
+import { shallow } from 'vue-test-utils'
+
+test('Foo', () => {
+  // render the component
+  const wrapper = shallow(Hello)
+
+  // should not allow for username less than 7 characters, excludes whitespace
+  wrapper.setData({ username: ' '.repeat(7) })
```
codebryo: Nice addition. <3

src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -0,0 +1,247 @@
...
+    const wrapper = factory({ username: 'Lachlan'  })
+
+    expect(wrapper.find('.error').exists()).toBeFalsy()
+  })
+})
+```
+
+Points to note:
+
+At the top, we declare the factory function, which simply returns a new `wrapper` instance, and sets and the `values` object to `data`. This way, we don't need to duplicate `const wrapper = shallow(Foo)` in every test. Another great benefit to this is when more complex components with a method or computed property you might want to mock or stub in every test, you only need to declare it once.
```
codebryo: and sets ~and~ the `values`

codebryo: That second part of the sentence in general seems a little off.
What about: 

> At the top, we declare the factory function, which simply returns a new `wrapper` instance, and merges the`values` object into `data`.

lmiller1990:
Thanks both, I'll update with @codebryo 's points and change entry number tonight!

sdras:
It might be due to some changes that were made between the time you forked and when you submitted the PR. The weird part is that you have only changed that one file. Can you please try to rebase? If that doesn't work, I'll look into it more and get it merged for you.

lmiller1990:
Rebasing did the trick!

PR #1379, opened 2018-01-10
Modify the api doc of extends.

ustbhuangyi:
Since the merge logic are the same, user can use `mixins` instead of `extends`.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -977,7 +977,7 @@ type: api
 
   Allows declaratively extending another component (could be either a plain options object or a constructor) without having to use `Vue.extend`. This is primarily intended to make it easier to extend between single file components.
 
-  This is similar to `mixins`, you can use `mixins` instead of `extends`.
+  This is similar to `mixins`.
 
 - **Example:**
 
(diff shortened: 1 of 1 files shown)


ustbhuangyi:
refer to this issue: https://github.com/vuejs/vue/issues/7412

sdras:
I see where you're going with this and I approve, but the way that it was changed make that last sentence superfluous. Perhaps that last sentence should no longer be it's own paragraph and just say "this is similar to mixins"

jungleBadger:
Hi there, I also tried to find details about extends/mixins and after some tests and research, one thing that can I think that can be enhanced in the docs is adding the information that `extends` can incorporate the `<template>` tag in the host component as well whereas mixins don't (perhaps the render function, but it isn't the same, right?).

Does mixin provide the same functionality? Would it make sense to add this to the documentation (I can make the PR)?


Another thing that I found odd is this quote:

> they will be merged against the eventual options using the same option merging logic in Vue.extend(). e.g. If your mixin contains a created hook and the component itself also has one, both functions will be called.

This is confusing to me because both extends and mixin provides the same behavior (calling the hooks twice if needed) and this passage gives an idea this is caused by calling Vue.extend() method. I thought it might be a bug as well, but after reading through [vuejs/vue#7412](https://github.com/vuejs/vue/issues/7412)  I found that this is normal, so perhaps is a way to enhance the docs as well?



Please let me know if my assumptions do make sense and if does I can surely open the  PR with the suggested changes.

PR #1380, opened 2018-01-11
Showing all possible params of watch()

ffxsam:
Surprised that the documentation didn't include the fact that the 2nd argument is the previous value. This can be really useful in some cases!


sdras:
Agreed, I think this does clarify the example. 👍

ffxsam:
Thanks!

PR #1384, opened 2018-01-15
Describe `data` merging via mixins

nylki:
I was looking for an answer on how to mix data from a mixin object inside component objects, but couldn't find one in the docs.
By experimentation I figured it out myself, but I think this should be part of the official mixin documentation.

Other aspects of mixin usage are explained (like what happens with hook functions) but there was previously no explaination how `data` merging is handled with mixins.
This is a proposal, I'd be happy to change the wording or the example if it is not fitting.

### DIFF
--- a/src/v2/guide/mixins.md
+++ b/src/v2/guide/mixins.md
@@ -31,6 +31,33 @@ var Component = Vue.extend({
 var component = new Component() // => "hello from mixin!"
 ```
 
+## Data Merging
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks for bringing this up. 🙂 I decided on a slightly different structure and example, which I just added in this commit: a49c314f90776e3cee6ba7ebad0451f0ca2f647e. I'd love hear your thoughts!

nylki:
@chrisvfritz You are welcome.
Your version makes sense to me, also useful to know that it is a shallow merge 👍

Issue #1382, opened 2018-01-14
Inaccuracy in comparisons guide (ownership of weex)


HyndexCoder:
Weex has been transferred from Alibaba group to Apache foundation.

chrisvfritz:
@Jinjiang Can you confirm this?

Jinjiang:
Exactly, Weex is just incubating from Alibaba Group to Apache Foundation. This requires a time course. I think we can update the doc when the incubation success. Or add a explanation if necessary.

Thanks :-)

chrisvfritz:
@Jinjiang Whichever you think is best sounds good to me. 🙂

Jinjiang:
@chrisvfritz I have create a PR for this. And @Hanks10100 please have a review.
Thanks.

Hanks10100:
+1. I think we should update the document to explain the relationship between Weex and ASF for now, then update it again when Weex finished the incubating.

PR #1396, opened 2018-01-22
Updated description of Weex

Jinjiang:
Ref: #1382
/ping @Hanks10100
Thanks.

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -120,7 +120,7 @@ Since you don't need to know about JSX, ES2015, or build systems to get started 
 
 ### Native Rendering
 
-React Native enables you to write native-rendered apps for iOS and Android using the same React component model. This is great in that as a developer, you can apply your knowledge of a framework across multiple platforms. On this front, Vue has an official collaboration with [Weex](https://weex.apache.org/), a cross-platform UI framework driven by Alibaba Group and being incubated by the Apache Software Foundation (ASF), which uses Vue as one of its JavaScript framework runtimes. This means with Weex, you can use the same Vue component syntax to author components that can not only be rendered in the browser, but also natively on iOS and Android!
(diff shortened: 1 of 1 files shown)


src/v2/guide/comparison.md
```diff
@@ -120,7 +120,7 @@ Since you don't need to know about JSX, ES2015, or build systems to get started
...
+React Native enables you to write native-rendered apps for iOS and Android using the same React component model. This is great in that as a developer, you can apply your knowledge of a framework across multiple platforms. On this front, Vue has an official collaboration with [Weex](https://weex.apache.org/), a cross-platform UI framework driven by Alibaba Group and being incubated to Apache Software Foundation (ASF), which uses Vue as its JavaScript framework runtime. This means with Weex, you can use the same Vue component syntax to author components that can not only be rendered in the browser, but also natively on iOS and Android!
```
Hanks10100: "... which uses Vue as one of its ..."

I think adding "one of" will be more accurate.

Jinjiang: Updated. Is that OK?

Jinjiang:
Yes it already has another react-compatible framework called rax. https://github.com/alibaba/rax

chrisvfritz:
@Jinjiang Thanks for clarifying. 🙂 I might want to revise that a bit, because from the wording it could be interpreted that Vue is one of the JS frameworks that _the ASF_ uses. 

Actually, I wonder if we might be complicating this section too much. What would you think if we just changed it to say Weex was _created_ by Alibaba Group? For example: 

> [...] On this front, Vue has an official collaboration with [Weex](https://weex.apache.org/), a cross-platform UI framework created by Alibaba Group. Weex allows you to use the same Vue component syntax to author components that can not only be rendered in the browser, but also natively on iOS and Android!

Jinjiang:
That's OK for me. 👌
And I hope @Hanks10100 could have a check too.
Thanks

Hanks10100:
Use "created" instead of "driven" looks good for me. But I think we should still mention the ASF, even if not mentioned Alibaba.

How many front-frameworks Weex support does not matter here in Vue's document. Mention it or not both fine.

Jinjiang:
@chrisvfritz @Hanks10100 agree with both of you.
And because there are already some announcement about the incubation by ASF, so I preserve the ASF part description.
How about now? 🙂

PR #1403, opened 2018-01-28
Documented Boolean falg behaviour

VsevolodTrofimov:
It dosesn't seem to be in the docs, even though it's vue specific behavior that could both be used to improve code quality and lead to floating bugs.

Raised issue here https://github.com/vuejs/vue/issues/7484

Was told to open PR to the docs.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -460,6 +460,29 @@ In addition, `type` can also be a custom constructor function and the assertion 
 
 When prop validation fails, Vue will produce a console warning (if using the development build). Note that props are validated __before__ a component instance is created, so within `default` or `validator` functions, instance properties such as from `data`, `computed`, or `methods` will not be available.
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
The value `undefined` is not actually evaluated as `true` for boolean props, as you'll [see here](https://jsfiddle.net/u951yq7u/). With that clarified, I don't believe there aren't any bugs to warn about. 

As for the remaining item - the valueless attribute option - I'm not sure that's worth expanding the documentation for, since that's just expected behavior for boolean attributes. From the HTML spec:

> The presence of a boolean attribute on an element represents the true value, and the absence of the attribute represents the false value.

Let me know if that makes sense or you disagree, as I'm happy to reopen if I've missed something.

VsevolodTrofimov:
Yeah, sorry, there was a bug in our project regarding the undefined.

PR #1405, opened 2018-01-29
feat(functional): describe how to pass down attributes

posva:
I see people not knowing this. Latest issue about it https://github.com/vuejs/vue/issues/7554

### DIFF
--- a/src/v2/guide/render-function.md
+++ b/src/v2/guide/render-function.md
@@ -579,6 +579,22 @@ Vue.component('smart-list', {
 })
 ```
 
+### Inheriting attributes like regular components
+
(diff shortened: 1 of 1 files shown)


blake-newman:
@posva see: https://github.com/vuejs/vue/issues/7557

Relates to this (in particular when not specifying prop definition)

chrisvfritz:
@posva @blake-newman I tweaked the section slightly, noting how this passes event listeners as well and the slight difference from normal components in that case. What do you think?

posva:
Love that last sentence 😆

posva:
LGTM I think the problem specified by Blake is a bit different (it looks like a bug to me)

PR #1406, opened 2018-01-30
Update： reactivity.md

forzeny:
@2.5+ $nextTick was using setImmediate > MessageChannel > setTimeout
@<2.5 $nextTick was using Promise > MutationObserver > setTimeout


forzeny:
it was different between domEvent and default deal for $nextTick， the real order is promise -> setImmediate  -> MessageChannel -> setTimeout ?

PR #1419, opened 2018-02-06
DOCS: Quotes around class are unnecessary when using createElement

snaptopixel:
Related to vuejs/vue#7605


snaptopixel:
Here you go @LinusBorg

PR #1429, opened 2018-02-13
Update list.md

01abhishekjain:
1. Give example to for better understanding of caveats.
2. Inform about `vm.$set` instance method as another way to circumvent the caveat.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -291,7 +291,7 @@ Vue.set(vm.items, indexOfItem, newValue)
 vm.items.splice(indexOfItem, 1, newValue)
 ```
 
-You can also use the `vm.$set` instance method, which is an alias for the global `Vue.set`:
+You can also use the [`vm.$set`](https://vuejs.org/v2/api/#vm-set) instance method, which is an alias for the global `Vue.set`:
 
 ``` js
 vm.$set(vm.items, indexOfItem, newValue)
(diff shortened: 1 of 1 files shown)


src/v2/guide/list.md
```diff
@@ -268,21 +268,39 @@ Due to limitations in JavaScript, Vue **cannot** detect the following changes to
...
-Vue.set(example1.items, indexOfItem, newValue)
+Vue.set(vm.items, indexOfItem, newValue)
 ```
 ``` js
 // Array.prototype.splice
-example1.items.splice(indexOfItem, 1, newValue)
+vm.items.splice(indexOfItem, 1, newValue)
+```
+
+You can also use the `vm.$set` instance method, which is an alias for the global `Vue.set`:
```
sdras: I think this needs a link to the API docs to be included: https://vuejs.org/v2/api/#vm-set

sdras:
This looks like a really good addition, I just have one comment for a fix.

src/v2/guide/list.md
01abhishekjain: Sounds like a good idea, especially since `vm.$set` is also talked about later, in the [Object change detection caveats](https://vuejs.org/v2/guide/list.html#Object-Change-Detection-Caveats) section.

OSS noob question: how should I go about it? Should I edit my commit or add a new commit to this PR? What?

sdras:
Looks good!

PR #1437, opened 2018-02-16
Update events.md

01abhishekjain:
The capture phase event propagation trickles downwards, so any event registered on an element on the capture phase will execute _before_ any other event is executed on any of its child elements.


sdras:
Thanks

01abhishekjain:
Achievement unlocked: Get a PR merge by Sarah Drasner!

Woohoo!!! 🎉

Issue #1439, opened 2018-02-18
Docs - Improve typographic styles for print media


NiloCK:
A number of stylistic choices for the documentation work much better for web than for print.

 - Dark grey (and not black) text on a white background
 - 1.6em line height
 - code snippet boxes with greyed backgrounds

These could be improved in a `@media print` query.

NiloCK:
One problem persists after the commits above - page breaks are occurring directly after `h1`, `h2`, `hn` blocks. The css properties `page-break-after` and `page-break-before` aren't very reliable.

One solution is described [here](http://www.pagetoscreen.net/journal/item/oh_please_avoid_those_page_breaks_after), and involves wrapping the headers and the next element with a div that's given the more reliable `page-break-inside: avoid` property. Doing this via JS on page load is easy enough, but **some might consider that a bit too hacky and intrusive**.

If no-one objects to the notion (or suggests a better alternative) in the next few days, I'll have a look at implementing.

sdras:
Hm, that does sound a bit hacky indeed, especially for not-so-common use case. What about trying with page-break-after and page-break-before and testing first- maybe it will prove to be more effective than we think. Worth trying first, anyways. :)

I added a comment about links as well. Sounds like the right track for everything else.

chrisvfritz:
This has been merged! Happy to keep accepting new contributions, but considering this resolved for now. 🙂

PR #1440, opened 2018-02-18
Improve typography for print media (#1439)

NiloCK:
This commit leaves the docs in reasonably good shape for printing from the web. Can close (#1439) after the merge.


sdras:
It would be good to have the print styles include the links within the document, accomplished with `a:link:after, a:visited:after { content:" [" attr(href) "] "; }`

chrisvfritz:
Great suggestion @sdras! And thank you @NiloCK for this great addition. 🙂

PR #1442, opened 2018-02-19
Update conditional.md

Misaka-0x447f:
to tell users they cannot use double quotes for "string in string" in component template.

### DIFF
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -28,6 +28,8 @@ It is also possible to add an "else block" with `v-else`:
 <h1 v-else>No</h1>
 ```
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
You actually _can_ use double quotes, you just can't mix quotes. For example, [these](https://jsfiddle.net/mhvxso8q/1/) [two](https://jsfiddle.net/fuqe226j) alternatives do work. And because this is only a limitation of in-DOM HTML, you technically [could do this](https://jsfiddle.net/c8tzcyfr/) in string templates or SFCs if you wanted.

Misaka-0x447f:
Okay..... please consider add this in the document.

Issue #1336, opened 2017-12-20
Vue.js Guide: Document use of "event parameters" in $emit(eventName, [...params]) for custom events


marko-knoebl:
I'm just learning vue.js and I noticed that https://vuejs.org/v2/guide/components.html#Custom-Events doesn't give any explanation/introduction about the optional event parameters that can be passed to a custom event.

The signature is given as just "$emit(eventName)".

Further down optional parameters are used in code but never introduced (first appearance in the section ".sync Modifier": "this.$emit('update:foo', newValue)").

I think this should be explicitly mentioned at the beginning of the section.

By the way, all in all the vue.js guide is amazing, thanks for the great work!

sdras:
I've also noticed this area of the documentation is thin and could be explained in more detail. This is a good point, I'll try to put a PR in for it soon. Thanks!

chrisvfritz:
One additional note: we don't have to explicitly explain this, but in our examples I think it'd be good to follow the convention of emitting a single payload, rather than multiple arguments. For example, instead of:

```js
$emit('some-event', 'foo', 3)
```

Doing this:

```js
$emit('some-event', { text: 'foo', count: '3' })
```

sdras:
This is merged 🎉

HarinderKaur11:
Hi, 

@sdras @chrisvfritz 
I am new to VueJs and I am doubtful about passing the optional payload
Could you please tell me how to pass the value returned by a computed function in the child component to the parent component using the optional payload.

I want to implement a separate search component which returns the search results to all other components. The computed function looks like this:

get computedSports () {
        if (!this.searchModel || this.searchModel.length === 0)
            return this.sports
        else
            return this.fuseSearch.search(this.searchModel)
    }


This is how I am trying to pass the value returned by computed function to its parent component:
@input="$bus.$emit('computed-sports', computedSports)"

In the parent component, this is how I want access the value of that function:
v-on:computed-sports="" <--- what value should come here?

chrisvfritz:
Sounds like you're building a fun app. 🙂 I believe it should work if you remove the `$bus` part, then you can use a method as the value of the event listener in the parent. That method will receive the value of `computedSports` as the first argument.

HarinderKaur11:
@chrisvfritz Thank you very much! It worked :)

zarub2k:
@chrisvfritz Worked great!

PR #1444, opened 2018-02-19
components: document using of event parameters in $emit

dennybiasiolli:
ref. #1336

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -507,7 +507,7 @@ We have learned that the parent can pass data down to the child using props, but
 Every Vue instance implements an [events interface](../api/#Instance-Methods-Events), which means it can:
 
 - Listen to an event using `$on(eventName)`
-- Trigger an event using `$emit(eventName)`
+- Trigger an event using `$emit(eventName, optionalPayload)`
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/components.md
```diff
@@ -507,7 +507,7 @@ We have learned that the parent can pass data down to the child using props, but
 Every Vue instance implements an [events interface](../api/#Instance-Methods-Events), which means it can:
 
 - Listen to an event using `$on(eventName)`
-- Trigger an event using `$emit(eventName)`
+- Trigger an event using `$emit(eventName[, optionalPayload])`
```
sdras: This should be `$emit(eventName, [optionalPayload])`

sdras: Though, truthfully, it doesn't need to be in an array, so maybe that should be more clear, too.

sdras:
I really like this addition! This is definitely something we should be covering a bit better in the docs. I made some comments, if we resolve those we'll be close to merging.

src/v2/guide/components.md
dennybiasiolli: @sdras I put it into square brackets to specify it is optional, do you have any suggestion on the format?
If you think it's better, I can write it in this way
>  Trigger an event using `$emit(eventName)` or `$emit(eventName, optionalPayload)`
In order to make it explicit ;)

dennybiasiolli: Or I can edit as
> Trigger an event using `$emit(eventName, optionalPayload)`

if it's clear enough

sdras:
LGTM 👍

PR #1447, opened 2018-02-20
Document how to disable production mode via CLI

oz123:
While the option already documented is good, it hard codes
 production mode. Advanced users of webpack might already know,
 that you can use 2 config files (1 for development, and 1 for
 production) with a common base (thus 3 configuration files).
 However, for new comeres to modern javascript with a build step, it
 might be more straight forward to show also how to achieve the same
 via a command line option.
 This is also in paar with the section on Browserify which shows the
 CLI option.

### DIFF
--- a/src/v2/guide/deployment.md
+++ b/src/v2/guide/deployment.md
@@ -36,6 +36,13 @@ module.exports = {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR (and sorry it took so long!) but I don't think it's up to the standard we'd love to see in Vue.

Issue #32, opened 2014-08-26
In the benchmarks, Angular interacts with local storage.


otac0n:
The version of TODO MVC included for angular saves and loads data from local storage.  I removed this interaction and gained about 25% performance.

PR #1449, opened 2018-02-21
Traducción de Custom Directives #32

EstebanBorai:


chrisvfritz:
It looks like this was accidentally submitted here instead of to the Spanish translation. 😉

PR #1450, opened 2018-02-21
Updated computed change detection documentation

FEA5T:
As discussed:

https://github.com/vuejs/vue-devtools/issues/393

Feedback welcome.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -77,6 +77,42 @@ You can open the console and play with the example vm yourself. The value of `vm
 
 You can data-bind to computed properties in templates just like a normal property. Vue is aware that `vm.reversedMessage` depends on `vm.message`, so it will update any bindings that depend on `vm.reversedMessage` when `vm.message` changes. And the best part is that we've created this dependency relationship declaratively: the computed getter function has no side effects, which makes it easier to test and understand.
 
+### Change Detection
+
+<p class="tip">Note that computed properties are evaluated lazily.
+
+**This means that computed properties only re-evaluate when:**
+1. Their dependencies have changed, and
+2. They are subscribed to by a data binding (or another subscriber that is eventually subscribed to by a data binding)
+3. That data binding is currently rendered in the dom/template
+
+</p>
+
+If no data bindings subscribe to a computed property, the computed property will only be evaluated once (on vue instance creation).
+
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I agree that a note about this would be helpful in the Vue Devtools docs, but I don't think it belongs here in the guide. This is more of an under-the-hood optimization than something the Vue users have to keep mind when building apps. The edge case of debugging a computed property that's not even being used is also quite rare I think.

FEA5T:
@chrisvfritz 

> I agree that a note about this would be helpful in the Vue Devtools docs

Absolutely, well IMO, it should be fixed in devtools so that devtools actually updates.

> This is more of an under-the-hood optimization

It is an over-the-hood optimization because I notice it.  When you can see something and it is constantly visible...well...thats above the hood as it gets.

> than something the Vue users have to keep mind when building apps.

It is something I have to constantly keep in mind when building apps.  Maybe your workflow is different, that doesn't make mine (and many others') invalid.

> The edge case of debugging a computed property

What?  Is debugging an edge case now?

Verifying that a computed property works before consuming it is not an edge case.  By nature, you cannot consume a function until it is written (and written correctly).  As an analogy I check to see if my pan is hot before I put food in the pan.  Knowing if my pan is hot is not an edge case.  It's part of the dev flow, developer experience (DX).  Observing that a function runs even once is quite common before consuming the return value (or in this case consuming via dom).  In fact, it is half the point of dev-tools.  This is js after all, not precompiled code...

> quite rare I think.

Took me two years and many days of debugging with no solution to learn about this.  Ran into this almost every week.

I am giving vue the best kind of user feedback, really, pure user feedback from actual experience using the lib and what I am hearing is basically "I didn't have that issue, so, it doesn't matter".  The reality is that when a few users are complaining that you have a problem, there are really 5000 out there that don't care enough to speak up.

Sorry for ranting on you.  This is like the 7th thing in a row I have suggested that has been closed for a less-than-adequate reason.

chrisvfritz:
@FEA5T Honestly, I'm hesitant to respond at all with your combative tone and deliberately taking my words out of context in this post - if your issues are being closed, that might be a big part of it. And please remember, the Vue team is looking at a much bigger picture that we don't always have time to fully explain, so please try to give us the benefit of the doubt. 🙂

PR #1452, opened 2018-02-21
Update computed.md with explanation about v-for

kslstn:
Explanation about not using computed on v-for element properties and alternative solutions.


src/v2/guide/computed.md
```diff
@@ -77,6 +77,8 @@ You can open the console and play with the example vm yourself. The value of `vm
 
 You can data-bind to computed properties in templates just like a normal property. Vue is aware that `vm.reversedMessage` depends on `vm.message`, so it will update any bindings that depend on `vm.reversedMessage` when `vm.message` changes. And the best part is that we've created this dependency relationship declaratively: the computed getter function has no side effects, which makes it easier to test and understand.
```
FEA5T: "cannot be to" >> "cannot be used to",

Also, can you add 2 quick code examples, one not working with v-for, and another working with a method?

For tips, use '<p class="tip">Note that...</p>'

chrisvfritz:
Thanks for the contribution! Unfortunately, I'd rather not mention `v-for` here, as it hasn't yet been introduced at this point in the doc so will be confusing for anyone new to Vue.

A contribution like this would still be welcome on the list rendering page or anywhere in the API doc though, if you feel it would be helpful there. 🙂

PR #1453, opened 2018-02-22
add note about prop, and remove assignment

posva:
ref https://github.com/vuejs/vue/issues/7683

### DIFF
--- a/src/v2/guide/render-function.md
+++ b/src/v2/guide/render-function.md
@@ -319,6 +319,7 @@ Wherever something can be easily accomplished in plain JavaScript, Vue render fu
 This could be rewritten with JavaScript's `if`/`else` and `map` in a render function:
 
 ``` js
+props: ['items'],
 render: function (createElement) {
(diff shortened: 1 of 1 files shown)


src/v2/guide/render-function.md
```diff
@@ -343,14 +343,15 @@ render: function (createElement) {
...
       input: function (event) {
-        self.value = event.target.value
         self.$emit('input', event.target.value)
       }
     }
   })
 }
 ```
 
+_Note_: In the example above we consider a prop named `value` already exists.
```
JounQin: What about just add `props: ['value']` above `render` function?

posva: The code is just showing the render function, adding props here would make it inconsistent with the other samples

chrisvfritz:
@posva @JounQin I agree with you both, so added a prop field to the example, but also added props to other examples that might need the same for clarity and consistency. 🙂

PR #1454, opened 2018-02-22
feat(render-function): adds functional template ex for passing attrs

alexsasharegan:
# Render Function Page

Ask and thee shall receive, @chrisvfritz .
https://twitter.com/chrisvfritz/status/966792150087323653

Adds a small example of passing attrs and events to child components when in the functional template context.


chrisvfritz:
Thanks! I just simplified the example slightly (using a button instead of an SVG), but otherwise looks great!

PR #1445, opened 2018-02-20
Update events.md

01abhishekjain:
Change `@` convention to `v-on` because otherwise browser considers it to be a `mailto:` type hyperlink. See screenshot below:

screen shot 2018-02-20 at 1 18 55 pm

PR #1455, opened 2018-02-23
fix(Events): escape @ in code tags

naeramarth7:
Issue already mentioned in #1445 which replaces `@` with `v-bind`.

This Escape `@` in code tags by prepending a space favoring the shorthand.

image


chrisvfritz:
Thanks! I just updated to `v-on`, since that's also what the surrounding examples use.

Issue #1432, opened 2018-02-15
Can I contribute a blog example?


jakelumetta:
I'd like to contribute a serverless blog example to https://github.com/vuejs/vuejs.org/tree/master/src/v2/examples

It would use ButterCMS API for the blog backend. Similar to this Firebase example: https://vuejs.org/v2/examples/firebase.html

Let me know your thoughts! Thanks

sdras:
Sounds like a nice idea! I'm sure it will be useful to people.

If you decide you would like to do a write up of the example, let me know! We can add it to the [cookbook](https://vuejs.org/v2/cookbook/index.html).

Thanks!

sdras:
You can open a PR- is it an example or a cookbook recipe? I can label it appropriately here.

Examples go in the directory you pointed at, while cookbook entries are [linked up here.](https://github.com/vuejs/vuejs.org/tree/master/src/v2/cookbook)

Thanks!

jakelumetta:
@sdras Sounds good - I think it's a fit for cookbook but you tell me :)

PR is here: https://github.com/vuejs/vuejs.org/pull/1457

sdras:
Excellent! I'll have a look and get back to you! On my way to a conference but I'll review as soon as I can.

sdras:
Done!

PR #1457, opened 2018-02-26
Adding serverless blog example

jakelumetta:
PR for https://github.com/vuejs/vuejs.org/issues/1432

CC @sdras 

Looking forward to your feedback!

### DIFF
--- a/src/v2/cookbook/serverless-blog.md
+++ b/src/v2/cookbook/serverless-blog.md
@@ -1,8 +1,14 @@
+---
+title: Create a CMS-Powered Blog
+type: cookbook
+order: 5
+---
+
 # Create a serverless CMS-Powered Blog Using Vue.js
 
-So you've just launched your Vue.js website, congrats! Now you want to add a blog that quickly plugs into your website and you don't want to have to spin up a whole server just to host a Wordpress instance (or any DB-powered CMS for that matter). You want to just be able to add a few  Vue.js blog components and some routes and have it all just work, right?
+So you've just launched your Vue.js website, congrats! Now you want to add a blog that quickly plugs into your website and you don't want to have to spin up a whole server just to host a Wordpress instance (or any DB-powered CMS for that matter). You want to just be able to add a few Vue.js blog components and some routes and have it all just work, right? What you're looking for a "serverless" blog that's powered entirely by API's you can consume directly from your Vue.js application. This tutorial will teach you how to do just that, let's dive in!
 
-This tutorial will teach you how to do just that. We're going to quickly build a serverless CMS-powered blog with Vue.js. It uses ButterCMS, an API-first CMS that lets you manage content using the ButterCMS dashboard and integrate our content API into your Vue.js app. You can use ButterCMS for new or existing Vue.js projects.
+We're going to quickly build a serverless CMS-powered blog with Vue.js. It uses [ButterCMS](https://buttercms.com/), an API-first CMS that lets you manage content using the ButterCMS dashboard and integrate our content API into your Vue.js app. You can use ButterCMS for new or existing Vue.js projects.
  
 ![Butter Dashboard](https://user-images.githubusercontent.com/160873/36677285-648798e4-1ad3-11e8-9454-d22fca8280b7.png "Butter Dashboard")
 
@@ -49,7 +55,7 @@ butter.post.list({page: 1, page_size: 10}).then(function(response) {
 This API request fetches your blog posts. Your account comes with one example post which you'll see in the response.
 
 ## Display posts
-To display posts we create a simple `/blog` route (using vue-router) in our app and fetch blog posts from the Butter API, as well as a `/blog/:slug` route to handle individual posts. See our [API reference](https://buttercms.com/docs/api/?javascript#blog-posts) for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
+To display posts we create a `/blog` route (using vue-router) in our app and fetch blog posts from the Butter API, as well as a `/blog/:slug` route to handle individual posts. See our [API reference](https://buttercms.com/docs/api/?javascript#blog-posts) for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
 
 See our API reference for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
 
@@ -133,6 +139,11 @@ Display the result
 </template>
 ```
 
+Here's what it looks like (note we added CSS from https://bulma.io/ for quick styling):
+
+![buttercms-bloglist](https://user-images.githubusercontent.com/160873/36868500-1b22e374-1d5e-11e8-82a0-20c8dc312716.png)
+
+
 Now create `components/BlogPost.vue` which will be your Blog Post page to list a single post.
 
 ```javascript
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
+# Create a serverless CMS-Powered Blog Using Vue.js
```
sdras: You need to index this article, here's an example of how to do so:

```---
title: Create a Serverless CMS-Powered Blog
type: cookbook
order: 5
---
```

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
...
+
+So you've just launched your Vue.js website, congrats! Now you want to add a blog that quickly plugs into your website and you don't want to have to spin up a whole server just to host a Wordpress instance (or any DB-powered CMS for that matter). You want to just be able to add a few  Vue.js blog components and some routes and have it all just work, right?
```
sdras: picky, but there's an extra space between "few" and "Vue.js"

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
...
+
+This tutorial will teach you how to do just that. We're going to quickly build a serverless CMS-powered blog with Vue.js. It uses ButterCMS, an API-first CMS that lets you manage content using the ButterCMS dashboard and integrate our content API into your Vue.js app. You can use ButterCMS for new or existing Vue.js projects.
```
sdras: Please link up ButterCMS, i.e.
`It uses [ButterCMS](https://buttercms.com/)`

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
...
+```javascript
+butter.post.list({page: 1, page_size: 10}).then(function(response) {
+  console.log(response)
+})
+```
+
+This API request fetches your blog posts. Your account comes with one example post which you'll see in the response.
+
+## Display posts
+To display posts we create a simple `/blog` route (using vue-router) in our app and fetch blog posts from the Butter API, as well as a `/blog/:slug` route to handle individual posts. See our [API reference](https://buttercms.com/docs/api/?javascript#blog-posts) for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
```
sdras: We try to avoid the word "simple" because, depending on the skill level of the reader, it may not be very simple for them.

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
...
+      {{ post.meta.previous_post.title }}
+    </router-link>
+    <router-link v-if="post.meta.next_post" :to="/blog/ + post.meta.next_post.slug" class="button">
+      {{ post.meta.next_post.title }}
+    </router-link>
+  </div>
+</template>
+```
+
+Now our app is pulling all blog posts and we can navigate to individual posts. However, our next/previous post buttons are not working.
```
sdras: It might be nice to add a screenshot here so that people can visualize what you're referring to

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,268 @@
...
+
+One thing to note when using routes with params is that when the user navigates from /blog/foo to /blog/bar, the same component instance will be reused. Since both routes render the same component, this is more efficient than destroying the old instance and then creating a new one. However, this also means that the lifecycle hooks of the component will not be called.
```
sdras: The "however" here would be better served as a tip. Like:

```<p class="tip">Be aware, that using the component this way will mean that the lifecycle hooks of the component will not be called. Visit the Vue.js docs...</p>```

**Comment by sdras** [comment]
Overall, its a great start! I think this is a great introduction. This is also easy to follow. There are a couple of major things that need to be changed to get the PR through:

- We need the initial comment to properly index the recipe (that's probably why all the checks are failing
- This doesn't really follow the [acceptable format for cookbook entries](https://vuejs.org/v2/cookbook/cookbook-contributions.html). It think it's pretty clear we can stray a bit, with the exception of *when to avoid this pattern* and *alternative patterns*, which this particular recipe needs in order to sound like documentation and not advertisement
- You mention "Serverless" in the title and barely mention it in the article. It either needs to be clarified in the post or removed. At this point, I'm thinking removal makes more sense because it's not a central theme of the article, which seems more based around CMS and blog creation.

**Comment by jakelumetta** [comment]
@sdras Thanks for the great feedback!  I've made your suggested changes but did have one question.  Regarding your feedback on "when to avoid this pattern" - could you provide more detail in the context of adding a blog to an application? 

I could also use a bit more direction on what you're looking for in terms of "Alternative patterns".  Would this be an alternate way to add a blog to your Vue.js app?

Thanks!

**Comment by sdras** [comment]
@jakelumetta good question! You could probably explore such options as [Nuxtent](https://nuxtent.now.sh), which allow you to use Vue Component inside your markdown files, which allow you to write and process markdown files instead of using a static site generator like jekyll. That would be a good candidate for a quick compare/contrast or at least mention in case people want to know what their options are.

**Comment by jakelumetta** [comment]
@sdras Very cool. Wasn't aware of Nuxtent actually, thanks :)  

I just added an Alternative Pattern section.  Let me know your thoughts.

File: src/v2/cookbook/serverless-blog.md [hunk 7]
```diff
@@ -0,0 +1,289 @@
...
+  console.log(response)
+})
+```
+
+This API request fetches your blog posts. Your account comes with one example post which you'll see in the response.
+
+## Display posts
+To display posts we create a `/blog` route (using vue-router) in our app and fetch blog posts from the Butter API, as well as a `/blog/:slug` route to handle individual posts. See our [API reference](https://buttercms.com/docs/api/?javascript#blog-posts) for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
+
+See our API reference for additional options such as filtering by category or author. The response also includes some metadata we'll use for pagination.
```
sdras: In this section, you say "our API reference" a few times- please say "the ButterCMS API reference" as you're in the vue docs and "our API reference" should refer to Vue's API reference.

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,289 @@
...
+title: Create a CMS-Powered Blog
+type: cookbook
+order: 5
+---
+
+# Create a serverless CMS-Powered Blog Using Vue.js
+
+So you've just launched your Vue.js website, congrats! Now you want to add a blog that quickly plugs into your website and you don't want to have to spin up a whole server just to host a Wordpress instance (or any DB-powered CMS for that matter). You want to just be able to add a few Vue.js blog components and some routes and have it all just work, right? What you're looking for a "serverless" blog that's powered entirely by API's you can consume directly from your Vue.js application. This tutorial will teach you how to do just that, let's dive in!
+
+We're going to quickly build a serverless CMS-powered blog with Vue.js. It uses [ButterCMS](https://buttercms.com/), an API-first CMS that lets you manage content using the ButterCMS dashboard and integrate our content API into your Vue.js app. You can use ButterCMS for new or existing Vue.js projects.
```
sdras: You say Serverless here and at the end without any reference or explanation to what is Serverless about it, and why that's important, or how it works. I would either remove that descriptor or spend more time explaining Serverless concepts and why that matters here. 

In my opinion, it has little to do with the rest of the article, so I'd just remove the reference for clarity and flow, but it's up to you.

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,289 @@
...
+      })
+  }
+},
+created() {
+  ...
+  this.getCategories()
+  this.getPostsByCategory()
+}
+```
+## Wrap up
```
sdras: Please put the wrap up section after the alternative patterns section. I don't think you need the part about the feedback and PR, because that's kind of self explanatory, but I could go either way on it :) It's not a bad thing to include.

sdras:
This is looking great! We're almost there. I left a few more comments, when we've squared those away, I think we'll be good to merge

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,289 @@
...
+
+That's it! You now have a fully functional serverless blog running in your app.  We hope this tutorial was helpful and made your development experience with Vue.js even more enjoyable :)
+
+## Alternative Pattern
```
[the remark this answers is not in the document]
jakelumetta: 👍

src/v2/cookbook/serverless-blog.md
```diff
@@ -0,0 +1,289 @@
...
+![buttercms-blogdetail](https://user-images.githubusercontent.com/160873/36868506-218c86b6-1d5e-11e8-8691-0409d91366d6.png)
+
+
+Now our app is pulling all blog posts and we can navigate to individual posts. However, our next/previous post buttons are not working.
+
+One thing to note when using routes with params is that when the user navigates from /blog/foo to /blog/bar, the same component instance will be reused. Since both routes render the same component, this is more efficient than destroying the old instance and then creating a new one. 
+
+<p class="tip">Be aware, that using the component this way will mean that the lifecycle hooks of the component will not be called. Visit the Vue.js docs to learn more about [Dynamic Route Matching](https://router.vuejs.org/en/essentials/dynamic-matching.html)</p>
+
+To fix this we need to simply watch the `$route` object and call `getPost()` when the route changes.
```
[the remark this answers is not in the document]
jakelumetta: 👍

src/v2/cookbook/serverless-blog.md
jakelumetta: 👍

jakelumetta:
@sdras awesome - I've made all of your suggested updates. Thanks for the great editing help. 👍

sdras:
Perfect! Well done.

PR #1460, opened 2018-02-28
.Exact modifier annotation error

jxDeveloper:
Modify the.Exact modifier annotation error in events.md


re-fort:
@jxDeveloper 
I think current description is correct.
Without `.exact` modifier, it will fire events even if (Ctrl and)Alt or (Ctrl and)Shift are pressed.

PR #1461, opened 2018-02-28
Update components.md

seriousmac:
updated js code is needed to execute well. In my humble opinion, it should be added.


chrisvfritz:
Thanks! I only made small simplification, since the `Vue.component` in the earlier example will mean we don't have to re-define it locally. (We only do that in the embedded code to avoid interfering with other examples on the page.)

PR #1462, opened 2018-02-28
clarify required props and validation

decademoon:
Refer to https://github.com/vuejs/vue/issues/7720#issuecomment-369198737

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -458,6 +458,10 @@ The `type` can be one of the following native constructors:
 
 In addition, `type` can also be a custom constructor function and the assertion will be made with an `instanceof` check.
 
+If `required` is `true`, then Vue will issue a console warning (if using the development build) if the prop is omitted from the template.
+
+Prop validation does not occur when the prop value is `null` or `undefined` and `required` is `false` (the default).
+
 When prop validation fails, Vue will produce a console warning (if using the development build). Note that props are validated __before__ a component instance is created, so within `default` or `validator` functions, instance properties such as from `data`, `computed`, or `methods` will not be available.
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Thanks for the PR! I'm not sure this is additive though. My thinking is we describe the behavior of `required: true` in the example above, then talk about the console warning in the paragraph below.

Could you share more context for the confusion you're trying to avoid?

decademoon:
The docs don't explain what `required` enforces, which is exactly the following:

1. The prop cannot be omitted from the template, and
2. Validation will not be performed if the prop value is nully and `required` is false.

I was confused because I wanted to enforce that a prop's value can never be nully, but Vue cannot enforce this, so this clarification should make it clearer what `required` actually means.

Pretty much this issue https://github.com/vuejs/vue/issues/7720#issuecomment-369198737.

> My thinking is we describe the behavior of required: true in the example above, then talk about the console warning in the paragraph below.

That's fine, as long as it's clear what exactly `required` enforces.

chrisvfritz:
Thanks for explaining in more detail. I think I understand now. Please still correct me if I'm mistaken, but it sounds like there are two things you'd like the doc to be clearer on:

- That `required: true` requires the prop to receive a non-nully value.
- That the default (`required: false`) will cause `null` to skip validation (so no error is logged) and will not fall back to the default (if provided).

Is that correct?

decademoon:
Yes, that's correct.

> That `required: true` requires the prop to receive a non-nully value.

It might be worth mentioning that the prop also cannot be omitted from the template. If a prop is required and has a non-null default value, then omitting the prop from the template would still result in a warning, even though the prop's value would not be null.

PR #1464, opened 2018-02-28
Add in debugging recipe, delete unused categories from cookbook TOC

sdras:
Kenneth Auchenberg wrote a great recipe for debugging Vue within VSCode, and we agreed it should live in our cookbook as well. In addition to his great contribution, I've also added:

- Alternative pattern: vue devtools
- Alternative pattern: vuetron
- Alternative pattern: simple debugging statement
- As discussed, removal of sections for cookbook until we have a few more and can see more pertinent groupings


sdras:
@auchenberg I would love to get your take on this as well.

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
+---
+title: Debugging in VSCode and Chrome
```
auchenberg: We usually say "VS Code" or "Visual Studio Code"

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+type: cookbook
+order: 6
+---
+
+Every application reaches a point where it's necessary to understand failures, small to large. In this recipe, we explore a few workflows for VSCode users, who are using Chrome to test.
```
auchenberg: VS Code

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+This recipe shows how to use the [Debugger for Chrome](https://github.com/Microsoft/vscode-chrome-debug) extension with VS Code to debug Vue.js applications generated by the [Vue CLI](https://github.com/vuejs/vue-cli).
+
+## Prerequisites
+
+You must have Chrome and VSCode installed. Make sure to the latest version of [Debugger for Chrome](https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome) extension installed in VS Code.
```
auchenberg: VS Code

sdras:
Oh, that's too bad about the invalid sourcemaps. I'll synch with you offline about timeline. I can get the other changes in in the meantime.

src/v2/cookbook/adding-instance-properties.md
```diff
@@ -106,34 +107,40 @@ In case you're not aware, methods added to a prototype in JavaScript gain the co
...
   data: {
     message: 'Hello'
   },
-  created: function () {
-    console.log(this.message)    // => "Hello"
+  created: function() {
+    console.log(this.message) // => "Hello"
     this.$reverseText('message')
-    console.log(this.message)    // => "olleH"
+    console.log(this.message) // => "olleH"
```
chrisvfritz: It looks like these will were probably automatically formatted by Prettier. We've previously followed the StandardJS conventions in the Vue docs, but I've actually been thinking about moving towards Standard + Prettier overall, with linting built into the project so this is probably OK.

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+order: 6
+---
+
+Every application reaches a point where it's necessary to understand failures, small to large. In this recipe, we explore a few workflows for VS Code users, who are using Chrome to test.
+
+This recipe shows how to use the [Debugger for Chrome](https://github.com/Microsoft/VSCode-chrome-debug) extension with VS Code to debug Vue.js applications generated by the [Vue CLI](https://github.com/vuejs/vue-cli).
+
+## Prerequisites
+
+You must have Chrome and VS Code installed. Make sure to the latest version of [Debugger for Chrome](https://marketplace.visualstudio.com/items?itemName=msjsdiag.debugger-for-chrome) extension installed in VS Code.
```
chrisvfritz: > Make sure to the latest version

Missing word here I think.

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+Install and create a project with the [vue-cli](https://github.com/vuejs/vue-cli) by running
+
+```
+npm install -g @vue/cli
+# or
+yarn global add @vue/cli
+
+vue create my-project
+```
```
chrisvfritz: I'm wondering if we should just defer people to the instructions in the Vue CLI README. That way, we don't have to update recipes if the API of Vue CLI changes. Thoughts?

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+Change to the newly created application directory and open VS Code.
+
+```
+cd my-project
+
+# if you have an alias set up (recommended)
+code .
+```
```
chrisvfritz: This part doesn't necessarily have to be done in the terminal and I've found some people prefer to instead open VS Code, then open a folder from there. Maybe we should assume knowledge of how to use the basics of VS Code and just ask people to open the project folder? That way, we're not assuming an alias is set up and that this is how the user prefers to open projects.

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Update your webpack configuration
```
chrisvfritz: When referring to project names, we use a Title Case convention (e.g. `Webpack` instead of `webpack`). We're also currently using Title Case for titles (e.g. `Update Your Webpack Configuration` instead of `Update your webpack configuration`).

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+Before you can debug your Vue components from VS Code you need to update the generated webpack config to build sourcemaps that contains more information for our debugger.
```
chrisvfritz: Last time I'll mention it, but `webpack`->`Webpack` again here.

chrisvfritz: > sourcemaps that contains more information

`contains`->`contain`?

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+* Go to `config/index.js` and find the `devtool` property. Update it to:
+
+```json
+devtool: 'source-map',
+```
+
+### Configure launch.json File
```
chrisvfritz: `launch.json` should probably be in backticks.

src/v2/cookbook/debugging-in-vscode.md
chrisvfritz: I think we should probably also be more specific than "more information for our debugger" for those who don't know what source maps are.

src/v2/cookbook/debugging-in-vscode.md
chrisvfritz: Actually, with these titles, I wonder if it might be better to use titles that describe what we're trying to achieve, instead of the implementation details to get there.

For example, instead of `Update your webpack configuration`, we could say `Showing Source Code in the Chrome Devtools`. Then instead of `Configure launch.json File`, we could say, `Launching Our App from VS Code`.

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+      "breakOnLoad": true,
+      "sourceMapPathOverrides": {
+        "webpack:///src/*": "${webRoot}/*"
+      }
+    }
+  ]
+}
+```
+
+## Start Debugging An Application
```
chrisvfritz: Maybe `Adding a Breakpoint`?

src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Vue Devtools
+
+There are other methods of debugging, varying in complexity. The most popular and simple of which is to use the excellent [vue-devtools](https://chrome.google.com/webstore/detail/vuejs-devtools/nhdogjmejiglipccpnnnanhbledajbpd). Some of the benefits of working with the devtools are that they enable you to live-edit data properties and see the changes reflected immediately. The other major benefit is the ability to do time travel debugging for Vuex.
+
+![Devtools Timetravel Debugger](/images/devtools-timetravel.gif)
+
+<p class="tip">There are a couple of gotchas to note for Vue Devtools: 
+- If the page uses a production/minified build of Vue.js, devtools inspection is disabled by default so the Vue pane won't show up. 
+- To make it work for pages opened via file:// protocol, you need to check "Allow access to file URLs" for this extension in Chrome's extension management panel</p>
```
chrisvfritz: Whenever there's another doc that explains something in more detail, it's best to link to it so that we don't have to maintain multiple versions of the same instructions. In this case, I'd just link to the Vue Devtools README, since it contains all this information already.

src/v2/cookbook/debugging-in-vscode.md
sdras: Yeah, that makes a lot of sense to me.

src/v2/cookbook/debugging-in-vscode.md
sdras: Yeah, we don't mention here how to set up the alias, either, so this might avoid confusion

src/v2/cookbook/debugging-in-vscode.md
sdras: Yeah, that makes more sense and is probably more easily googleable (not a word, but I'm keeping it)

src/v2/cookbook/debugging-in-vscode.md
sdras: Good idea.

sdras: Actually, I'm thinking about keeping a note about it not working for minified version, but maybe changing the wording to say "for instance, using a CDN" because I get this question _a ton_ and it really wouldn't hurt to have it in one more place, IMO.

sdras:
Ok, I think I addressed all of the concerns in the review. Let me know if we need further changes :)

sdras:
Sourcemap issue made some traction, getting this recipe in there :)

PR #1466, opened 2018-03-03
Fix a few typos

stakx:
Just a couple of typos that I noticed. I wasn't quite sure whether `master` is the right target branch for this; if not, please let me know.

### DIFF
--- a/src/v2/cookbook/unit-testing-vue-components.md
+++ b/src/v2/cookbook/unit-testing-vue-components.md
@@ -106,7 +106,7 @@ Let's take a look at the component code first:
     <div class="message">
       {{ message }}
     </div>
-    Enter your username: <input v-model="usernane">
+    Enter your username: <input v-model="username">
     <div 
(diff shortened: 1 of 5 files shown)


src/v2/cookbook/unit-testing-vue-components.md
```diff
@@ -106,7 +106,7 @@ Let's take a look at the component code first:
     <div class="message">
       {{ message }}
     </div>
-    Enter your username: <input v-model="usernane">
+    Enter your username: <input v-model="username">
```
stakx: Given that this article talks about testing, I'm not 100 % sure whether this typo is intentional or not. It seems, however, that the tests here are about something else (number of characters in the input user name).

sdras: Yeah, agreed 👍

sdras:
Thank you! This looks great.

PR #1468, opened 2018-03-06
Cookbook: creating a custom directive for scroll events

sdras:
This cookbook example shows how you can coordinate animations triggered on scroll with the use of custom directives, and then also how to optimize further and pass arguments for repetitive code.

### DIFF
--- a/src/v2/cookbook/creating-custom-scroll-directives.md
+++ b/src/v2/cookbook/creating-custom-scroll-directives.md
@@ -6,7 +6,7 @@ order: 5
 
 ## Simple Example
 
-There are many times that we might want to add a bit of behavior, especially animation, to a scroll event on a site. There are many ways to do so, but the path with the least amount of code and dependencies is to use a [custom directive](https://vuejs.org/v2/guide/custom-directive.html) to create hook for anything that fires off a particular scroll event.
+There are many times that we might want to add a bit of behavior, especially animation, to a scroll event on a site. There are many ways to do so, but the path with the least amount of code and dependencies is to use a [custom directive](https://vuejs.org/v2/guide/custom-directive.html) to create a hook for anything that fires off a particular scroll event.
 
 ``` js
 Vue.directive('scroll', {
@@ -42,6 +42,8 @@ new Vue({
   </div>
 </div>
 ```
+
+<p class="tip">Remember! The directive must be registered before the Vue instance</p>
 
 We'd also need a style property that will transition the intermediary values here, in this case: 
 
@@ -92,7 +94,7 @@ Though we would remove the previous CSS transition from this implementation beca
 
 Vue is rich with options for directives, most of which cover very common use-cases, which can create a very productive developer experience. But even if you have an edge case not covered by the framework, it's got you covered in this case as well, because you can quite easily create a custom directive to fit your need.
 
-Attaching and removing scroll events to elements is a really good use case for a custom directive because, just like other directives we use, they are necessarily tied to the element and otherwise, we'd have to find the reference for it in the DOM. This pattern avoids the need for traversal, and keeps the event logic paired with the node that it's in reference to.
+Attaching and removing scroll events to elements is a really good use case for this technique because just like other directives we use, they are necessarily tied to the element and otherwise, we'd have to find the reference for it in the DOM. This pattern avoids the need for traversal, and keeps the event logic paired with the node that it's in reference to.
 
 ## Real-World Example: Using a Custom Scroll Directive for Cascading Animations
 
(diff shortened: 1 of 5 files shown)


src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
+---
+title: Creating Custom Scroll Directives
+type: cookbook
+order: 5
+---
+
+## Simple Example
+
+There are many times that we might want to add a bit of behavior, especially animation, to a scroll event on a site. There are many ways to do so, but the path with the least amount of code and dependencies is to use a [custom directive](https://vuejs.org/v2/guide/custom-directive.html) to create a hook for anything that fires off a particular scroll event.
```
phanan: Maybe `is perhaps to use…` instead? Unless we're 100% sure that a directive is _always_ shorter.

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
...
+
+``` js
+Vue.directive('scroll', {
+  inserted: function(el, binding) {
```
phanan: Call me nick-picky, but there should be a space after `function` following to Vue's coding standard.

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
...
+
+// main app
+new Vue({
+  el: '#app',
+  methods: {
+   handleScroll: function(evt, el) {
+    if (window.scrollY > 50) {
+      el.setAttribute("style", "opacity: 1; transform: translate3d(0, -10px, 0)")
+    }
+    return window.scrollY > 100;
```
phanan: Looks like there's an indent issue here.

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
...
+``` html
+<div id="app">
+  <h1 class="centered">Scroll me</h1>
+  <div class="box" v-scroll="handleScroll">
+    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit. A atque amet harum aut ab veritatis earum porro praesentium ut corporis. Quasi provident dolorem officia iure fugiat, eius mollitia sequi quisquam.</p>
+  </div>
+</div>
+```
+
+<p class="tip">Remember! The directive must be registered before the Vue instance</p>
```
phanan: Maybe finish the second sentence with a period?

src/v2/cookbook/creating-custom-scroll-directives.md
phanan: Oh, and often we don't use semi-colons ;) (pun not intended).

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
...
+  methods: {
+   handleScroll: function(evt, el) {
+    if (window.scrollY > 50) {
+      TweenMax.to(el, 1.5, {
+        y: -10,
+        opacity: 1,
+        ease: Sine.easeOut
+      })
+    }
+    return window.scrollY > 100;
```
phanan: Same indent issue I guess.

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -0,0 +1,152 @@
...
+    }
+  }
+});
+```
+
+Though we would remove the previous CSS transition from this implementation because it's now handled with JavaScript.
+
+## The Benefit of Using Custom Directives
+
+Vue is rich with options for directives, most of which cover very common use-cases, which can create a very productive developer experience. But even if you have an edge case not covered by the framework, it's got you covered in this case as well, because you can quite easily create a custom directive to fit your need.
```
phanan: Shouldn't it be "fit your needs?"

src/v2/cookbook/creating-custom-scroll-directives.md
sdras: Ah good point. I can't think of a time when it wouldn't be shorter but I'm sure there are edge cases out there :)

src/v2/cookbook/creating-custom-scroll-directives.md
sdras: 👍

src/v2/cookbook/creating-custom-scroll-directives.md
sdras: I should just run it through prettier. Too bad the default formatter for codepen is a different style than what we use so I can't use it. Thanks!

src/v2/cookbook/creating-custom-scroll-directives.md
sdras: yes :)

sdras:
Thanks for the review! I updated the PR based on the suggestions. Also included: the ordering of the cookbook was all over the place, so I updated the entries to use solid numbers in the order that the PRs were received to avoid the collision here. I'll start making sections for them in a subsequent PR.

PR #1475, opened 2018-03-09
Delete the line 'npm install'

pranayaryal:
This is already done by the command 'vue init ..' which installs the node_modules/


sdras:
True, this is relatively new with the upgrade. I should actually revisit these docs now, thanks for the reminder.

PR #1481, opened 2018-03-10
Clean up cookbook toc and add it to learn dropdown

sdras:
Once we get the other two PRs in, we'll have almost ten recipes, so that seems like a good time to put the cookbook into beta and link it up. 🎉 Chris and I went back and forth on the sections, but it makes sense to remove them for now because they don't really fall into the pattern we have currently.

I also placed it in the "learn" dropdown, because that seemed to make the most sense- after examples. I'm flexible and open to other paths here.


chrisvfritz:
This looks really good! We can always reorganize into sections later if we find it helpful, since they don't affect the site organization. If you want it to start out in beta, I think we'll also want to add beta tags to a few places. 

Then I think the only other thing I was hoping to do before we launched the cookbook is convert the existing examples into cookbook recipes, since the cookbook is kind of a spiritual successor of the examples section. Then I can set up redirects for those example pages to their new place in the cookbook.

What do you think?

sdras:
I actually think we should have both, and keep them separate. Examples are really useful for a different reason- sometimes you just want to dive into some code and get a feel for it. I think when I was new to Vue, I tested out the examples before I even read the Guide (I read the comparison, played with the examples, and then was sold enough to invest time in reading).

Seems to me the cookbook recipes serve a different purpose- for people who are already invested in Vue who are seeking out either a deeper understanding or for some particular implementation. That's my vote, anyways.

I can add in the beta tags tomorrow :)

chrisvfritz:
Oh, interesting - I hadn't thought about it that way. I tend to avoid standalone examples like that when first learning, because I feel like I lack the context to make an unbiased judgment. My fear with encouraging that trajectory of learning is it's so easy to just see something different and write it off, like many people unfortunately do when they first see templates, directives, and even SFCs without reading the guide for context. You fortunately came with a very open mind, but I think many people don't.

If some do want to learn that way though, I worry that our current examples aren't ideal for it. Perhaps it would be better to provide a single, well-commented playground example that acts as a miniature tour of common features - maybe built off the code from our marketing video. We could then link to it at the beginning of the introduction (rather than hiding it in a menu) to make it more easily discoverable for those who just want to dive into some code. I think a plain ES5 version and a more advanced one with ES2015+ and SFCs would also be good.

What would you think about that in place of our current examples? Do you think it might better serve the same purpose for you? Or are there aspects to that learning path I'm still not grokking? (And if it's too complicated to explain over text, I'm always happy to hop on a video call. 🙂)

sdras:
Oh, that would be an interesting route as well. Sure, let's hop on a call and discuss it sometime this week, this is a bigger discussion and will be too long async on a comment thread.

sdras:
As per our discussion today, moving forward with putting the existing examples in the cookbook and replacing examples section with pared down examples, documented in this issue: #1494. This will happen after we put the cookbook into beta though, so as not to slow it down as a resource.

Issue #1230, opened 2017-10-20
Moving "Components" page into its own category


yyx990803:
We are having more and more content in the component page, which itself is arguably one of the most important concept in the framework. However currently in the sidebar it only has a very low-profile link to the entire page. The page itself is also getting extremely long, making it difficult to navigate.

I think it's time that we break the page into multiple ones and make "Components" a category of its own. In the sidebar, we could have listing like below:

### Components
- Intro
- Props
- Events
- Slots
- Async
- Misc

For the new component composition page, I'm not entirely sure whether it should be included here or in the dedicated Reusability and Composition section. Thoughts?

chrisvfritz:
Agreed - I plan to do this soon. I was thinking of keeping the components page, but changing it to more of an intro with only the basics of slots, props, and events. The reasoning is I still want to keep basic component usage in Essentials. 

Then all the more advanced content would be split between new pages in a new Advanced Components section (name TBD). The new component composition page would then be moved there.

How does that sound to you?

yyx990803:
Sounds good!

sdras:
This has been done for a while :)

PR #1482, opened 2018-03-10
Moving Components Page into its Own Category

sdras:
ref #1230

I've heard from a few people now that this page is too long and therefore a little confusing, so I split it up. I actually didn't the issue or Evan's suggestion for the sidebar in the issue until now, but the sections I chose are actually pretty close. Happy to change things around, though. Just aiming for clarity and ease of use, suggestions welcome.

Feel free to tag more reviewers as well.

---

### REVIEW NOTES FROM CHRIS

As you review, **these are things I'd like to address now**: 

- Typos
- Inconsistencies
- False information
- Vital information that's missing from Components Basics
- Suggestions for page renames
- Anything that's _worse_ than it was before

For the sake of getting this merged in a somewhat timely manner, **I'd prefer reviews did _not_ include ideas for further improvements**. Please create new issues for those wonderful ideas, as we can keep improving this after it's merged. 🙂


yyx990803:
This is awesome, thanks Sarah! I'll give it a deeper look next week.

For anyone else who wants to help review this, the preview deployment can be found at https://deploy-preview-1482--vuejs.netlify.com/v2/guide/components.html

chrisvfritz:
@yyx990803 Just a note here: I'm working with Sarah to polish this a bit more and combine it with some of my own local work, so you can put off reviewing until later this week - I'll tag retag you. 🙂

yyx990803:
@chrisvfritz cool, thanks!

chrisvfritz:
Update: I just finished by rough drafts of all these pages and doing my second pass today. If all goes well, this might be ready for review tonight! Once approved, we'll just have to set up redirects for all the broken links. Not looking forward to that part. 😅

chrisvfritz:
OK, I still need to do one more polish pass for a few of these pages, but it should be worthwhile to review now. :-)

chrisvfritz:
@sdras I can't tag you for review since it's your PR, but I'd also love to see your comments, since I'm the last one to touch each page. 😄

sdras:
Sounds good, I should have some time next week to go through it :)

chrisvfritz:
Yay! I think it'd be really nice if we could have this in, in time for VueConf. 🙂

LinusBorg:
I just read the whole thing in one sitting and didn't spot anything noteworthy - this is realy great work, and a huge and needed improvement to this section.

Awesome work Chris & Sarah!

src/v2/guide/components-slots.md
```diff
@@ -0,0 +1,229 @@
...
+</button>
+```
+
+If the slot is provided content by the parent, it will replace the default content.
+
+## Scoped Slots
+
+> New in 2.1.0+
+
+Sometimes you'll want to provide a component with a reusable slot that can data from the child component. For example, a simple `<todo-list>` component may contain the following in its template:
```
phanan: That can [?] data?

src/v2/guide/components-registration.md
```diff
@@ -0,0 +1,194 @@
...
+)
+
+requireComponent.keys().forEach(fileName => {
+  // Get component config
+  const componentConfig = requireComponent(fileName)
+
+  // Get PascalCase name of component
+  const componentName = upperFirst(
+    camelCase(
+      fileName
```
phanan: We can do both with `fileName.replace(/^\.\/(.*)\.\w+$/, '$1')`

src/v2/guide/components-slots.md
```diff
@@ -0,0 +1,229 @@
...
+  </main>
+  <footer>
+    <p>Here's some contact info</p>
+  </footer>
+</div>
+```
+
+## Default Slot Content
+
+There are cases when it's useful to provide a slot with default content. For example, a `<submit-button>` component might want the content of the button to be "Submit" by default, by also allow users to override with "Save", "Upload", or anything else.
```
phanan: "but also".

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,311 @@
...
+```js
+methods: {
+  // Used to focus the input from the parent
+  focus: function () {
+    this.$refs.input.focus()
+  }
+}
+```
+
+Thus allowing the parent component to focus the input inside `<base-input>` with:
```
phanan: Shouldn't it be "focus into" or something like that? I'm not very sure about this one.

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,311 @@
...
+
+```js
+this.$refs.usernameInput.focus()
+```
+
+When `ref` is used together with `v-for`, the ref you get will be an array containing the child components mirroring the data source.
+
+<p class="tip"><code>$refs</code> are only populated after the component has been rendered, and it is not reactive. It is only meant as an escape hatch for direct child manipulation - you should avoid accessing <code>$refs</code> from within templates or computed properties.</p>
```
phanan: Either "is only populated" or "they are not reactive."

src/v2/guide/components-props.md
```diff
@@ -0,0 +1,314 @@
...
+
+```js
+Vue.component('blog-post', {
+  props: {
+    author: Person
+  }
+})
+```
+
+To validate that the value of the `author` prop was created with `new Person`.
```
phanan: Extreme nitpicking mode activated: it should be a lowercase "t."

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,311 @@
...
+this.$root.foo = 2
+
+// Access root computed properties
+this.$root.bar
+
+// Call root methods
+this.$root.baz()
+```
+
+<p class="tip">This can be convenient for demos or very small apps with a handful of components. However, the pattern does not scale well to medium or large-scale applications, so we strongly recommend using <a href="https://github.com/vuejs/vuex">https://github.com/vuejs/vuex</a> to manage state in most cases.</p>
```
phanan: I'd suggest this instead: `<a href="https://github.com/vuejs/vuex">vuex</a>`.

src/v2/guide/components-slots.md
chrisvfritz: Fixing. 🙂

src/v2/guide/components-registration.md
chrisvfritz: I did intentionally split it up so that I wasn't doing too much in one line, but this actually reads pretty well to me. 👍

src/v2/guide/components-edge-cases.md
chrisvfritz: I think "focus the input", "focus into the input", and "set focus on the input" are all used pretty commonly, so I just used the shortest one. 😄

src/v2/guide/components-props.md
chrisvfritz: Extreme nitpicking mode very welcome. 🙂 Thanks!

src/v2/guide/components-edge-cases.md
chrisvfritz: Yeah, this was a mistake. 😅

chrisvfritz:
Thanks @LinusBorg and @phanan! I've addressed all the great comments. 🙂

src/v2/guide/components-slots.md
```diff
@@ -0,0 +1,229 @@
...
+<navigation-link url="/profile">
+  <!-- Add an icon with a component -->
+  <font-awesome-icon name="user"></font-awesome-icon>
+  Your Profile
+</navigation-link>
+```
+
+If `<navigation-link>` did **not** contain a `<slot>` element, any content passed to it would simply be discarded.
+
+## Compilation Scope
```
sdras: Reading through, the flow of this seemed out of order to me. I think it would make more sense to address named slots and default slots first, and then this section closer to scoped slots. What do you think?

src/v2/guide/components-dynamic-async.md
```diff
@@ -0,0 +1,101 @@
...
+<component v-bind:is="currentTabComponent"></component>
+```
+
+When switching between these components though, you'll sometimes want to maintain their state or avoid re-rendering for performance reasons. In these cases, you can wrap a dynamic component with a `<keep-alive>` element:
+
+``` html
+<!-- Inactive components will be cached! -->
+<keep-alive>
+  <component v-bind:is="currentTabComponent"></component>
+</keep-alive>
```
sdras: This, and the API don't fully show off the benefits of `keep-alive`, IMO, but that's out of scope of this PR, so I've filed an [issue for myself](https://github.com/vuejs/vuejs.org/issues/1493) to take care of it when this is merged.

src/v2/guide/components.md
```diff
@@ -1,1481 +1,595 @@
...
-  el: '#some-element',
-  // options
+// Define a new component called button-counter
+Vue.component('button-counter', {
+  data: function () {
+    return {
+      count: 0
+    }
+  },
+  template: '<button v-on:click="count++">You clicked me {{ count }} times.</button>'
```
sdras: This was a nice addition to make this example more clear.

src/v2/guide/components.md
```diff
@@ -1,1481 +1,595 @@
...
+}
+.dynamic-component-demo-tab {
+  border: 1px solid #ccc;
+  padding: 10px;
+}
+</style>
 {% endraw %}
 
-In this example, it's important to note that the child component is still completely decoupled from what happens outside of it. All it does is report information about its own activity, just in case a parent component might care.
+The above is made possible by Vue's `<component>` element with the `is` special attribute:
```
sdras: This seems to be a better way of organizing this, but we miss out on the good example that was provided before, where we show that `:is` can be really important if you need to preserve tags. We don't necessarily need to use the same example, but this becomes a bit of a shallow definition of what `is` does, IMO.

src/v2/guide/components-props.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+Will be equivalent to:
+
+``` html
+<blog-post
+  v-bind:id="post.id"
+  v-bind:title="post.title"
+></blog-post>
+```
```
sdras: The syntax highlighting looks strange to me, here, is anyone else [seeing that](https://deploy-preview-1482--vuejs.netlify.com/v2/guide/components.html#Passing-Data-to-Child-Components-with-Props)?
screen shot 2018-03-17 at 9 32 16 am

src/v2/guide/components-registration.md
```diff
@@ -0,0 +1,191 @@
...
+```js
+var ComponentA = { /* ... */ }
+
+var ComponentB = {
+  components: {
+    'component-a': ComponentA
+  },
+  // ...
+}
+```
```
sdras: Do we also want to mention here how you can do just:
```
var ComponentB = {
  components: {
    ComponentA
  },
  // ...
}
```

Might be nice to show people.

src/v2/api/index.md
```diff
@@ -1426,7 +1426,7 @@ type: api
 
 - **Details:**
 
-  Contains parent-scope `v-on` event listeners (without `.native` modifiers). This can be passed down to an inner component via `v-on="$listeners"` - useful when creating higher-order components.
+  Contains parent-scope `v-on` event listeners (without `.native` modifiers). This can be passed down to an inner component via `v-on="$listeners"` - useful when creating transparent wrapper components.
```
sdras: I think we've talked about this before, are we all ok with transparent wrapper instead of higher-order? I think higher-order is more commonly used so it's possible it will be understood much faster by people learning, but I could go both ways on it.

sdras:
All in all, this looks good! I love the organization of the base structure and then deep-dive, really great. I've added some comments here, all pretty minor.

src/v2/guide/components-dynamic-async.md
chrisvfritz: I agree and think that's a great idea!

src/v2/guide/components-slots.md
chrisvfritz: I like that! I just moved that section down. 🙂

src/v2/guide/components.md
chrisvfritz: I'm not sure I completely understand what you mean, but I'll be in a video call with you soon so can probably get clarification then. 🙂

src/v2/guide/components-props.md
chrisvfritz: That's because I marked the syntax as JS instead of HTML. 😅 Fixed.

src/v2/api/index.md
chrisvfritz: Isn't this different from an HOC? The definition [used by React](https://reactjs.org/docs/higher-order-components.html) is:

> a function that takes a component and returns a new component

This is a component rather than a function and it returns raw elements rather than a new component.

src/v2/guide/components-registration.md
chrisvfritz: Agreed! Added. 🙂

src/v2/guide/components.md
sdras: Sounds good, I'll find the line you deleted (which is in another file) that I'm referring to as well, so that what I'm referring to missing is more clear. This PR is a little funny to comment on because some of the changes are in two completely different places :)

src/v2/guide/components-slots.md
sdras: Thanks!

src/v2/api/index.md
sdras: Not necessarily, from the same document just a paragraph lower:

> a higher-order component transforms a component into another component.

Having worked with both, I'm not sure I'd make such a distinction, but if you think that it's helpful for understanding, I'm all for it.

src/v2/guide/components.md
```diff
@@ -1,1481 +1,595 @@
...
+It's common for an app to be organized into a tree of nested components:
 
-``` html
-<table>
-  <my-row>...</my-row>
-</table>
-```
+![Component Tree](/images/components.png)
 
-The custom component `<my-row>` will be hoisted out as invalid content, thus causing errors in the eventual rendered output. A workaround is to use the `is` special attribute:
```
sdras: This, with the lines above it up to line 98, is what I was referring to- I'm adding this comment in here for quick reference as a talking point :)

src/v2/guide/components.md
chrisvfritz: Ah, so if I'm understanding correctly, then this line:

> The custom component `<my-row>` will be hoisted out as invalid content, thus causing errors in the eventual rendered output. A workaround is to use the `is` special attribute:

Was not actually deleted but rewritten and moved around, so it's now:

> The custom component `<todo-item>` will be hoisted out as invalid content, causing errors in the eventual rendered output. Fortunately, the `is` special attribute offers a workaround:

It can definitely get confusing looking at the diff. 😅 I have that section a little lower, in the subsection about `DOM Template Parsing Caveats`. Do you feel like that's a good place for it? Is there any other important information you think I might have missed?

src/v2/guide/components-dynamic-async.md
chrisvfritz: Couldn't help myself and added it. 😅 Your idea was too good!

src/v2/guide/components.md
sdras: Ah, ok there it is- so many moving parts, I missed where it went to. Thanks!

src/v2/api/index.md
sdras: Sounds good!

sdras:
I can't approve my own PR but I just looked through all the changes, and it looks good to me 👍

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+title: Custom Events
+type: guide
+order: 103
+---
+
+> This page assumes you've already read the [Components Basics](components.html). Read that first if you are new to components.
+
+## Event Names
+
+Unlike components and props, event names don't have provide any automatic case transformation. Instead, the name of an emitted event must exactly match the name used to listen to that event. For example, if emitting a camelCased event name:
```
znck: > event names don't ~have~ provide any automatic case 

`have` should be dropped here.

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+this.$emit('myEvent')
+```
+
+Listening to the kebab-cased version will have no effect:
+
+```html
+<my-component v-on:my-event="doSomething"></my-component>
+```
+
+The reason for this is that unlike components and props, event names will never be used as variable or property names in JavaScript, so there's no reason to use camelCase or PascalCase. Additionally, `v-on` event listeners inside DOM templates will be automatically transformed to lowercase (due to HTML's case-insensitivity), so `v-on:myEvent` would become `v-on:myevent` -- making `myEvent` impossible to listen to.
```
znck: > ~The reason for this is that u~Unlike components and props, event names will never be used as variable or property names in JavaScript, so there's no reason to use camelCase or PascalCase.

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+
+For these reasons, we recommend you **always use kebab-case for event names**.
+
+## Customizing Component `v-model`
+
+> New in 2.2.0+
+
+By default, `v-model` on a component uses `value` as the prop and `input` as the event, but some input types such as checkboxes and radio buttons may want to use the `value` attribute for a [different purpose](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Value). Using the `model` option can avoid a conflict in such cases:
```
znck: > can avoid ~a~ conflict in such cases

The indefinite article "a" is unnecessary.

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+  {{ label }}
+  <input
+    v-bind="$attrs"
+    v-bind:value="value"
+    v-on:input="$emit('input', $event.target.value)"
+  >
+</label>
+```
+
+In that case, the `.native` listener in the parent would silently break. There would be no errors, but the `onFocus` handler wouldn't be called when we expected it to.
```
znck: > There would be no errors, but the `onFocus` handler wouldn't be called when we expected it to.

Can we simplify this statement?

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+})
+```
+
+Now the `<base-input>` component is a **fully transparent wrapper**, meaning it can be used exactly like a normal `<input>` element: all the same attributes and listeners will work.
+
+## `.sync` Modifier
+
+> New in 2.3.0+
+
+In some cases we may need "two-way binding" for a prop. Unfortunately, true two-way binding can create maintenance issues, because child components can mutate the parent without the source of that mutation being obvious in both the parent and the child.
```
znck: > In some cases, we

A missing coma.

src/v2/guide/components-custom-events.md
```diff
@@ -0,0 +1,158 @@
...
+<text-document v-bind:title.sync="doc.title"></text-document>
+```
+
+The `.sync` modifier can also be used with `v-bind` when using an object to set multiple props at once:
+
+```html
+<text-document v-bind.sync="doc"></text-document>
+```
+
+This has the effect of adding `v-on` update listeners for not only `title`, but also any other properties on the `doc` object.
```
znck: Instead of `This`, we can repeat the context.

> The `.sync` modifier when used with `v-bind` has the effect of ...

src/v2/guide/components-dynamic-async.md
```diff
@@ -0,0 +1,273 @@
...
+.dynamic-component-demo-post > :first-child {
+  margin-top: 0 !important;
+  padding-top: 0 !important;
+}
+</style>
+{% endraw %}
+
+You'll notice that if you select a post, switch to the _Archive_ tab, then switch back to _Posts_, it's no longer showing the post you selected. That's because each time you switch to a new tab, Vue creates a new instance of the `currentTabComponent`.
+
+This is normally useful behavior, but in this case, we'd really like those tab component instances to be cached once they're created for the first time. To solve this problem, we can wrap our dynamic component with a `<keep-alive>` element:
```
znck: It's better to restate the context, instead of using this.

> Not caching component instances is normally useful behaviour, but in above example, we'd like the tab component instances to be cached ...

src/v2/guide/components-dynamic-async.md
```diff
@@ -0,0 +1,273 @@
...
+
+Now the _Posts_ tab maintains its state (the selected post) even when it's not rendered. See [this fiddle](https://jsfiddle.net/chrisvfritz/Lp20op9o/) for the complete code.
+
+<p class="tip">Note that `<keep-alive>` requires the components being switched between to all have names, either using the `name` option on a component, or through local/global registration.</p>
+
+Check out more details on `<keep-alive>` in the [API reference](../api/#keep-alive).
+
+## Async Components
+
+In large applications, we may need to divide the app into smaller chunks and only load a component from the server when it's actually needed. To make that easier, Vue allows you to define your component as a factory function that asynchronously resolves your component definition. Vue will only trigger the factory function when the component actually needs to be rendered and will cache the result for future re-renders. For example:
```
znck: We can remove `actually` word from this paragraph as the meaning of the statement remains same.

> ...only load a component from the server when it's ~actually~ needed.
> ...when the component ~actually~ needs to be rendered and will cache the result...

src/v2/guide/components-custom-events.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-custom-events.md
chrisvfritz: I like it! Updated. 🙂

src/v2/guide/components-custom-events.md
chrisvfritz: I'm going to leave this in actually, because while many dialects of English would allow omitting the "a", I think it would sound strange in American English.

src/v2/guide/components-custom-events.md
chrisvfritz: Hm, nothing is coming to mind, but I'm definitely open to suggestions!

src/v2/guide/components-custom-events.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-custom-events.md
chrisvfritz: Hm, this seems redundant to me since we just established the context in the previous sentence. @sdras What do you think?

src/v2/guide/components-dynamic-async.md
chrisvfritz: I agree with you in this case. 🙂 Updated.

src/v2/guide/components-dynamic-async.md
chrisvfritz: I like it! Removed. 🙂

src/v2/guide/components-custom-events.md
znck: Does code snippets between paragraphs break context? 

For large code snippets, it happens. I tend to go back to last line before continuing to next paragraph. For one line snippet, I am not sure. Moreover, the amount code breaking reading context may vary from person to person.

src/v2/guide/components-custom-events.md
phanan: About to say the same. How is "a" unnecessary?

src/v2/guide/components-custom-events.md
phanan: I prefer "This" here. One line of code in between is not enough to repeat the statement.

src/v2/guide/components-custom-events.md
chrisvfritz: @znck can correct me if I'm wrong, but I think in Indian English it's more common, and even in American English there will be no article in some contexts. For example, "there was conflict between two countries."

src/v2/guide/components-custom-events.md
chrisvfritz: @znck Yeah, I agree with the principle in general, but in this particular case think it will be better for most people to leave it out.

src/v2/guide/components-custom-events.md
sdras: Yeah, in American English, omitting the a in this case would sound strange.

src/v2/guide/components-custom-events.md
sdras: Yeah, I read it through and it seemed redundant to change it, it's pretty clear to me what this refers to. But that's a good thing to look out for!

src/v2/guide/components-custom-events.md
phanan: I often (read: almost always) rely on Grammarly for this, and it seems to dislike omitting the article in @chrisvfritz's example:

image

chrisvfritz: @phanan 👍 for Grammarly - great tool! I do disagree with it sometimes though. 😄 I think this is a phrase people use on the news quite often.

chrisvfritz:
It looks like it might not be possible to alias hash links (e.g. `/components#Base-Example` with Hexo, but I just set up some client-side redirects in 1d90cba. This way, people clicking on an old link including a hash will be automatically forwarded to the correct section of the new documentation.

src/v2/guide/components.md
```diff
@@ -1,1481 +1,597 @@
...
+At the very least, you'll want to include the post's content:
 
-### Passing Data with Props
+```html
+<h3>{{ post.title }}</h3>
+<div v-html="post.content"></div>
+```
 
-Every component instance has its own **isolated scope**. This means you cannot (and should not) directly reference parent data in a child component's template. Data can be passed down to child components using **props**.
+If you try this in your template however, Vue will show an error, explaining that **every component must have a single root element**. Vue enforces this limitation to enable some interesting features, which you'll learn about later.
```
yyx990803: This is an actual technical limitation during the implementation of Vue 2, and we may eventually support fragments as component template root in the future, so I'd avoid phrasing it as if we did it to enable specific features.

src/v2/guide/components.md
```diff
@@ -1,1481 +1,597 @@
...
-Here's an example:
+This will lead to issues when using components with elements that have such restrictions. For example:
 
 ``` html
-<div id="counter-event-example">
-  <p>{{ total }}</p>
-  <button-counter v-on:increment="incrementTotal"></button-counter>
-  <button-counter v-on:increment="incrementTotal"></button-counter>
-</div>
+<ul>
```
yyx990803: This example isn't entirely correct - per the spec `<ul>` does have restriction on what elements can go inside, but most browser don't actually enforce it with the hoisting behavior. `<table>` and `<select>`, on the other hand, do enforce it with hoisting.

src/v2/guide/components-registration.md
```diff
@@ -0,0 +1,204 @@
...
+```js
+import ComponentA from './ComponentA.vue'
+
+export default {
+  components: {
+    ComponentA
+  },
+  // ...
+}
+```
```
yyx990803: I think we should move the tip explaining the ES2015+ shorthand up here, since this is the first time the usage has appeared. In addition,  we should also explain the naming resolution rules - i.e. the fact that you can use it in the template both as `component-a` and `ComponentA`.

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,319 @@
...
+``` html
+<my-component inline-template>
+  <div>
+    <p>These are compiled as the component's own template.</p>
+    <p>Not parent's transclusion content.</p>
+  </div>
+</my-component>
+```
+
+<p class="tip">However, <code>inline-template</code> makes the scope of your templates harder to reason about. As a best practice, prefer defining templates inside the component using the <code>template<code> option or in a <code>&lt;template&gt;<code> element in a <code>.vue<code> file.</p>
```
yyx990803: Missing slash in ending `</code>` tags here

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,319 @@
...
+Vue.component('hello-world', {
+  template: '#hello-world-template'
+})
+```
+
+<p class="tip">These can be useful for demos with large templates or in extremely small applications, but should otherwise be avoided, because they separate templates from the rest of the component definition.</p>
+
+## Controlling Updates
+
+Thanks to Vue's Reactivity system, always knows when it needs to update (if you use it correctly). There are edge cases, however, when you might want to force an update, despite the fact that no reactive data has changed. Then there are other cases when you might want to prevent unnecessary updates.
```
yyx990803: "Thanks to Vue's Reactivity system, [it] always knows when it needs to update"

src/v2/guide/components-edge-cases.md
```diff
@@ -0,0 +1,319 @@
...
+
+### Forcing an Update
+
+<p class="tip">If you find yourself needing to force in update in Vue, in 99.99% of cases, you've made a mistake somewhere.</p>
```
yyx990803: force "an" update

src/v2/guide/components.md
chrisvfritz: Fixed. 👍

src/v2/guide/components.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-edge-cases.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-edge-cases.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-edge-cases.md
chrisvfritz: Fixed. 👍

src/v2/guide/components-registration.md
chrisvfritz: Fixed and fixed. 👍

src/v2/guide/components-edge-cases.md
sdras: If this is fixed, it's not showing up for me, just FYI (maybe not pushed yet?) I thought I would call it out just in case

sdras:
🎉

PR #1486, opened 2018-03-13
add link to MDN event reference in v-on

JGJP:
May be outside the scope of these docs, `v-on:change` just struck me as a common use case. The link to MDN could be useful for those that are not so familiar with how many DOM events there really are.


chrisvfritz:
I agree that listing non-Vue specific events is probably out of scope, but I do like the idea of linking to the MDN doc listing those events! I just tweaked this to make "native DOM events" a clickable link. 🙂

JGJP:
That makes sense, thanks!

PR #1487, opened 2018-03-14
Include .fade-enter-to and .fade-leave class in css example

conanak99:
In the original documentation, we implicitly assume that element got `opacity:1`.
I think we should explicitly include the `opacity:1` on fade-enter-to and fade-leave class. This will make the transition from `.fade-enter -> .fade-enter-to` and `.fade-leave -> .fade-leave-to` easier to understand

### DIFF
--- a/src/v2/guide/transitions.md
+++ b/src/v2/guide/transitions.md
@@ -53,6 +53,9 @@ new Vue({
 .fade-enter, .fade-leave-to /* .fade-leave-active below version 2.1.8 */ {
   opacity: 0;
 }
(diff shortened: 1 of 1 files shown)


sdras:
Thanks for the PR! I don't think we should add this, though. Aside from being redundant because it's the default, it's actually an antipattern and a bit dangerous. If at some point someone decides to change the opacity of the object itself to 0.9 or 0.8, you'll get a flash of the opacity:1 when it starts to change, which will appear broken. If you can think of something else in a description that could help clarify what is not clear to you, that would be better.

Issue #1087, opened 2017-08-19
Separate search page


andreasvirkus:
Greetings!

I'd like to see a separate search page so we could add Vue.js' docs as a bang to duckduckgo.
The duck's bang requires a bang url, e.g. `https://vuejs.org/?q={{{s}}}`.

That way Vue's docs would be easily searchable with (for example) `!vue v-once`.

I could make a PR for this as well, would just like to know how you'd like this implemented
(I see we use Algolia's docsearch as the autocomplete wrapper).

chrisvfritz:
I'd accept a PR for a dedicated search page. How about at `/v2/search`?

chrisvfritz:
I'm going to close this since there doesn't seem to be much interest. Happy to reopen in the future. 🙂

NoelDeMartin:
@andreasvirkus @chrisvfritz Hey just letting you know that duck duck go accepted the bang! so you can now search using it, for example "v-model !vue" and it'll take you to the vue docs :).

PR #1488, opened 2018-03-14
Implement search page

NoelDeMartin:
Implementation for issue #1087. The url would be `https://vuejs.org/v2/search/?q={{search query}}`.

I have never worked with algolia nor hexo before, so please check that I didn't miss anything. I think the styles could also be improved. And I tried to follow the coding style the best I could (like not using semicolon), but I may have missed something.

In any case, I'll be happy to fix anything that you consider so please let me know what you think :D.


chrisvfritz:
@phanan Would you mind reviewing this? I think this review requires more design expertise and that's definitely an area where you excel far beyond me. 😄

themes/vue/layout/search-page.ejs
```diff
@@ -0,0 +1,160 @@
+<div id="search-page">
+    <form class="search-form">
+        <input
+            class="search-query"
+            v-model="search"
+            placeholder="Search VueJS"
```
phanan: I'd suggest just "Search" or "Search Vue.js" here.

themes/vue/layout/search-page.ejs
```diff
@@ -0,0 +1,160 @@
...
+        ></search-result>
+    </template>
+
+    <p v-else>No results were found.</p>
+
+    <div ref="infiniteScrollAnchor"></div>
+
+</div>
+
+<script src="//cdn.jsdelivr.net/algoliasearch/3/algoliasearch.min.js"></script>
```
phanan: These should always be `https://` regardless of the original site protocol.

themes/vue/layout/search-page.ejs
```diff
@@ -0,0 +1,160 @@
...
+<script src="//cdn.jsdelivr.net/algoliasearch.helper/2/algoliasearch.helper.min.js"></script>
+<script src="//polyfill.io/v2/polyfill.min.js?features=IntersectionObserver"></script>
+<script>
+var match = window.location.pathname.match(/^\/(v\d+)/)
+var version = match ? match[1] : 'v2'
+var algolia = algoliasearch('BH4D9OD16A', '85cc3221c9f23bfbaa4e3913dd7625ea')
+var algoliaHelper = algoliasearchHelper(algolia, 'vuejs', {
+    hitsPerPage: 15,
```
phanan: We use 2-space indentation (both JS and HTML) for our docs, can you please check?

themes/vue/layout/search-page.ejs
```diff
@@ -0,0 +1,160 @@
...
+        var observer = new IntersectionObserver(function(entries) {
+            if (entries[0].isIntersecting && searchPage.totalPages > searchPage.lastPage + 1) {
+                searchPage.addPage()
+            }
+        })
+        observer.observe(this.$refs.infiniteScrollAnchor)
+    },
+    methods: {
+        addPage() {
+            algoliaHelper.setCurrentPage(this.lastPage+1).search()
```
phanan: Space around `+`

NoelDeMartin:
Ok I have applied the changes you mentioned, to answer your questions:

- DuckDuckGo is a search engine with a feature called "bangs", which allows you to search directly in other websites, and since it can be configured as a default search engine for the browser, makes it very easy to search for documentation. What we want to achieve is that you would write something like "v-model !vue" in your navigation bar and it would take you directly to this search results page, without needing to visit the search engine. More info about that here: [https://duckduckgo.com/bang](https://duckduckgo.com/bang). About this being an orphan page, I see some value here regardless of using duck duck go or not, because you can get more results in this page (it's implemented with infinite scroll), I don't know how easy it would be to add a "show more results" button on the instant search on top linking to this page. If you think it makes sense I could look into implementing that as well.
- I based my implementation on this: [https://symfony.com/search](https://symfony.com/search) and it behaves as you say. Personally I don't like it because it clutters browsing history and back arrow behaviour. But I guess it's a personal preference, so I don't know, whatever you prefer.
- I don't know which menu should be here, since this searches everywhere (Guide, API, etc.). Sponsor info I'm not sure where to find it in other pages (other than the ones dedicated to that), if you tell me where to find an example I could add it.

And about the https links, I did it like the search on the top is implemented now, so maybe you want to update it as well:
```
<!-- search -->
<link href="//cdn.jsdelivr.net/docsearch.js/1/docsearch.min.css" rel='stylesheet' type='text/css'>
<%- css('css/search') %>
<script src="//cdn.jsdelivr.net/docsearch.js/1/docsearch.min.js"></script>
```

phanan:
* @chrisvfritz what do you think about the page being orphan? I don't have a background about this, so I can't tell
* Re: URL, IMO it should behave like the Symfony one
* The sponsor info is in `partials/sponsors_sidebar.ejs`, which is included in `sidebar.ejs`, which in turn is automatically included if your `index.md` has a `type` attribute. Simply adding an arbitrary `type` won't work that well though, as such will render the whole menu tree to the left as well (which you don't want), so you'll need to play around a bit with the logic.

chrisvfritz:
I think it being an orphan is fine. If people want to search while already on the site, there's already a search bar available at the top, so people might only use a dedicated search page when being directed from other services like DuckDuckGo.

NoelDeMartin:
Ok I have added the push history behaviour and sponsors information. I have adblock installed and seems like sponsors are treated as ads, because I wasn't seeing them anywhere. That explains why I couldn't see them anywhere other than the sponsors page.

phanan:
You can (and should) totally disable adblock for the site to test it.

2018-03-19 9:13 GMT+00:00 Noel De Martin <notifications@github.com>:

NoelDeMartin:
Yeah of course, I already did, I didn't think about it before but now that I noticed I have it whitelisted.

phanan:
LGTM now, thanks a lot! If we could have arrow-key navigation for the results, it would be even better, but I believe this is good enough for now.

themes/vue/source/css/_search-page.styl
```diff
@@ -0,0 +1,24 @@
...
+      font-size: 17.55px
+    .summary
+      padding: 0
+      margin: 0
+    .breadcrumb
+      color: $light
+      & + .breadcrumb::before
+        content: "\203A\A0"
+        margin-left: 5px
+        color: $light
```
phanan: Oh, we need an EOL here btw.

NoelDeMartin:
Cool, thanks :D. Yeah I think arrow navigation and adding a "view more results" link in top bar search would be good improvements in the future.

dear-lizhihua:
nice page!

chrisvfritz:
One quick note: if I press Enter in the search input, it currently refreshes the page and clears the search. Could we instead have it match the behavior of the top search bar, where Enter takes you to the first result?

NoelDeMartin:
@chrisvfritz Yeah that's a good idea, and easy to implement. You can test in on the last commit.

chrisvfritz:
Fantastic! A few other notes:

* I noticed that this page uses a layout that assumes a sidebar, making it look a little strange at many screen widths because it's not quite centered. We might not need to use a completely different layout to fix this though. We have a media query to fall back to a centered layout (I think at 900px), so we could possibly use those styles for every screen width for this page.
* Perhaps a nitpick, but "Search Vue.js" sounds a little ambiguous to me, like it might be searching the source code rather than this site. Maybe we should change it to just "Search" or "Search this site"?

Akryum:
The page is missing the 'search by Algolia' logo which is mandatory.

Akryum:
Also: preview url: https://deploy-preview-1488--vuejs.netlify.com/v2/search/

NoelDeMartin:
@chrisvfritz Yeah, the sidebar layout looks weird, but I needed to add it because the sponsors section is located inside of the sidebar. The thing is that it has a fixed position, so it could technically be outside of the sidebar, but on mobile it is inside of the menu so I'm not sure how to go about it, I guess someone who's worked more with the site can have a better opinion. I also removed what appears by default in the sidebar, because it was a menu with only one section so I thought it'd be better to show nothing instead.

I can change the text on the search box once it's decided 100%, because I already changed it once. I still vote for using "Search VueJS" which was my initial proposal, but I don't mind which one you want to put either way.

@Akryum I didn't know it was mandatory, I'll do it in then.

NoelDeMartin:
@chrisvfritz Is there anything else needed for this to be merged?

chrisvfritz:
Nope, thanks for the ping. This just fell off my radar. 🙂

PR #1500, opened 2018-03-19
Update index.md

hjb912:
update the original example, because there is no effect.


Jinjiang:
It's not a working demo but just a piece of code snippet to explain how to register a component. The whole working demo is just `#app-7` below there. So it doesn't need to be modified imo.
Thanks.

chrisvfritz:
Yes, I agree with @Jinjiang. This is partway through to building `app-7` rather than its own demo. Thanks for the PR though! It's good for us to double check these things.

PR #1507, opened 2018-03-19
Order of the console.log statements

gjdipietro:
Corrected the order that the console.log statements would run.

``` js

Vue.prototype.appName = 'My App'

new Vue({
  data: {
    appName: 'The name of some other app'
  },
  beforeCreate: function () {
    console.log(this.appName)
  },
  created: function () {
    console.log(this.appName)
  }
})
```

Will print first "My App" then "The name of some other app".


sdras:
Just verified 👍

PR #1511, opened 2018-03-20
Remove blockquotes from "Guidelines-for-Recipes"

jmsfwk:


sdras:
Thanks for the PR! We added the blockquotes to call attention to this section, we're going to keep this formatting.

jmsfwk:
In that case I would recommend removing the list inside the blockquotes.

I can make a pull request for that tomorrow if you are interested.

Issue #1510, opened 2018-03-20
[cookbook] $route/query change handling


donnysim:
I think it would be great to have a section for handling $route and/or query changes using vue-router as there are many forum and articles about it, but haven't found any [answers] that touch real case project problems.

For example:
`<router-view></router-view>` vs `<router-view :key="$route.fullPath"></router-view>` vs `<router-view :key="$route.path"></router-view>`. What they solve and what problems they introduce - e.g. performance, losing control over partial page updates etc..

I can think of couple example scenarios that could use some guidance:

- a query change does not mean that the page should reload all components, the user only switched the tab in a component (with query to be able to link to the tab), in addition it should not cause any components in page to reload their data;
- a `page` query param should reset to value `1` when a search filter is changed (can easily lead to infinite loops when watching `$route` and then resetting the `page` query param);
- there might be multiple components on the page that each depend on unique query params, but on change, only one should reload it's data, not both;
- advanced `back` button handling, e.g. an edit page with search uses `push` to add query changes to browser history, but the back button should go back to the previous page that brought the user to the edit page, not the last search query or the resources index page;

These are couple of example that would bring immense value if documented properly, yet each of them would require some layer around router and query handling. 

While I did propose this, I'm not confident in my abilities to solve this in a simple and understandable way 😅

sdras:
This sounds great! I think it would be really helpful.

> While I did propose this, I'm not confident in my abilities to solve this in a simple and understandable way

If you can try your best to follow the format, I can help you shape the language and explanations. :) Teamwork!

PR #1519, opened 2018-03-21
[Cookbook] Vue route change handling

donnysim:
@sdras in regards to #1510 

Well, english is not my primary and I never been good at these things but here goes the draft, the more you write, the more you start to doubt yourself if this was a good idea 😨

### DIFF
--- a/src/v2/cookbook/vue-route-change-handling.md
+++ b/src/v2/cookbook/vue-route-change-handling.md
@@ -373,4 +373,4 @@ In this instance, if the user got here through a bookmark or new tab, then there
 This should provide you with some insights and ideas on how something like this could be solved.
 
 ## Summary
-All solutions come back to the simple term `it depends`. Every application has different needs, different structure and requirements, for example maybe the route is being tracked in VueX; maybe component data is loaded/resolved before the page transition happens... there are endless factors that determine what tools and approach we use to achieve our goal and they should be chosen depending on those factors.
+All solutions come back to the simple term `it depends`. Every application has different needs, different structure and requirements, for example maybe the route is being tracked in Vuex; maybe component data is loaded/resolved before the page transition happens... there are endless factors that determine what tools and approach we use to achieve our goal and they should be chosen depending on those factors.


sdras:
You are way too hard on yourself! This is great! I'm slammed getting prepped for VueConf US next week, but I plan to dig in and give you a thorough review the week after. Thanks so much for your time putting this together, it's going to be a great resource.

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+    loadData() { /* logic to load data */ },
+  },
+});
+```
+
+```html
+<router-view :key="$route.fullPath"></router-view>
+```
+
+This will assure that when going from `/orders` to `/orders?page=2` or from `/orders?page=2` to `/orders?page=1`, a fresh instance of `OrderIndex` component will be provided, which will trigger the `created` method and download fresh data each time. There's also a subtle alternative to it:
```
posva: will be provided -> is created, triggering the `created` hook and downloading fresh data.

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+```html
+<router-view :key="$route.path"></router-view>
+```
+
+The main difference is that `/orders` and `/orders?page=2` would not re-create the components and data would not be reloaded, but going from `/orders` to `/invoices` with the same `OrderIndex` component would.
```
posva: I think this can be rephrased by saying _changes on the query eg: going from `/orders` to `/orders?page=2`, won't make up a different `key` and therefore, no data is fetched. However, going to a different route will keep the same behaviour.

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+The `:key` solution isn't widely known so the most commonly used way is the plain `router-view`:
```
posva: Here, I will rather say, that sometimes we don't want to recreate the view component so we go with a key-less router-view and `$watch` the `$route`

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+      handler: 'loadData',
+    },
+  },
+  methods: {
+    loadData() { /* logic to load data */ },
+  },
+});
+```
+
+<p class="tip">Note! While watching `$route` and using `:key="$route.fullPath"` the watcher will never fire unless the component is in another `router-view` scope without `key`.</p>
```
posva: I don't get this sentence

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+Now each time the route changes, components will reload their data.
```
posva: ~components~ `OrderIndex` reloads its data

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+<p class="tip">Remember! Don't modify `$route.query` directly - it will not be persisted to the url and components will not reload, use `$router.push` or `$router.replace` instead.</p>
```
posva: ~persisted~ -> it does not change the url, use `$router.push`...

donnysim:
Thank you for taking a look at this! My main concern right now is what to do with the gist and History code snippets - they are on the bigger side and maybe the gist is not appropriate? Maybe it could be a codesandbox and users copy the code from there if needed?

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+<p class="tip">Remember! Don't modify `$route.query` directly - it does not change the url, use `$router.push` or `$router.replace` instead.</p>
+
+## Simple, but not good enough
+While both `:key` methods work, there are some drawbacks to each one. Lets start with:
+
+```html
+<router-view :key="$route.fullPath"></router-view>
+```
+
+The `$route.fullPath` returns the full url including the query e.g. `/orders?page=2&filter=search`, so what happens when you visit from `/orders` to `/orders?page=1`? The page is recreated with new instances of components and the data is reloaded, yet the route is technically the same as before. At this point the only possible workaround would be to cache the results using something like [VueX](https://vuex.vuejs.org) or plain object. We could use `$route.path` as a `key`, but this will require introducing a watcher and handling query changes which sometimes can be difficult and unexpected side effects if implemented incorrectly. There are more examples why `:key` on the `router-view` could introduce more reloads than necessary which we'll cover further.
```
posva: _visit -> go_ or _visit orders?page=1 from /orders_ (with the markup you wrote ofc)
No upper X in Vuex
or **a** plain object
if not implemented correctly reads better
I'd remove the last sentence as it doesn't read quite nice and doesn't really say anything

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+Next we have the watch `$route` path without `key` on `router-view`. This gives us more freedom when to reload the data, but it also mean we have to handle it manually. Taking the `/orders` to `/orders?page=1` example, you now can decide if the page query parameter really changed and when to reload the data:
```
posva: I don't completely get the first 2 sentences.
What are you trying to explain here?
It's also easier to watch `$route.query` directly

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+    // We have something like /order?page=2
+    this.loadData();
+  } else {
+    // Our current page and previous page doesn't match
+    this.loadData();
+  }
+}
+```
+
+<p class="tip">Note! All values in `$route.query` are stored as a `string`.</p>
```
posva: The _Note!_ is not necessary
as ~a~ Strings

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+While the snippet could be optimized, it's still quite something just for a `page` query parameter, imagine how it would grow if you have 2 or more similar fields or `page` parameter should reset to `1` when `filter` is changed.
```
posva: I'm really not convinced by this whole section 🤔 
Is it about syncing pagination + query with url what you were aiming for?

donnysim: Not really about syncing, but avoiding unnecessary data reloads. As an example, if pagination component does not prevent user from clicking on the already active page button, which updates the url from `/orders` to `/orders?page=1`, the date should not reload as nothing really changed - the user already was on page 1.

src/v2/cookbook/vue-route-change-handling.md
donnysim: First two sentences are basically saying - we're going for the good old `<router-view></router-view>` logic, no keys etc.. will tweak it a bit.

I kind of think watching `$route.query` or `$route` is just preference and situation dependent. If you have a modal in the same page but it has it's own route, then you might want to watch `$route` to check the from/to so that closing the modal does not start reloading the data etc. though `$route.query` might be more used. I can change this if you prefer it.

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+## Simple, but not good enough
+While both `:key` methods work, there are some drawbacks to each one. Lets start with:
+
+```html
+<router-view :key="$route.fullPath"></router-view>
+```
+
+The `$route.fullPath` returns the full url including the query e.g. `/orders?page=2&filter=search`, so what happens when you go or visit `/orders?page=1` from `/orders`? The page is recreated with new instances of components and the data is reloaded, yet the route is technically the same as before. At this point the only possible workaround would be to cache the results using something like [Vuex](https://vuex.vuejs.org) or a plain object. We could use `$route.path` as a `key`, but this will require introducing a watcher and handling query changes which sometimes can be difficult and unexpected side effects if not implemented correctly.
+
+Other option we have is the `<router-view>` without a `key`. This gives us more freedom to choose when to reload the data, but it also mean we have to decide when it should take place. Taking the `/orders` to `/orders?page=1` example, we can now determine if the `page` query parameter really changed before attempting to reload the data:
```
sdras: Some grammar issues here: 

"Another option we have is the `<router-view>` without a `key`. This gives us more freedom to choose when to reload the data, but it can also mean we have to decide when it should take place. If we take the `/orders` to `/orders?page=1` example, we can now determine if the `page` query parameter really changed before attempting to reload the data"

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+<p class="tip">Remember! Don't modify `$route.query` directly - it does not change the url, use `$router.push` or `$router.replace` instead.</p>
+
+## Simple, but not good enough
+While both `:key` methods work, there are some drawbacks to each one. Lets start with:
+
+```html
+<router-view :key="$route.fullPath"></router-view>
+```
+
+The `$route.fullPath` returns the full url including the query e.g. `/orders?page=2&filter=search`, so what happens when you go or visit `/orders?page=1` from `/orders`? The page is recreated with new instances of components and the data is reloaded, yet the route is technically the same as before. At this point the only possible workaround would be to cache the results using something like [Vuex](https://vuex.vuejs.org) or a plain object. We could use `$route.path` as a `key`, but this will require introducing a watcher and handling query changes which sometimes can be difficult and unexpected side effects if not implemented correctly.
```
sdras: grammar: "and _cause_ unexpected side effects.."

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+```js
+$route(to, from) {
+  if (to.query.page === from.query.page) {
+    // Page did not change
+    return;
+  } else if (to.query.page && !from.query.page) {
+    // We went from /orders to /orders?page=#
+    
+    if (to.query.page === '1') {
+	  // Technically the same page
```
sdras: Small issue, but the indentation is incorrect here.

src/v2/cookbook/vue-route-change-handling.md
```diff
@@ -0,0 +1,376 @@
...
+
+So at this point you'd rather deal with reloading most of the data and be happy with your life, if only you weren't a developer and this wouldn't haunt you that is... and it just keeps piling on - what if the query parameter is a `tab` name, it should not reload anything on the page, just switch the active tab; what if I have index view, and the edit is a modal on top of the view with it's unique route (e.g. `order.edit`), why opening it reloads the list, and so on... these are the things that the `<router-view :key="$route.fullPath"/>` prevents you from achieving, that and the insane amounts of code and logic required to handle the `$route` watcher.
+
+## Partial solution
+While advanced solution could be invented, for most of the use cases a more linear approach is sufficient, so in spite of that, the current solution has limitations:
+
+ - if the user changes url query directly, it will not reset query values (without navigation interception this could cause multiple history entries added when values are being reset);
```
sdras: Please capitalize the first word and put a period at the end of each sentence for each li :)

sdras:
Just did a quick grammar check, I have to go through it in more detail- I hope to make some time this week! Thanks for your patience.

PR #1520, opened 2018-03-21
Using the term "Base" instead of "Simple"

sdras:
- Based on a conversation with @chrisvfritz, changing the term "Simple Example" in the cookbook to "Base Example". He had the idea to change it off a recipe I made by mistake- I like it because the terms simple should be avoided when possible in communicating or documenting something- what's simple to me might not be simple to you, and thus can seem condescending. ref: [http://bradfrost.com/blog/post/just/](http://bradfrost.com/blog/post/just/)
- I also added some language about expectations for the post to allow them a bit of flexibility but clarify what we're looking for.

### DIFF
--- a/src/v2/cookbook/adding-instance-properties.md
+++ b/src/v2/cookbook/adding-instance-properties.md
@@ -4,7 +4,7 @@ type: cookbook
 order: 2
 ---
 
-## Simple Example
+## Base Example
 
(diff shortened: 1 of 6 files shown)


sdras:
Sigh... prettier strikes again on the formatting. I'll go through and clean that up. 🙃

PR #1525, opened 2018-03-24
Clarify time-traveling

phanan:
Vuex/vue-devtools doesn't really provide time traveling. Not that I know of, at least :)

### DIFF
--- a/src/v2/guide/state-management.md
+++ b/src/v2/guide/state-management.md
@@ -6,7 +6,7 @@ order: 502
 
 ## Official Flux-Like Implementation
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I like the clarification, 🙂 but wonder if we should use a different link. I worry that people might assume time travel debugging is a Windows feature or something. Maybe we should link to this image instead:

https://raw.githubusercontent.com/vuejs/vue-devtools/master/media/demo.gif

What do you think?

phanan:
Sure. Do you want to update the link yourself though? I’m going to bed now,
too tired :(

Chris Fritz <notifications@github.com> schrieb am So. 25. März 2018 um
03:40:

Issue #1316, opened 2017-12-12
Documentation/Community Issue - Memory Leaks 


freeman-g:
Hello,

My company has invested pretty heavily in Vue for a large and fairly complex SPA that we use as a core part of our business. Overall, we love, love, love Vue, but we recently uncovered a bunch of memory leak issues. We couldn’t find much information about this in the community or the Vue docs, so I'm creating this issue to have a dialog about this topic.

Most of the memory leak issues that we found were related to wrapper components. For example, we built a wrapper component for TinyMCE (the popular Rich Text Editor) and it caused a large memory leak. This one was particularly insidious because we used the `beforeDestroy` hook in our wrapper component and called the TinyMCE destroy method, however, when navigating away from a page (vue-router) with a TinyMCE component, Vue patched the DOM faster than TinyMCE was able to destroy, which caused the memory leak.

In another case, we used Sortable.js to enable some drag and drop functionality and in this case we didn’t call the Sortable destroy method which caused another memory leak. So I think this is a pretty big trap out there that Vue apps could run into.

Looking in the vuejs.org docs, there is nothing much about memory, memory issues or memory best practices. We spent hours debugging this with Chrome tools and different combinations of components, keep-alive settings, etc.

I think that lots of community components (especially wrapper components) also could have memory leaks (we discovered one in vue-tables-2 https://github.com/matfish2/vue-tables-2/issues/347)

I think generally the main lesson learned is: whenever we use wrapper components or 3rd party components that manipulate the DOM, make sure to use their native api's destroy method in a beforeDestroy hook and also make sure to test the memory usage and garbage collection. We tested by using Chrome Task Manager (Shift-Esc) and watching the memory usage.

Anyway, any thoughts and dialog on this would be much appreciated.

chrisvfritz:
Hi @freeman-g - thanks for opening the issue. 🙂  I also appreciate you hunting around the community and helping projects find and eliminate these leaks. I think this would be a great topic to cover in the upcoming Vue Cookbook, which @sdras is leading. That said, while we can (and plan to) provide more examples and explanations of best practices for wrapper components, Vue _is_ limited in what it can do to fix the problem of memory leaks in frontend applications more broadly. 

[...]

PR #1526, opened 2018-03-25
Add avoiding-memory-leaks cookbook page

freeman-g:
@sdras, this pull request is related to [#1316](https://github.com/vuejs/vuejs.org/issues/1316) on the topic of "Avoiding Memory Leaks".

I created a simple example of some code which causes a memory leak and then how to resolve that memory leak. In this case I put a couple of examples on CodePen so that people could actually see the memory leak in action.

Let me know if you want to me to change anything.

Thanks


sdras:
Awesome! We're all at VueConf US this week, but I'll be able to give it a good review the week following! Thanks so much for submitting, I look forward to reading it.

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Simple Example
+
+The following example shows a memory leak caused by using the Choices.js library in a Vue component (and not properly cleaning it up). Later, we will show how to remove the Choices.js footprint and avoid the memory leak.
+
+In the example below, we load up a select with a lot of options and then we use a show/hide button with a `v-if` directive to add it and remove it from the virtual DOM. The problem with this example is that the v-if directive removes the parent element from the DOM, but we did not clean up the additional DOM pieces created by Choices.js, causing a memory leak. 
+
+To see this memory leak in action, open this [CodePen example](https://codepen.io/freeman-g/pen/qobpxo) and then, using Chrome, open the Chrome Task Manager (Chrome Menu > More Tools > Task Manager). Now, click the show/hide button 50 or so times. You should see the memory usage in the Chrome Task Manager increase and never be reclaimed.
+
+```
```
sdras: This needs to be split up into js and html tags to do syntax highlighting properly and follow the form of the rest of the documentation, which can be accomplished with `js` or `html` at the end of the three backticks

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,138 @@
...
+  }
+})
+
+```
+
+In the above example, we can use our `hide()` method to do some clean up and solve the memory leak prior to removing the select from the DOM. To accomplish this, we will keep a property in our Vue instance’s data object and we will use the [Choices API’s](https://github.com/jshjohnson/Choices) `destroy()` method to perform the clean up.
+
+Check the memory usage again with this [updated CodePen example](https://codepen.io/freeman-g/pen/mxWMor).
+
+```
```
sdras: same here

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,138 @@
...
+
+Memory leaks in Vue applications do not typically come from Vue itself, rather they can happen when incorporating other libraries into an application.
+
+## Simple Example
+
+The following example shows a memory leak caused by using the Choices.js library in a Vue component (and not properly cleaning it up). Later, we will show how to remove the Choices.js footprint and avoid the memory leak.
+
+In the example below, we load up a select with a lot of options and then we use a show/hide button with a `v-if` directive to add it and remove it from the virtual DOM. The problem with this example is that the v-if directive removes the parent element from the DOM, but we did not clean up the additional DOM pieces created by Choices.js, causing a memory leak. 
+
+To see this memory leak in action, open this [CodePen example](https://codepen.io/freeman-g/pen/qobpxo) and then, using Chrome, open the Chrome Task Manager (Chrome Menu > More Tools > Task Manager). Now, click the show/hide button 50 or so times. You should see the memory usage in the Chrome Task Manager increase and never be reclaimed.
```
sdras: Usually we embed codepens, but I think in this case it would be really helpful to show a gif of what you're describing in action. That way they can go to the pen to see if for themselves, but they don't have to leave the flow of the recipe to see what you mean.

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,138 @@
...
+    },
+    hide() {
+      // now we can use the reference to Choices to perform clean up here 
+      // prior to removing the elements from the DOM
+      this.choicesSelect.destroy()
+      this.showChoices = false
+    }
+  }
+})
+```
```
sdras: This is a really nice small example

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,138 @@
...
+## Additional Context
+
+In the above example, we used a `v-if` directive to illustrate the memory leak, but a more common real-world example happens when using **vue-router** to route to components in a Single Page Application.
+
+Just like the v-if directive, vue-router removes elements from the virtual DOM and replaces those with new elements when a user navigates around your application. The Vue `beforeDestroy()` [lifecycle hook](https://vuejs.org/v2/guide/instance.html#Lifecycle-Diagram) is a good place to solve the same sort of issue in a vue-router based application.
+
+We could move our clean up into the `beforeDestroy()` hook like this:
+
+```
+mounted: function() {
```
sdras: I'm not sure we need `mounted` here. I guess it does provide a bit more context, but it seems a little weird to me with nothing in it. The three dots above `beforeDestroy` could probably be enough.

sdras:
This is a very well-written recipe, the progression is really nice! I have only a few small requests. I think this will be really helpful to people. Thanks!

freeman-g:
@sdras thanks for the feedback, everything you said made sense to me. I made the updates and pushed a new commit.

In the commit, I added an image of the chrome task manager to show people what I was talking about. I wasn't sure if you were envisioning just that or something more like the image below, with both the page and the task manager in the image. Let me know if you think the image below would be better than what I did.

Thanks!

image

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,139 @@
...
+## Resolving the Memory Leak
+
+In the above example, we can use our `hide()` method to do some clean up and solve the memory leak prior to removing the select from the DOM. To accomplish this, we will keep a property in our Vue instance’s data object and we will use the [Choices API’s](https://github.com/jshjohnson/Choices) `destroy()` method to perform the clean up.
+
+Check the memory usage again with this [updated CodePen example](https://codepen.io/freeman-g/pen/mxWMor).
+
+```js
+const app = new Vue({
+  el: "#app",
+  data: function() {
```
sdras: In the vue docs, we separate function with a space, like this `function ()`. I had an issue with this because of Prettier, so if that's your case too and you use VS Code, you can use this extension to toggle it on and off for the file: https://marketplace.visualstudio.com/items?itemName=tombonnike.vscode-status-bar-format-toggle

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,139 @@
...
+---
+## Introduction
+
+If you are developing applications with Vue, then you need to watch out for memory leaks. This issue is especially important in Single Page Applications (SPAs) because by design, users should not have to refresh their browser when using an SPA, so it is up to the Javascript application to clean up components and make sure that garbage collection takes place as expected.
+
+Memory leaks in Vue applications do not typically come from Vue itself, rather they can happen when incorporating other libraries into an application.
+
+## Simple Example
+
+The following example shows a memory leak caused by using the Choices.js library in a Vue component and not properly cleaning it up. Later, we will show how to remove the Choices.js footprint and avoid the memory leak.
```
sdras: Can you please link up the choices.js lib here?

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,139 @@
...
+
+In the example below, we load up a select with a lot of options and then we use a show/hide button with a `v-if` directive to add it and remove it from the virtual DOM. The problem with this example is that the v-if directive removes the parent element from the DOM, but we did not clean up the additional DOM pieces created by Choices.js, causing a memory leak.
```
sdras: It might be good to link up the api reference to `v-if` here, though not critical because it's a pretty basic directive. I leave it up to you.

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,139 @@
...
+      this.choicesSelect.destroy()
+      this.showChoices = false
+    }
+  }
+})
+```
+
+## Additional Context
+
+In the above example, we used a `v-if` directive to illustrate the memory leak, but a more common real-world example happens when using **vue-router** to route to components in a Single Page Application.
```
sdras: This should link up to vue-router, thanks

sdras:
That screenshot looks great!

Reading through this again, we are missing both "details about the value" section and "alternative patterns". I think these would really help this recipe- in "details about the value" you wouldn't have to go into a lot of detail, but since you have a concrete use case, you could discuss for a second what happens when memory leaks aren't dealt with appropriately, especially when you reach scale. It would also help give the "why". [Consider this article](https://auth0.com/blog/four-types-of-leaks-in-your-javascript-code-and-how-to-get-rid-of-them/): when they discuss the importance of unwanted references, it sets the stage for showing how to solve the problem. Does that make sense?

Alternative patterns section is tricky for this one, but I think we could make it work. Even linking off to an article like the above could possibly be helpful- you could state that this isn't the only kind of memory leak that can happen in a Vue application, and here are some other things to look out for. That said, I'm open to other ideas as well!

Please don't hesitate to let me know if you need clarification on anything. I'm excited for this recipe, I think it will be super helpful.

freeman-g:
Thanks for the feedback - I changed the section "Additional context" to "Real world example" because I noticed that the contributing guide marked that section as required.

Apart from that, I think addressed everything except for the Alternative patterns part.

I was kinda stuck on the Alternative patterns part. For the alternative patterns I had another idea that we could discuss wrapper components as an alternative pattern. With wrapper components as another alternative, we could essentially make self-memory-managing components. What do you think about that?

sdras:
Hi @freeman-g! I love the idea of talking about `keep-alive` in the alternative patterns section because it could be pretty easy to understand for the reader, plus you could just link off to the docs. The wrapper components would also work! But puts more work on you to explain concepts. I think this recipe is really close. Thanks for your patience!

freeman-g:
@sdras 
Cool - I added an Alternative Patterns section discussing keep-alive. Cheers!

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,169 @@
...
+
+Memory leaks in Vue applications do not typically come from Vue itself, rather they can happen when incorporating other libraries into an application.
+
+## Simple Example
+
+The following example shows a memory leak caused by using the [Choices.js](https://github.com/jshjohnson/Choices) library in a Vue component and not properly cleaning it up. Later, we will show how to remove the Choices.js footprint and avoid the memory leak.
+
+In the example below, we load up a select with a lot of options and then we use a show/hide button with a [v-if](/v2/guide/conditional.html) directive to add it and remove it from the virtual DOM. The problem with this example is that the v-if directive removes the parent element from the DOM, but we did not clean up the additional DOM pieces created by Choices.js, causing a memory leak.
```
sdras: v-if should be in backticks here `v-if`, for example

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,169 @@
...
+        this.initializeChoices()
+      })
+    },
+    hide: function () {
+      this.showChoices = false
+    }
+  }
+})
+```
+To see this memory leak in action, open this [CodePen example](https://codepen.io/freeman-g/pen/qobpxo) using Chrome and then open the Chrome Task Manager (Chrome Menu > More Tools > Task Manager). Now, click the show/hide button 50 or so times. You should see the memory usage in the Chrome Task Manager increase and never be reclaimed.
```
sdras: Huh, this is not located here for me. For me it's Chrome top navigation > Window > Task Manager. I'm on version 66.0.3359.117 Mac, US. If it's different for you, we could always include both and just say it's either here or here.

src/v2/cookbook/avoiding-memory-leaks.md
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+})
+```
+
+## Details about the Value
+
+Memory management and performance testing can easily be neglected in the excitement of shipping quickly, however, keeping a small memory footprint is still important to your overall user experience. 
+
+Consider the types of devices your users may be using and what their normal flow will be. Could they use memory constrained laptops or mobile devices? Do your users typically do lots of in-application navigation? If either of these are true, then good memory management practices can help you avoid the worst case scenario of crashing a user’s browser. Even if neither of these are true, you can still potentially have degradation of performance over extended usage of your app if you are not careful.
```
sdras: Great job with this section

sdras:
This is SO CLOSE! And it's SO AWESOME! I think you did a great job in the alternative patterns section, bravo. There are only a couple really small things left and I'll merge it in!

src/v2/cookbook/avoiding-memory-leaks.md
freeman-g: Hi - I also have it under Window > Task manager. I can change to this since its easier to find there.

On Windows, there is a Shift-Esc shortcut to open the Task Manager. I can't find a mac shortcut.

Should I explain both the windows and mac approaches? Something like this?

```
...then open the Chrome Task Manager. To open the Chrome Task Manager on Mac, choose Chrome Top Navigation > Window > Task Manager or on Windows, use the Shift+Esc shortcut.
```

sdras: Yes that's perfect 👍

freeman-g:
Ok - hopefully we've got it this time! Thanks

sdras:
Yesssss! 🎉Congrats!

freeman-g:
@sdras - Awesome thanks for helping me with my very first PR, and thanks for everything y’all are doing with Vue!

PR #1533, opened 2018-03-27
Update adding-instance-properties.md

olegvakulenko:
Property names mismatch.

var App = Object.freeze({
  name: 'My App',
  **description**: '2.1.4',
  helpers: {
    // This is a purely functional version of
    // the $reverseText method we saw earlier
    reverseText: function(text) {
      return text
        .split('')
        .reverse()
        .join('')
    }
  }
})

new Vue({
  data: {
    appVersion: App.**version**
  },
  methods: {
    reverseText: App.helpers.reverseText
  }
})


sdras:
Nice catch, thanks.

PR #1535, opened 2018-03-28
3rd Party Plugin Example Update: Add trigger method to watcher in Select2 example

jryd:
In the Select2 example, the watcher for `value` is missing the `.trigger('change')` method to update the selected drop down value in the Select2 plugin.
Without this, the `val(value)` will be set correctly, but the selected option that appears in the select box will not change.


chrisvfritz:
Thanks! 🙂 I just updated the existing example, so that people linking directly to the fiddle will still see the corrected version.

PR #1549, opened 2018-03-31
Update adding-instance-properties.md

arrilot:


sdras:
Thanks for the PR! Unfortunately, this makes less sense grammatically than what we had before, so I have to close it out.

PR #1550, opened 2018-03-31
cookbook entry for storage

cfjedimaster:
First draft for a new cookbook entry. I did not optimize the image as I saw the other images were pretty big too. That feels wrong, but I wanted to be consistent. :)


src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,162 @@
+---
+title: Client-Side Storage
+type: cookbook
+order: 10
+---
+
+## Base Example
+
+Client-side storage is an excellent way to quickly add performance gains to an application. By storing data on the browser itself, you can skip fetching information from the server every time the user needs it. While especially useful when offline, even online users will benefit from using data locally versus a remote server. Client-side storage can be done with [cookies](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies), [Local Storage](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API) (technically "Web Storage"), [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API), and [WebSQL](https://www.w3.org/TR/webdatabase/) (a deprecated method that should not be used in new projects).
```
sdras: great job linking off to further documentation 👍

src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,162 @@
...
+const app = new Vue({
+  el:'#app',
+  data:{
+    name:''
+  },
+  mounted() {
+    if(localStorage.name) this.name = localStorage.name;
+  },
+  watch:{
+    name(newName,oldName) {
```
sdras: I wouldn't include the second param here if it's going to remain unused.

sdras:
This looks very clean! I did an initial pass, I need another day or so and I'll test out all the examples and continue the review. Thank you so much Raymond.

src/v2/cookbook/client-side-storage.md
cfjedimaster: Normally I'd agree - but do you think it's worthwhile keeping it there just as a reminder of the watch API in general?

sdras: No, I don't think so. You can link to the api section of the docs for details like this along with an explanation for clarity. If you wouldn't write the function in this manner for this example, I'd remove it because it makes it more complex and thus, harder to understand at a glance, even if only slightly.

cfjedimaster: Ok - fair enough. Removed (and the codepen was edited as well).

src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,162 @@
...
+
+In this cookbook entry we'll focus on Local Storage, the simplest of the storage mechanisms. Local Storage uses a key/value system for storing data. It is limited to storing only simple values but complex data can be stored if you are willing to encode and decode the values with JSON. In general, Local Storage is appropriate for smaller sets of data you would want to persist, things like user preferences or form data. Larger data with more complex storage needs would be better stored typically in IndexedDB.
```
sdras: I'm not sure everyone would consider an array complex data, so it might be better to be more specific here- "It would be better to store data structures such as objects or arrays in something like IndexedDB" for example

src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,162 @@
...
+```
+
+In this application, we've switched to use the Local Storage APIs versus "direct" access. Both work but the API method is generally preferred. `mounted` now has to grab the value and parse the JSON value. If anything goes wrong here we assume the data is corrupt and delete it. (Remember, any time your web application uses client-side storage, the user has access to it and can modify it at will.)
+
+We have three methods now to handle working with cat. Both `addCat` and `removeCat` handle updating the "live" Vue data stored in `this.cats`. They then run `saveCats` which handles serializing and persisting the data. You can play with this version below:
+
+<p data-height="265" data-theme-id="0" data-slug-hash="qoYbyW" data-default-tab="js,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="localstorage, complex" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/qoYbyW/">localstorage, complex</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://static.codepen.io/assets/embed/ei.js"></script>
+
+## Alternative Patterns
```
sdras: I like this alternative patterns section but it might be nice to have a wrapping up section.

cfjedimaster:
Thank you for the review! As an FYI, I'm traveling out of country this week so I may be a bit slow to respond.

src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,162 @@
...
+Focus on the `mounted` and `watch` parts. We use `mounted` to handle loading the value from local storage. To handle writing the data base, we simply watch the `name` value and on change, immediately write it. 
+
+You can run this yourself here:
+
+<p data-height="265" data-theme-id="0" data-slug-hash="KodaKb" data-default-tab="js,result" data-user="cfjedimaster" data-embed-version="2" data-pen-title="testing localstorage" class="codepen">See the Pen <a href="https://codepen.io/cfjedimaster/pen/KodaKb/">testing localstorage</a> by Raymond Camden (<a href="https://codepen.io/cfjedimaster">@cfjedimaster</a>) on <a href="https://codepen.io">CodePen</a>.</p>
+<script async src="https://static.codepen.io/assets/embed/ei.js"></script>
+
+Type something in the form and then reload this page. You'll note that the value you typed previously will show up automatically. Don't forget that your browser provides excellent developer tools for inspecting client-side storage. 
+
+![Storage devtools in Firefox](/images/devtools-storage.png)
```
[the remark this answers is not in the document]
cfjedimaster: I was very intentionally trying *not* to use Chrome as I feel like Firefox needs more attention. :) I can absolutely add another screen shot, but to me, it feels like overkill. I will do as you ask though if you feel strongly about this. :)

src/v2/cookbook/client-side-storage.md
cfjedimaster: Query - would that go before or after the Alternative Patterns?

cfjedimaster:
I've made the changes requested, except for the one item I disagreed on (although not strongly), the tip about removing/clearing local storage, and the last item.

For the tip - I agree - but is there a standard markup for the site I should use to denote the tip?

src/v2/cookbook/client-side-storage.md
sdras: "Wrapping up" is just a sentence or two at the end, after the Alternative Patterns section

src/v2/cookbook/client-side-storage.md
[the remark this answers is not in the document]
sdras: I guess Firefox can use the attention but Chrome is more useful to people reading this. I'm not sure our job is to force adoption in browsers, it's rather our job to make it easy for people to work with the content. I'm not going to push you on it, it's fine as is. But the idea about writing technical content in the first place, IMO, is meeting people where they are to help them learn. If I was reading through this, I would then have to go research how to do this in Chrome.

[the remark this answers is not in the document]
sdras: Nice :) thank you

sdras:
Thank you Raymond! Tips use `<p class="tip"></p>`

src/v2/cookbook/client-side-storage.md
```diff
@@ -0,0 +1,176 @@
...
+<script async src="https://static.codepen.io/assets/embed/ei.js"></script>
+
+## Alternative Patterns
+
+While the Local Storage API is relatively simple, it is missing some basic features that would be useful in many applications. The following plugins wrap Local Storage access and make it easier to use, while also adding functionality like default values.
+
+* [vue-local-storage](https://github.com/pinguinjkeke/vue-local-storage)
+* [vue-reactive-storage](https://github.com/ropbla9/vue-reactive-storage)
+
+While the browser will never replace a server persistence system, having multiple ways to cache data locally can be a huge performance boost for your application, and working with it in Vue.js makes it even more powerful.
```
sdras: This should literally be an h2, like ## Wrapping Up

sdras:
Hi! We're so close. We just need that h2 in there, and yes, please optimize the images- I'm not sure historically why the others aren't optimized, but they should be. Once those two things are addressed, we're good to merge! 🍻

cfjedimaster:
Changes made.

sdras:
Argh! Please forgive me, I just noticed that your order is 10 and we already have a 10, your entry would have to be 11. THEN we can merge.

cfjedimaster:
Done - and I quit. :p

sdras:
You're a hero, Raymond. Thanks for all of your hard work getting this in.

cfjedimaster:
Glad to help the Vue cause! :)

PR #1552, opened 2018-04-01
Use correct spelling (https://webpack.js.org/branding/#the-name)

dear-lizhihua:


chrisvfritz:
Thanks, but this is actually intentional. 🙂 That's what Webpack uses for their internal branding, but in these docs, we choose to respect more general grammar and ecosystem conventions (e.g. project names should be Title Case, acronyms should be UPPERCASE), rather than specific project conventions (e.g. "webpack" and "npm"). 

The reason is that respecting the individual branding quirks for each library tends to cause confusion and often looks very strange to users. For example, in this sentence:

> webpack and Browserify both support [...]

The lowercase letter at the beginning of the sentence looks like a mistake and also next to Browserify, people often wonder why "Webpack" wouldn't also be capitalized.

PR #1553, opened 2018-04-01
Clarify usage of data in render function doc, fixes #1527

### LINKED ISSUE #1527: Docs: Need clarification on data referenced for functional components
In the docs for Functional Components in the [Render Functions](https://vuejs.org/v2/guide/render-function.html) section, `data` is referred in two different places that is initially confusing:

> In cases like this, we can mark components as functional, which means that they're stateless (no data) and instanceless (no this context).

>Everything the component needs is passed through context, which is an object containing:
>    ...
>    data: The entire data object passed to the component

From talking with @chrisvfritz, these refer to two different things. So, this issue is to track how it is either defined or clarify it for future readers.

chrisvfritz:
How's this for a potential clarification, building on [@LinusBorg's suggestion](https://github.com/vuejs/vuejs.org/issues/1527#issuecomment-376431930)?

### DIFF
--- a/src/v2/guide/render-function.md
+++ b/src/v2/guide/render-function.md
@@ -499,7 +499,7 @@ For more on how JSX maps to JavaScript, see the [usage docs](https://github.com/
 
 The anchored heading component we created earlier is relatively simple. It doesn't manage any state, watch any state passed to it, and it has no lifecycle methods. Really, it's only a function with some props.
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
Also tagging @bencodezen for your thoughts. 🙂

bencodezen:
This looks great! 🙌 💯

PR #1554, opened 2018-04-01
Clarify key in list rendering doc, fixes #1540

### LINKED ISSUE #1540: Clarify two different uses of `keys` near eachother in intro
I just had to clarify for someone just starting [this section](https://vuejs.org/v2/guide/list.html#v-for-with-an-Object) and [the section following](https://vuejs.org/v2/guide/list.html#key) because we use keys for both but they're referring to different types of keys.

chrisvfritz:
Does this sufficiently highlight the distinction?

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -244,11 +244,11 @@ type: api
 
 - **See also:** [Async Update Queue](../guide/reactivity.html#Async-Update-Queue)
 
-### Vue.set( target, key, value )
+### Vue.set( target, propertyName/index, value )
 
 - **Arguments:**
   - `{Object | Array} target`
-  - `{string | number} key`
+  - `{string | number} propertyName/index`
   - `{any} value`
 
 - **Returns:** the set value.
@@ -261,11 +261,11 @@ type: api
 
 - **See also:** [Reactivity in Depth](../guide/reactivity.html)
 
-### Vue.delete( target, key )
+### Vue.delete( target, propertyName/index )
 
 - **Arguments:**
   - `{Object | Array} target`
-  - `{string | number} key/index`
+  - `{string | number} propertyName/index`
 
   > Only in 2.2.0+: Also works with Array + index.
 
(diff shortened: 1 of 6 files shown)


sdras:
Hm. I think it's a start but I wonder if it's still confusing to newcomers. Let me ruminate on this and ask a couple of people who don't have the whole "curse of knowledge" going on that we do.

sdras:
So, I asked a couple of students and they said they were confused- mainly because these sections are next to each other and refer to the same word used for different things, both with `v-for`.

Here are a few possible solutions that I see (open to other avenues):
1. We could have a tip/note that explains the differences. This could work but possibly imposes a cognitive load on people just learning concepts.
2. We could separate out the sections. This is not ideal, because they do flow from one to another.
3. We could change the example above the `key` section to use `index` or something else instead. I think this last one would work most seamlessly for the reader, but has two problems- it's the most work of the three, and then we haven't shown people that they can also use `key` in a `v-for`. 

What do you think?

chrisvfritz:
Hm, I agree none of those are ideal. What would you think about instead renaming "The `key` Attribute" to something like "Maintaining State"? That way, we won't be using the word "key" _directly_ after the we use it in another context. Instead, there are a few paragraphs of cleanser context. 😃 Would that work?

sdras:
Yes! I think that would work. Maybe we should also have a note that explicitly states that there is a fundamental difference here. Sounds good.

chrisvfritz:
It actually just occured to me that for plain objects, property keys can more specifically be called names - since a key can be a name (string) or index (number). [MDN's page on objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects) even _primarily_ uses "name" and "propertyName" to refer to keys. So I just updated the PR to use language more consistent with MDN across the docs, which I think should also do a lot to clarify this distinction so that readers don't have to keep context-switching.

What do you think @sdras?

sdras:
This is perfect! Thank you Chris! Sorry I missed this PR for so long

PR #1557, opened 2018-04-03
Update components.md

cfjedimaster:
I removed the underscores around "prop"erty. I'm not sure what was intended there- maybe bold or italics, but I figured at minimum it should be removed until someone can correct it better.


chrisvfritz:
Yes, that was supposed to render as italics. 🙂

Issue #1545, opened 2018-03-30
[Cookbook] Packaging a Vue Single File Component (SFC) to share on npm


mgdodge:
I have code and documentation that I prepared as part of my lightning talk at VueConfUS 2018, and [was told it would be a great addition to the cook book](https://github.com/mgdodge/vue-sfc-rollup/issues/1).

  - **Solve a specific, common problem**: I was unable to find any concise documentation describing the steps required to package my SFC to share via npm. This recipe would describe the minimal steps to get an SFC ready to publish to npm.
  - **Start with the simplest possible example**: Simple example includes a total of 3 files outside the sample SFC itself. Those 3 files are all very brief and well documented.
  - **Link to other docs, rather than re-explaining concepts**: I provide links to the official npm documentation which describe how to name npm packages, what fields to include, etc.
  - **Explain the process, rather than just the end result**: While I did write a wizard to automate the process, I intentionally kept things as simple as possible so that the process would be easy to explain through comments, easily identified/editable template strings, and sample bash commands. This process would be explained in the recipe.
  - **Explain the pros and cons of your strategy, including when it is and isn’t appropriate**: I point out that this is a great way to quickly publish simple components, but is not meant to replace the full development experience orchestrated by the vue cli. This is also not a good candidate (as written) for full applications or complex components with multiple external resources (CSS, images, etc.) though those could be added to this recipe as an exercise for the user.
  - **Mention alternative solutions...**: The reason I wrote this recipe was because I was unable to find any adequate alternatives. I found one or two templates for getting started, but nothing that was very simple and included the entire process. Happy to link to any that I find in the meantime, or suggestions to others. Also, this uses rollup instead of webpack, but this could very easily be done with webpack (especially now that v4 has tree-shaking abilities as well.)

My npm submission for this, found at https://github.com/mgdodge/vue-sfc-rollup, includes a lot of the example code and README content that would go into this recipe. A proper PR will be made when this is accepted as a recipe.

sdras:
This is really detailed, thank you! As mentioned before, I think this will be a really useful recipe and benefit the community because it will help people release what they're working on. Onward! Can't wait for the PR.

sdras:
This is in

PR #1558, opened 2018-04-03
First stab at SFC to npm documentation

mgdodge:
PR for #1545 

CC @sdras 

Feedback welcome!


sdras:
You are awesome! I'm slammed at the moment, but I'll be able to review soon :)

znck:
With 2.2+, we have SSR build which uses string concatenation and as a result it is 2x to 10x faster. If you bundle for browser, the SSR performance is hampered. If you bundle for server, it won't work in browser. So, publishing `.vue` file is not a bad idea. 

In my opinion, there should be a build which can be used directly in browser. Then, a .vue component which uses css in style, html in template and es2015 in script,  so that it can be used in most cases.

@yyx990803 WDYT?

yyx990803:
Thanks for the guide, @mgdodge !

There are a few things we want to resolve for the long run:

1. Make it possible to use `*.vue` as the primary format for component distribution. Currently there is a blocker: Pre-processors inside `*.vue` files would require the consumer to have appropriate build setups. This can be alleviated by us providing a utility to compile raw SFCs down to "plain" SFCs that do not require any special pre-processing.

2. Before (1) gets resolved,  the methods described in this PR can serve as the recommended approach for the time being, with one caveat: SSR. It might be possible to produce an extra SSR dist file, but it requires some build aliasing so that the server build is using the SSR dist files.

mgdodge:
@posva Good catch on the typo, thanks! And the other suggestions are great. Will make the changes shortly.

@yyx990803 I understand the long-term goals you mentioned, and agree with (1) in that it would be nice to present a 'plain' SFC. Forgive my ignorance on the way SSR works (no use cases at my day job,) but I thought in the end all vue components have to be converted to render functions anyway...which is why I thought this would be adequate. Is this not the case?

Regardless, I can update the recipe to include usage of the `browser` field for the package.json file. According to the [package.json spec](https://github.com/defunctzombie/package-browser-field-spec#alternate-main---basic), we can be creative in this (terribly misnamed, imho) to override the `main` field.

If my understanding is correct, the following package.json:
```json
{
  "name": "my-component",
  "version": "1.2.3",
  "main": "dist/my-component.umd.js",
  "module": "dist/my-component.esm.js",
  "unpkg": "dist/my-component.min.js",
  "browser": {
    "./sfc": "src/my-component.vue"
  },
  ...
}
```
should do the trick. Using the component via import/require `import MyComponent from 'my-component';` in an older build process will still resolve to `main`. Someone using a newer one will get `module`. The `unpkg` version is still intended for use in `script` tags. A slight tweak to the import - `import MyComponent from 'my-component/sfc';` and it resolves the alias to the .vue file. Lowest impact I can come up with to satisfy all scenarios.

This is all based on some very rudimentary testing, so **I would love to have someone more familiar with the SSR process verify this**. SFC authors can indicate that the .vue file file may be imported via `my-component/sfc` for any performance optimization SSR provides.

mgdodge:
@posva @yyx990803 @znck I have updated the PR to address your concerns. Thanks for the advice!

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+title: Packaging Vue Components for npm
+type: cookbook
+order: 10
+---
+
+## Base Example
+
+Vue components by nature are meant to be re-used. This is easy when the component is only used within a single application. But how can you write a component once and use it in multiple sites/applications? Perhaps the easiest solution is via npm.
+
+By packaging your component to be shared via npm, it can be imported/required into a build process for use full-fledged web applications:
```
sdras: Typo here: "for use in full-fledged.." (Intro is great!)

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+  <script src="https://unpkg.com/vue"></script>
+  <script src="https://unpkg.com/my-component"></script>
+  ...
+  <my-component></my-component>
+  ...
+```
+
+Not only does this help you avoid copy/pasting components around, but it also allows you to give back to the Vue community!
+
+## Can't I Just Share .vue Files Directly?
```
sdras: We should add the backticks around `.vue` here

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+
+Vue already allows components to be written as a single file. Because a Single File Component (SFC) is already just one file, you might ask:
+
+> "Why can't people use my .vue file directly? Isn't that the simplest way to share components?"
```
sdras: Here too

sdras: Just, all the way down :)

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+
+It's true, you could share .vue files directly, and as long as the [Vue build](https://vuejs.org/v2/guide/installation.html#Explanation-of-Different-Builds) being used contains the Vue compiler, that would be enough. For example, string concatenation is used as an optimization in the SSR build, so the .vue file might be preferred in this scenario (see [Packaging Components for npm > SSR Usage](#SSR-Usage) for details). But this excludes anyone who wishes to use the component directly in a browser via `<script>` tag, anyone who uses a runtime-only build, or a build process which doesn't understand what to do with .vue files.
```
sdras: I like how conversational this paragraph is, however the flow is a bit strange. "and as long as the build you are using contains the vue complier, that would be enough."

Later, maybe replacing "But" with "However," at the beginning of the sentence.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+
+In addition, pay attention to any dependencies that your SFC might have. For example, if you require a third party library for sorting or communication with an API, Rollup might roll those packages into the final code if not properly configured. To continue using this recipe, you would need to configure Rollup to exclude those files from the output, then update your documentation to inform your users about these dependencies.
+
+## Alternative Patterns
+
+At the time this recipe was written, Vue CLI 3 was itself in beta. This version of the CLI comes with a built-in `library` build mode, which creates CommonJS and UMD versions of a component. This might be adequate for your use cases, though you will still need to make sure your package.json file points to `main` and `unpkg` properly. Also, there will be no ES6 `module` output unless that capability is added to the CLI before its release or via plugin.
+
+The process here is the result of what was learned looking at other vue modules which have already been published by the greater vue community. The reason for the recipe was because there was a lack of well documented patterns for preparing a component for npm.
```
sdras: I would omit this paragraph, though I appreciate your course of action, it sounds unnecessary and separate from rest of the recipe.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,214 @@
...
+
+## Acknowledgements
+
+This recipe is the result of a lightning talk given by [Mike Dodge](https://twitter.com/mgdodgeycode) at VueConf.us in March 2018. He has published a utility to npm which will quickly scaffold a sample SFC using this recipe. You can download the utility, [vue-sfc-rollup](https://www.npmjs.com/package/vue-sfc-rollup), from npm. You can also [clone the repo](https://github.com/mgdodge/vue-sfc-rollup) and customize it to fill your needs.
```
sdras: "to fill it with your needs" is odd phrasing. I think "customize it" will suffice.

sdras:
Looking good! I reviewed for clarity and flow, I'm giving it one more round soon while reproducing. You can start on those edits though, if you want to push it forward in the meantime. :)

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: I will rework this paragraph a little to change the flow.

mgdodge:
Just curious on the status of this.

sdras:
Thank you for the ping, we haven't forgotten about you! I have MSFT Build coming up, and can review it after. Thanks for your patience.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
+---
+title: Packaging Vue Components for npm
+type: cookbook
+order: 10
```
sdras: We have order 10 and 11 now, this would need to be 12

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+  },
+  ...
+}
+```
+
+When webpack 2+, Rollup, or other modern build tools are used, they will pick up on the `module` build. Legacy applications would use the `main` build, and the `unpkg` build can be used directly in browsers. In fact, the [unpkg](https://unpkg.com) cdn automatically uses this when someone enters the URL for your module into their service!
+
+### SSR Usage
+
+You might have noticed something interesting - browsers aren't going to be using the `browser` version! That's because this field is actually intended to allow authors to provide [hints to bundlers](https://github.com/defunctzombie/package-browser-field-spec#spec) which in turn create their own packages for client side use. With a little creativity, this field allows us to map an alias to the `.vue` file itself. For example:
```
sdras: Style note- I do like the use of exclamation points to convey excitement but when they're used back to back like this, it seems a little over the top. I'd drop the second one here.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+
+```js
+import MyComponent from 'my-component/sfc'; // Note the '/sfc'
+```
+
+Compatible bundlers see the `browser` definition in package.json and translate requests for `my-component/sfc` into `my-component/src/my-component.vue`, resulting in the original `.vue` file being used instead. Now the SSR process can use the string concatenation optimizations it needs to for a boost in performance.
+
+> Note: When using `.vue` components directly, pay attention to any type of pre-processing required by `script` and `style` tags. These dependencies will be passed on to users. Consider providing 'plain' SFCs to keep things as light as possible.
```
sdras: Can you please put this in a `<p class="tip"></tip>`? That's our docs standard.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+
+### How do I make multiple versions of my component?
+
+There is no need to write your module multiple times. It is possible to prepare all 3 versions of your module in one step, in a matter of seconds. The example here uses [Rollup](https://rollupjs.org) due to its minimal configuration, but similar configuration is possible with other build tools. The package.json `scripts` section can be updated with a single entry for each build target, and a more generic `build` script that runs them all in one pass. The sample package.json file now looks like this:
```
sdras: Should we mention webpack here as well? It's also fine as is.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+
+In addition, pay attention to any dependencies that your SFC might have. For example, if you require a third party library for sorting or communication with an API, Rollup might roll those packages into the final code if not properly configured. To continue using this recipe, you would need to configure Rollup to exclude those files from the output, then update your documentation to inform your users about these dependencies.
+
+## Alternative Patterns
+
+At the time this recipe was written, Vue CLI 3 was itself in beta. This version of the CLI comes with a built-in `library` build mode, which creates CommonJS and UMD versions of a component. This might be adequate for your use cases, though you will still need to make sure your package.json file points to `main` and `unpkg` properly. Also, there will be no ES6 `module` output unless that capability is added to the CLI before its release or via plugin.
```
sdras: Is it worth mentioning https://github.com/vuejs/rollup-plugin-vue?

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+
+> "Why can't people use my `.vue` file directly? Isn't that the simplest way to share components?"
+
+It's true, you can share `.vue` files directly, and anyone using a [Vue build](https://vuejs.org/v2/guide/installation.html#Explanation-of-Different-Builds) containing the Vue compiler can consume it immediately. Also, the SSR build uses string concatenation as an optimization, so the `.vue` file might be preferred in this scenario (see [Packaging Components for npm > SSR Usage](#SSR-Usage) for details). However, this excludes anyone who wishes to use the component directly in a browser via `<script>` tag, anyone who uses a runtime-only build, or build processes which don't understand what to do with `.vue` files.
+
+Properly packaging your SFC for distribution via npm enables your component to be shared in a way which is ready to use everywhere!
+
+## Packaging Components for npm
+
+For the purposes of this section, assume the following file structure:
```
sdras: this type of explanation of structure before diving in is really useful to people reading 👍

src/v2/cookbook/packaing-sfc-for-npm.md
sdras: It might be worth just linking off to a resource like this, in case people are interested: https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c. Your call, though.

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+    "rollup-plugin-vue": "^3.0.0",
+    "vue": "^2.5.16",
+    "vue-template-compiler": "^2.5.16",
+    ...
+  },
+  ...
+}
+```
+
+That is all that is required in package.json to get up and running. Now, all that is needed is a small wrapper to export/auto-install the actual SFC, plus a mimimal Rollup configuration, and we're set!
```
sdras: Hm, when we say "this is all that is needed" which is repeated twice here, it may be hard on the person reading it if they are just grasping these concepts. Consider [this post](http://bradfrost.com/blog/post/just/). I believe you're trying to be supportive of the reader, but it may come across the other way. Can you please consider another way of phrasing one of these two sentences?

src/v2/cookbook/packaing-sfc-for-npm.md
```diff
@@ -0,0 +1,212 @@
...
+    "rollup": "^0.57.1",
+    "rollup-plugin-buble": "^0.19.2",
+    "rollup-plugin-vue": "^3.0.0",
+    "vue": "^2.5.16",
+    "vue-template-compiler": "^2.5.16",
+    ...
+  },
+  ...
+}
+```
```
sdras: One thing I find lacking here reading through is how you created this base package.json. Did you use a generator? Are you going off of another file? How are you deciding what versions should be used for each? Are you manually looking at the current version of LTS in terminal? These are some things that might deserve a quick couple of sentences to clarify to be most helpful.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Will do.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Will do

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Yep.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Not mentioning webpack, but will add link to the article - I did come across it while researching.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Adding another paragraph explaining that this package.json illustrates a starting point, and that versions can/should be kept up to date as necessary.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Rewriting.

src/v2/cookbook/packaing-sfc-for-npm.md
mgdodge: Not sure what you are getting at. rollup-plugin-vue is used by my recipe - similar to vue-loader for webpack, its purpose is to enable Rollup to understand vue files. It doesn't do anything else by itself. Saving the processed JS is up to us.

sdras: Sorry, my bad, I confused this with another library.

sdras:
Looks like all that we're missing from you are those details about the `package.json` and we can merge! Nice job.

mgdodge:
@sdras My last two updates addressed the package.json request you made. There are now two paragraphs - one where I first introduce the file structure and package.json, and the second when I am listing build scripts and dependencies.

Hopefully there isn't anything more that needs to be said. I'm not sure what you meant by looking at LTS in terminal...this stuff should work with any version of node, AFAIK.

PR #1560, opened 2018-04-05
Update join.md

Tapha:
Addition of the Vue subreddit to the list of communities listed.


chrisvfritz:
Hm, what do you think about this one @LinusBorg? To my knowledge, it's not an officially maintained subreddit and I'm hesitant to add non-official channels here.

LinusBorg:
I don't think we should add it. Reddit can be a toxic place in general and its an impractical format to ask for help. The articles hat are linked usually are available though the usual Twitter sources. Our newsletter the other ones out there.

chrisvfritz:
@Tapha Thank you for your work on the subreddit, but I think we will, unfortunately, have to draw the line at platforms maintained and controlled by the core team. This isn't meant to be a complete list of Vue communities, but rather a list of official support channels.

PR #1566, opened 2018-04-06
Suggest adding read-only instance properties

benmccallum:
Could be considered more likely the default/best practice as you'd rarely want it re-assignable later. 
Reference: https://vuejsdevelopers.com/2017/04/22/vue-js-libraries-plugins/

### DIFF
--- a/src/v2/cookbook/adding-instance-properties.md
+++ b/src/v2/cookbook/adding-instance-properties.md
@@ -10,6 +10,10 @@ There may be data/utilities you'd like to use in many components, but you don't 
 
 ```js
 Vue.prototype.$appName = 'My App'
+
+// Or, to make the instance property read-only and avoid accidental re-assignment later,
+// use Object.definePrototype() which by default uses { writeable: false } in the descriptor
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I appreciate the contribution, but I'm not sure this actually adds enough value, since `Vue.prototype.$appName` is extremely unlikely to be _accidentally_ reassigned later. What is more likely to happen user code is `this.$appName = 'something else'`, which is still unlikely to happen by accident, but would not actually mutate the prototype. For example:

```js
Vue.prototype.$appName = 'foo'

new Vue({
  created () {
  	this.$appName = 'bar'
    console.log(Vue.prototype.$appName) //=> "foo"
    console.log(this.$appName) //=> "bar"
  }
})
```

So later on, you'd have to be very explicitly with `Vue.prototype.$appName = 'something else'` in order to override it. Since this adds a bit more complexity to the example but probably has close to a 0% chance of preventing any problems, I think I have to close.

benmccallum:
Sorry about typo. Reading your comment it looks like that blog was wrong then! No worries, thanks!

PR #1568, opened 2018-04-07
Webpack -> webpack

Jinjiang:
ref: https://github.com/vuejs/vue-router/pull/1780#issuecomment-333884730
Thanks.

### DIFF
--- a/src/v2/cookbook/adding-instance-properties.md
+++ b/src/v2/cookbook/adding-instance-properties.md
@@ -158,7 +158,7 @@ So what are the alternatives?
 
 ### When Not Using a Module System
 
-In applications with **no** module system (e.g. via Webpack or Browserify), there's a pattern that's often used with _any_ JavaScript-enhanced frontend: a global `App` object.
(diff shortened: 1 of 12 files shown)


chrisvfritz:
Thanks, but this is [actually intentional](https://github.com/vuejs/vuejs.org/pull/1552#issuecomment-377817937). I'm open to further discussion on this issue though, if you think most users will have a better experience with lowercased `webpack`.

Jinjiang:
I think they are both fine for the reading experience. At the same time, I'd rather like to respect the original name as possible as we can. 🙂
Thanks.

chrisvfritz:
Then I'll keep it the way it is for now - it definitely is a difficult choice. Whether we respect the conventions of the wider community (Title Case for project names) and English grammar (capitals at the beginning of a sentence), or respect the branding guidelines of individual projects? Either way, we can't make everyone happy. 😅

Translators are definitely welcome to make a different choice in their versions though, because these choices may feel more or less natural depending on the language and culture.

Justineo:
On this case I personally would respect the project owners' intentions (if they do care, and we can tell that from their docs or websites). And I agree that lowercased branding like “webpack” in translated docs (like Chinese) won't confuse readers as much as in English since there won't be other lowercased words all around. 😄

PR #1575, opened 2018-04-12
Update name description

janschoenherr:
To me it seems, the removed part no longer holds true. In stack traces the component names no longer show. Instead the uniform "VueComponent" is used.


chrisvfritz:
It's still true. For example:

screen shot 2018-04-12 at 12 02 01 pm

If you post a fiddle/pen demonstrating the issue, I can help you figure out what's going wrong in your case. 🙂

Issue #1521, opened 2018-03-22
[Cookbook] Using Vue in a CMS template


lewisvatcky:
A section of the cookbook could be using Vue for isolated CMS-driven components. The idea could help demonstrate how Vue can be dropped into an existing project quite easily. The recipe could delve into a few different aspects of this use of Vue, including:

- How to initialise a Vue component from a CMS agnostic template DOM element
- How to add configuration via the DOM and inject config options into Vue
- How to structure Vue components (containers and presentational components)
- Configuring Vue in an HTML template build project

Let me know what you think! I'd be happy to contribute to this recipe 👍

sdras:
This is great! We'd love something like this. Thanks. Just tag me in the PR when you're ready :)

darkylmnx:
It would be nice to discuss the SEO issue also since blogs are content and SEO-driven

PR #1583, opened 2018-04-17
[Cookbook] Isolated apps in a CMS template

lewisvatcky:
@sdras

Cookbook recipe for #1521. Please let me know how it stacks up!


sdras:
Thanks! I should be able to look at it soon

PR #1587, opened 2018-04-19
Update class-and-style.md

byog:
I think this sentence
> This will only render the last value in the array which the browser supports

is directed at Vue and not above array in this article.


sdras:
Apologies, I'm not quite sure I understand what you mean by "is directed at Vue and not above array", but the way that it's written currently is more correct (just double checked in case). Feel free to reopen if I'm mistaken.

PR #1588, opened 2018-04-19
Update "cli" installation

gervaisb:
Update to install @vue/cli


sdras:
We're actually thinking about updating this page to just point to the readme for that repo, so that we have a single source of truth. Is that still the line of thinking here @chrisvfritz?

chrisvfritz:
Yeah, I just made that update. 🙂 Thanks for the PR though @gervaisb!

Issue #1581, opened 2018-04-15
Add tip/work around when use a debounce function in a component's watch, the debounce function needs to be defined in the created event


dtsao:
As discussed in the [forum](https://forum.vuejs.org/t/issues-with-vuejs-component-and-debounce/7224/15?u=davidt1154), a debounce function defined in a component's watch will not work as expected when there are multiple instances of the component. See [example](https://codesandbox.io/s/713y67rv6). The expected behavior is that the watch will fire the debounce function in all component instances, but it is only happening in one of the component instances.

The tip/work around is to define the debounce function in the component's created event instead of in the watch. See [example](https://codesandbox.io/s/w7lqpyxn88). With this change, the watch fires the debounce function in all component instances.

donnysim:
Technically everything works as it should. Every component has the same function instance but is called with different context, so what happens in this case is you're calling the same `debounce` function across all components but with different context. 

Short simplified example (changed debounce logic for simplification):

```js
watcher: function debouncedFn(userFn) {
  let wasCalled = false;
  
  if (!wasCalled) {
    userFn.call(this);
    wasCalled = true;
  }
}
```

first component that triggers `debouncedFn` will execute it with `debouncedFn.call(this, newValue, oldValue)` etc.. which will set the `wasCalled` to `true` for all other components that also use this watcher, so every other call will be ignored.

You can go around this by creating a debounced method:

```js
methods: {
  debouncedFn: _.debounce(function (newValue, oldValue) {
    // logic
  }),
},
```

and assign it to the watcher:

```js
watch: {
  watchThis: 'debouncedFn',
},
```

dtsao:
@donnysim, when I define the debouncedFn in a method instead of in the created event, the watch only updates one component instance. See [example](https://codesandbox.io/s/7y0mrx3k8x). 

This is unfortunately, the same behavior as in my original example, where the debouncedFn was defined in the watch. Unless I am missing something, it still looks like the debouncedFn needs to be defined in the created event in order for the watch to call the function in all component instances.

donnysim:
@dtsao oh right, you're completely correct, my brain forgot that methods are shared too 😅

NataliaTepluhina:
Closing as related PR was merged

PR #1600, opened 2018-04-26
Refactor debounce example to work with multiple instances

chrisvfritz:
Trying to demonstrate a debounce pattern that will work as expected for multiple simultaneous instances of a component (#1581). I want to get a few more eyes on this to make sure it's what we want to recommend.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -215,10 +215,10 @@ var watchExampleVM = new Vue({
     // whenever question changes, this function will run
     question: function (newQuestion, oldQuestion) {
       this.answer = 'Waiting for you to stop typing...'
-      this.getAnswer()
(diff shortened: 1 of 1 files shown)


LinusBorg:
Looking fine. Personally I like to simply replace the method with a debounce version of itself instead of creating a new prop on the vm, but maybe it's clearer your way as an example.

chrisvfritz:
Thanks everyone! 🙂

audable:
Found this issue while trying to understand what happens in the example. It makes very little sense to me as a beginner what happens in the line this.debouncedGetAnswer = _.debounce(this.getAnswer, 500)

I see no other examples in the guide assigning arbitrary properties to the vue instance.

Seeing the diff from the previous version i think clarity about the subject was completly lost here. I am lost at least.

PR #1602, opened 2018-04-27
Grammar Correction

4goodapp:


chrisvfritz:
Thanks for the PR. 🙂 I believe this is already correct though - and actually, I don't believe the suggested change of "when you want perform" is correct in English. Did the sentence, as it's currently written, create confusion for you?

4goodapp:
>  I don't believe the suggested change of "when you want perform" is correct in English.

Yeah, maybe, So feel free to remove it.

> Did the sentence, as it's currently written, create confusion for you?

Not really, Just though the one I suggest was better.

chrisvfritz:
Ah, OK. The other suggestion changes the meaning a little, so I'm going to leave it as it is actually. Thanks for the PR though. 🙂

PR #1604, opened 2018-04-28
[Cookbook] Handling Errors with Error Boundaries

### LINKED ISSUE #1595: [Cookbook] Creating and Using an ErrorBoundary Component
React 16 introduced a pattern of utilizing [ErrorBoundary](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html) components to handle UI errors that might otherwise break the application.  I touched on this idea in this [post](https://medium.com/@dillonchanis/handling-errors-in-vue-with-error-boundaries-91f6ead0093b) and think it might be a decent candidate for the cookbook.  I could reuse a lot of the examples within the post.

A rough outline:

**Base Example**
Create a simple component which throws an error during render.  Introduce `errorCaptured` hook and show how the hook will capture the error for us.  Extract the `errorCaptured` into its own component that wraps any other components and show how it still captures the errors and can render a different UI when an error occurs.

**Details about the Value**
Go into more detail about the `errorCaptured` hook (parameters, error propagation).  Details about why its a compelling pattern may include separation of error handling and UI display, code reuse, component composition vs. imperative code, granularity (individual widget vs. application error handling).

**Real-World Example**
Start off by rebuilding a more robust ErrorBoundary component that can take other components as props for fall back UIs, callback functions and emits error events.  Example could be similar to the one in my article where `v-for` over list results in an error and how we can use our new ErrorBoundary to remedy the situation.

**When To Avoid This Pattern**
The ErrorBoundary component will not work if it directly wraps functional components.  There are also only a few places the `errorCaptured` hook will catch errors.

**Alternative Patterns**
Since the ErrorBoundary component acts like `try/catch` we can use one in its place, especially in areas where the ErrorBoundary will not work.

Let me know if you think this would be a good candidate or not! Thanks!

dillonchanis:
@sdras 

Closes #1595 

I think this is a decent first draft.  I'm open for any changes of course, let me know when you get the chance.  Also, I'm totally unsure of the order! Looking forward to any feedback, thanks!

### DIFF
--- a/src/v2/cookbook/handling-errors-with-error-boundaries.md
+++ b/src/v2/cookbook/handling-errors-with-error-boundaries.md
@@ -1,0 +1,286 @@
+---
+title: Handling Errors with Error Boundaries
+type: cookbook
+order: 11
+---
+
+## Base Example
+
+Handling JavaScript errors in your Vue components is an integral part of building applications. In some cases, a JavaScript error can break a portion of the UI. This behavior can be avoided with the use of an "error boundary". An error boundary is a component that can catch JavaScript errors within child components and render a fallback UI.
+
+In Vue 2.5.0, the `errorCaptured` hook was introduced which can be used on Vue components. This [hook](https://vuejs.org/v2/api/#errorCaptured) will only capture errors thrown from any descendent components.
+
+```js
+export default {
+  name: 'ErrorBoundary',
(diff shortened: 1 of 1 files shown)


sdras:
Great! I will need a little bit here to take a look but first pass looks good!

PR #1606, opened 2018-05-01
Use properly variable scope declaration

animek66:
Maybe using a constant is a better idea? What you think?

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -42,7 +42,7 @@ At the core of Vue.js is a system that enables us to declaratively render data t
 </div>
 ```
 ``` js
-var app = new Vue({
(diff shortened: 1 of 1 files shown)


rnickson:
Browser support is about 97.61% globally, according to https://caniuse.com/#feat=const (looks like IE-only issue for me)

chrisvfritz:
Thanks, but this is very intentional. 🙂 Vue supports down to IE9 and `const` will not work in IE9 or IE10. Since our core docs don't assume a compilation step (e.g. with Babel), I feel it'd be irresponsible to demonstrate code that won't work in all the browsers we support.

Eventually though, we do plan to have a toggle for people who want to see examples that assume a build system.

PR #1607, opened 2018-05-01
Replace images with actual text/code

lipis:
- Much easier to copy/paste examples
- Easier to maintain

(If you guys OK with it, I could replace more images)


lipis:
Now I see why.. no colors :( 

https://deploy-preview-1607--vuejs.netlify.com/v2/guide/single-file-components.html

lipis:
Will experiment tomorrow.. maybe `html` would be OK

chrisvfritz:
The reason we made these code examples images was to demonstrate the syntax highlighting you can get in `.vue` files (which unfortunately isn't available through highlight.js). We haven't had a problem maintaining these (they essentially never have to change) and for these particular examples, I'm not sure it's as common for users to want to copy them, so I do appreciate the recommendation, but I think I have to close, unfortunately. Thanks for the PR! 🙂

PR #1609, opened 2018-05-03
Fixed expects for "is" prop for dynamic components

pranavjindal999:
`is` prop for dynamic components expects a globally registered name of component or component’s options object demonstrated in this [fiddle](https://jsfiddle.net/chrisvfritz/b2qj69o1/).


chrisvfritz:
Nice improvement. 🙂

PR #1628, opened 2018-05-14
Improve accessibility of code examples

chipit24:
Fixes #1624.

This PR is meant to update the code examples in the documentation to help promote good accessibility practises; the end result should not detract from the original intent of the code example.

This PR is currently a WIP and I will update it as I work through the docs.

PR #1630, opened 2018-05-15
Update todo-list-example for a11y

chipit24:
Continuation of #1628. This PR improves the accessibility of the todo list example. Again, the point of this PR is not to modify the original intent of the example, but to ensure it promotes good a11y practises.

The following article, which uses Vue.js in its examples, includes useful information about creating an a11y todo list component: https://inclusive-components.design/a-todo-list/.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -490,10 +490,11 @@ Here's a complete example of a simple todo list:
 ``` html
 <div id="todo-list-example">
   <form v-on:submit.prevent="addNewTodo">
+    <label for="new-todo">Add a todo</label>
     <input
-      aria-label="Write a new todo item"
-      placeholder="E.g. Feed the cat"
       v-model="newTodoText"
+      id="new-todo"
+      placeholder="E.g. Feed the cat"
     >
     <button>Add</button>
   </form>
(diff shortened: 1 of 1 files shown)


stephenbelyea:
Looks good @chipit24! Would consider more verbose "Add"/"Remove" button text on a real-world implementation, but this is sufficient for the sample context :)

src/v2/guide/list.md
```diff
@@ -489,11 +489,14 @@ Here's a complete example of a simple todo list:
...
 ``` html
 <div id="todo-list-example">
-  <input
-    v-model="newTodoText"
-    v-on:keyup.enter="addNewTodo"
-    placeholder="Add a todo"
-  >
+  <form v-on:submit.prevent="addNewTodo">
+    <input
+      aria-label="Write a new todo item"
```
sdras: After speaking with an accessibility expert, my understanding is that if we're going to include a placeholder, it would be obscured by this aria-label. It would probably be better to use a label element: `<label for="new-todo">Add a todo</label>` and add an id on the input, that way the placeholder will still be accessible.

src/v2/guide/list.md
```diff
@@ -489,11 +489,14 @@ Here's a complete example of a simple todo list:
...
+      placeholder="E.g. Feed the cat"
+      v-model="newTodoText"
```
sdras: I believe that we should keep the v-model as the first attribute, as that's what we're teaching here and it reduces the need for scanning on the part of the reader trying to learn.

src/v2/guide/list.md
chipit24: I agree with using a label. I also think a label element would be more familiar (and therefor less distracting) for readers than an aria-label attribute.

For the sake of my own understanding, would you know how the placeholder is obscured by aria-label?

sdras:
Awesome work, thank you!

PR #1632, opened 2018-05-16
Update components.md

artiomgiza:
Add clarification and full working example to one of the sections.

The reason that the previous examples and explanations uses simple variable (string):
```
props: ['title'] 
```
and current uses object variable:
```
props: ['post'] 
```
It can be not so clear from the explanations, but full example makes it very clear :)

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -210,26 +210,47 @@ That's all you need to know about props for now, but once you've finished readin
 When building out a `<blog-post>` component, your template will eventually contain more than just the title:
 
 ```html
-<h3>{{ post.title }}</h3>
+<h3>{{ title }}</h3>
 ```
 
 At the very least, you'll want to include the post's content:
 
 ```html
-<h3>{{ post.title }}</h3>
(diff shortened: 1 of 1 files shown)


src/v2/guide/components.md
```diff
@@ -229,6 +229,20 @@ If you try this in your template however, Vue will show an error, explaining tha
 </div>
 ```
 
+Here we pass entire post object to the `props` option instead of separately pass title and content:
```
chipit24: Adjusting the grammar, I think this reads better:

> Here, we pass the entire post object to the `props` option instead of separately passing the title and content.

artiomgiza: agree, aligned with your advice :)

artiomgiza:
@chrisvfritz, your proposal feels MUCH better than what I did before. Could not think of better explanation!

chrisvfritz:
Aw, thanks. 😊 I really appreciate you laying the foundation for this! It wouldn't have happened with you. 🙂

PR #1633, opened 2018-05-17
fixed message test

ivansieder:
Passing the 'message' attribute in the updated tests was missing.
Passing no object would result in the 'data' property being empty, which in turn wouldn't render 'Welcome to the Vue.js cookbook'.


ivansieder:
Damn, that was my error, of course the message is already on the instance, as that's per default in the component itself.

PR #1646, opened 2018-05-23
Reorganize component props to introduce prop types earlier, fixes #1644

### LINKED ISSUE #1644: prop with no value is misleading
[The docs](https://github.com/vuejs/vuejs.org/blob/master/src/v2/guide/components-props.md#passing-a-boolean) specifies the following:
```HTML
<!-- Including the prop with no value will imply `true`. -->
<blog-post favorited></blog-post>
```

In my brain this tells me to do the following:

```HTML
<blog-post favorited></blog-post>
```

```Javascript
export default {
    name: 'blog-post',

    props: {
        favorited: false,
    }
}
```

This results in the prop `favorited `equals to `""` == empty string. Not `true`

I know that in order to get this to work, one has to declare the prop of type boolean instead:
```Javascript
export default {
    name: 'blog-post',

    props: {
        favorited: Boolean,
    }
}
```

One could argue that the title of the doc snippet; "Passing a Boolean"  should be enough for the end user to "know what to do". However, for people new to vue, I think this is expecting too much. 🤔

chrisvfritz:
@BlitZz96 brought up a good point in [this issue](https://github.com/vuejs/vuejs.org/issues/1644) about users lacking the necessary background knowledge to fully understand Passing Static and Dynamic Props, at least as it relates to booleans and potentially other types in the future. To fix this, I added a new Prop Types section right before it.

### DIFF
--- a/src/v2/guide/components-props.md
+++ b/src/v2/guide/components-props.md
@@ -25,7 +25,29 @@ Vue.component('blog-post', {
 
 Again, if you're using string templates, this limitation does not apply.
 
-## Static and Dynamic Props
+## Prop Types
+
+So far, we've only seen props listed as an array of strings:
+
+```js
+props: ['title', 'likes', 'isPublished', 'commentIds', 'author']
+```
+
+Usually though, you'll want every prop to be a specific type of value. In these cases, you can list props as an object, where the properties' names and values contain the prop names and types, respectively:
+
+```js
+props: {
+  title: String,
+  likes: Number,
+  isPublished: Boolean,
+  commentIds: Array,
+  author: Object
+}
+```
+
+This not only documents your component, but will also warn users in the browser's JavaScript console if they pass the wrong type. You'll learn much more about [type checks and other prop validations](#Prop-Validation) further down this page.
+
+## Passing Static or Dynamic Props
 
 So far, you've seen props passed a static value, like in:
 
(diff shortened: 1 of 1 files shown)


chrisvfritz:
@BlitZz96 I can't tag you for review, but I'd love to hear your thoughts.

chrisvfritz:
Glad to hear it! Thanks again for pointing out the problem. 🙂 Now one more review from a core team member and I'll merge this.

src/v2/guide/components-props.md
```diff
@@ -25,7 +25,29 @@ Vue.component('blog-post', {
...
-## Static and Dynamic Props
+## Prop Types
+
+So far, we've only seen props listed as an array of strings:
+
+```js
+props: ['title', 'likes', 'isPublished', 'commentIds', 'author']
+```
+
+Usually though, you'll want every prop to be a specific type of value. In these cases, you can list props as an object, where the properties' names and values contain the prop names and types, respectively:
```
phanan: Isn't it "to be _of_ a specific type of value?"

eddyerburgh: I'm not sure on grammatical correctness, but "to be a" is more common than "to be of a". The sentence reads fine to me.

phanan: Yeah, not a native speaker myself, just drawing my inspiration from the infamous "Argument of type 'string | number' is not assignable" error. Then again, it's very minor and can be safely ignored if need be.

chrisvfritz: @phanan I think if we didn't have "of value" at the end I agree that "of type" would sound better, but as it is sounds good to me. 🙂 Thanks for the review!

PR #1647, opened 2018-05-24
grammarly correction

zzzgit:


chrisvfritz:
Thanks, but this is already correct actually. 🙂

PR #1649, opened 2018-05-24
A fix in example

keligijus:
When creating a component with the returned constructor an error is thrown `TypeError: Dp is not a constructor`. Passing an empty options object seems to fix this.


keligijus:
Here's a fiddle to test/prove https://jsfiddle.net/h8n88bz7/

chrisvfritz:
Note in the comment:

> retrieve a registered component (always return constructor)

So it doesn't register a component - it _retrieves_ a component that's already registered. If the component hasn't yet been registered, it won't be able to retrieve anything.

PR #1663, opened 2018-06-01
Update debugging-in-vscode.md

Alex-Sokolov:
Add note how to set source maps with Vue CLI 3


sdras:
This is great! I've also talked with the VS Code team, they're going to update their cross post as well. Thanks!

PR #1664, opened 2018-06-02
Added details about Lifecycle Hooks

Knogobert:
Mostly a clarification for beginners without prior knowledge about what Lifecycle hooks are. Makes it easier to understand when perhaps only knowing that the developer has a `mounted()` hook. 

I left out the function syntax because of generality. Add if it makes more sense to add. I am not fully knowledgable in Vue yet.


chrisvfritz:
Thanks! Great addition. 🙂

Issue #1672, opened 2018-06-06
Class and style page in classObject, active should be string


goelrohan6:
active is not a defined variable, it should be a string

![image](https://user-images.githubusercontent.com/14099191/41021730-14e6b1d4-6985-11e8-921d-709815cc920d.png)

goelrohan6:
#1673 I have created a pr, can someone please review :)

goelrohan6:
Yes @leopiccionia it can be written as [active] but the point is there is no variable active defined as 'some class'.

chrisvfritz:
Thanks, but the quotes are indeed unnecessary. Either will work, so we've opted for the shorter syntax. 🙂

PR #1673, opened 2018-06-06
fix vuejs/vuejs.org#1672

goelrohan6:


chrisvfritz:
Thanks, but this isn't actually necessary. See my comment in the issue for more info. 🙂

PR #1675, opened 2018-06-06
Add keyboard support for main navigation flyout menus

kentyunge:
I noticed that the main navigation on the vuejs.org site could not be accessed via a keyboard only.  This causes accessibility issues for a variety of people.  In this PR I attempted to remedy that by adding the following:

- `tabindex=0` to the flyout parent elements
- `aria-haspopup` and `aria-expanded` attributes to menu
- using `up` and `down` keyboard keys allow user to navigate through menu items
- using `tab` key allows user to exit flyout menu

I was unable to test on IE/Edge.  Additionally, this was a best guess on how a keyboard user would navigate the navigation items and could benefit from a review from someone more versed in ADA best practices.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,8 +1,7 @@
 {
   "name": "vuejs.org",
-  "version": "1.0.18",
-  "lockfileVersion": 1,
   "requires": true,
+  "lockfileVersion": 1,
   "dependencies": {
     "JSONStream": {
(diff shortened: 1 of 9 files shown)


kentyunge:
Update: Was able to test on Win 10 with IE11/Edge and made a few modifications to make it work in those browsers.

PR #1676, opened 2018-06-08
Add alt text to images to help with screen reader accessibility

kentyunge:
Many images on the vuejs.org site do not have alt text associated with them making it difficult for screen reader users.  I have added support to the main pages as well as the sponsor text.  I was unable to add support to the "Open Collective" sponsors as they are configured on the server-side and not in the config.yml file.


themes/vue/_config.yml
```diff
@@ -15,61 +15,88 @@ special_sponsors:
 platinum_sponsors:
   - url: http://tooltwist.com/
     img: tooltwist.png
+    name: tool twist
```
yyx990803: this should be "tooltwist"

yyx990803:
@chrisvfritz LGTM once you verify it renders correctly in the deploy preview.

Issue #1677, opened 2018-06-09
the grid-template example could be made sortable via the keyboard easily


YordanGeorgiev:
Sorry for not doing this the official way ... not ready to get familiar with the whole process , install proj etc. , yet the grid-component could be done sortable via the keyboard by adding the ` tabindex="1" v-on:keyup.enter="sortBy(key)"` : 

```
      <!-- component template -->
      <script type="text/x-template" id="grid-template">
        <table tabindex="1">
          <thead>
            <tr>
              <th v-for="key in columns"
                @click="sortBy(key)"
                :class="{ active: sortKey == key }" tabindex="1" v-on:keyup.enter="sortBy(key)">
                {{ key }}
                <span class="arrow" :class="sortOrders[key] > 0 ? 'asc' : 'dsc'">
                </span>
              </th>
            </tr>
          </thead>
```

Of course the following css could be added tooo ... 
```
th:focus {
   color: #fff;
}
```

In this way both the whole table is navigable with the arrows , and each column is sortable by hitting the Enter ...

YordanGeorgiev:
Also sorting does not properly work for floats : 
```
           a = a[sortKey]
           b = b[sortKey]
          return (a === b ? 0 : a > b ? 1 : -1) * order
```
Change to : 
```
          if ( (!isNaN(parseFloat(a)) && isFinite(a)) && (!isNaN(parseFloat(b)) && isFinite(b)) ) {
            return (a-b) * order
          } else {
            return (a === b ? 0 : a > b ? 1 : -1) * order
          }
```

seems to fix it ...

YordanGeorgiev:
https://jsfiddle.net/YordanGeorgiev/3bpkyc1v/9/

phanan:
Thanks for the PR. However, while I would consider myself a big proponent of accessibility, I believe this change would fall beyond the scope of the example and make it more complex than it should be.

PR #1680, opened 2018-06-10
add sorting for floats in grid. add sorting by keyboard in grid (#1677)

YordanGeorgiev:
https://github.com/vuejs/vuejs.org/issues/1677


chrisvfritz:
Thanks! I like the upgrades you made. 🙂 We'd like to avoid making the basic examples _too_ complex though, which I know is very subjective, but we have to draw the line somewhere. 😄 I think the cookbook would be a great place for a more detailed example though!

PR #1690, opened 2018-06-16
Fix code style in Client-Side Storage recipe

lex111:


src/v2/cookbook/client-side-storage.md
```diff
@@ -140,13 +146,16 @@ const app = new Vue({
   methods: {
     addCat() {
       // ensure they actually typed something
-      if(!this.newCat) return;
+      if (!this.newCat) {
```
sdras: I'm not sure this one needs to be fixed this way, maybe @chrisvfritz can sound off on it

sdras:
Thank you for doing this! It was on my list to take care of. I just need a response from Chris on that one part and I can merge it in :)

lex111:
As they say in many manuals on clean code and just good books, it is better to always put curly braces, even if only one expression :smiley:  Readability and easier to maintain - priceless. But the last word, of course, is yours.

sdras:
You know, I don't care enough about it to stall this PR, it might be nicer to read for newcomers and we can always update if necessary.

PR #1691, opened 2018-06-16
Add radio button custom `v-model` example

chriscalo:
Radio buttons are significantly more difficult than checkboxes to to wrap in a component. Here's a proposal for a minimal example to help people trying to get this right. Thoughts?

### DIFF
--- a/src/v2/guide/components-custom-events.md
+++ b/src/v2/guide/components-custom-events.md
@@ -59,6 +59,36 @@ the value of `lovingVue` will be passed to the `checked` prop. The `lovingVue` p
 
 <p class="tip">Note that you still have to declare the <code>checked</code> prop in component's <code>props</code> option.</p>
 
+Radio buttons are a bit different. A radio button should be checked when its `value` matches the `v-model`.
+
+```js
+Vue.component('base-radiobutton', {
+  model: {
+    prop: 'modelValue',
+  },
+  props: {
+    value: String,
+    modelValue: String,
+  },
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I'm glad you brought up this use case. 🙂 I think one significant difference between a checkbox and a radio button is that a checkbox can stand on its own (e.g. "I agree to the terms"), whereas a radio button only makes sense in a group (e.g. "Yes" and "No").

So when wrapping a radio button with a component, I wouldn't actually wrap _individual_ radios, but rather the entire group, like in [this example](https://codepen.io/chrisvfritz/pen/rKZrgb?editors=1010). This isn't really specific to Vue, so I'm hesitant to add another example here, but I think it might make a good cookbook recipe!

chriscalo:
(No problems with closing this PR, just have a follow-up question.)

I completely agree there's value in having a RadioButtonGroup component. The specific situation that necessitates having individual RadioButton components is when you want to inline additional supporting content _between_ the radio items.

See this example:
https://codesandbox.io/s/p30mn66y2q

Is there a better way to achieve this that doesn't necessitate understanding what I put into this PR?

chrisvfritz:
The way you've solved it looks good. 🙂 I believe the only additional knowledge introduced in the PR though is how radio inputs work, which isn't specific to Vue. Does that make sense?

chriscalo:
Makes perfect sense. Thanks, Chris!

PR #1692, opened 2018-06-19
Update transitions.md

wozaihouma:
add missing "list-complete-move css" in "list-complete" case


wozaihouma:
例子已经把 transition 从 active 转到元素上了 ，看起来已经包含了 -move

PR #1693, opened 2018-06-20
Update instance.md

giova2:
I've been noticed that the vue lifecycle was not anymore, so I am trying to put it there again


chrisvfritz:
There may have been a temporary caching issue that prevented the image from loading and it seems to be working now, so closing for now, but please do let me know if you're still not seeing the image. 🙂

PR #1695, opened 2018-06-25
add kebab-case to the sentence

sir-kain:
use the kebab-case term to better explain this situation


chrisvfritz:
The name does not actually have to be kebab-case - it should just match the name being used to reference that component in the template (e.g. `my-component`, `MyComponent`, or even `myComponent` would technically work). I'm happy to reopen if you still feel some clarification is needed though. 🙂

PR #1699, opened 2018-06-27
Improvements recipe of Create a CMS-Powered Blog

lex111:


Alex-Sokolov:
Maybe also change this:
```js
data() {
  return {
    post: {}
  }
},
```

to this:
```js
data() {
  return {
    post: null
  }
},
```

Empty object often not useful in applications & more complicated to track if object is empty

lex111:
@Alex-Sokolov the `v-for` directive requires an array, is not it?

Alex-Sokolov:
@lex111 No, if we change empty object to null, also we need to add v-if for checking, and until post loaded show loading placeholder or smth like that

lex111:
friendly ping @sdras

sdras:
Much better formatting, thank you for this.

PR #1709, opened 2018-07-13
Update transitions.md

byog:
More correct description for 'v-leave'


chrisvfritz:
I'm not sure this is accurate actually. If it was removed one frame _after the element was removed_, wouldn't that mean it's never removed? There'd be no element to remove it from at that point, right?

PR #1710, opened 2018-07-16
complete component's code

SimoralX:
if someone don't add this code,the component will not work.Thanks!
```
var app7-1 = new Vue({
    el: '#app-7-1'
});
```


chrisvfritz:
This is actually intentionally a partial example, with the full code appearing soon afterwards. 🙂

PR #1711, opened 2018-07-16
[Doc EN]: Remove trailing + correct return line + Vue Convention implementation into `cookbook`

MachinisteWeb:
Hi @sdras and @chrisvfritz!

I think this one is easy to merge ;)


Haeresis:
\o/

PR #1717, opened 2018-07-20
Fix last sponsor image link URL 404 - Prevent loss of sponsorship

roybarber:
The list of sponsors, Open Collectives last sponsor image gives a 'Not found' on click, instead we should direct users through to the Open Collective vue landing page, where they can now become a sponsor, preventing loss of sponsorship.
Since its all hard coded we have to replace the URL. Another option would be to reduce the loop numbers to remove the last image and add a static button afterwards.

### DIFF
--- a/themes/vue/layout/partials/sponsors.ejs
+++ b/themes/vue/layout/partials/sponsors.ejs
@@ -34,10 +34,18 @@
 
 <script>
 window.addEventListener('load', function () {
+  const searchUrl = [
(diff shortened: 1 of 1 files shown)


sdras:
Thanks for the PR! However, I'm not entirely sure this is necessary- I can't duplicate the 404 and the sponsor links and button functionality is in order. Feel free to reopen if I'm mistaken.

PR #1718, opened 2018-07-21
Fix vulnerabilities

sdras:
- We were getting a warning about a vulnerability in hoek, so updated I updated it to fix `npm i hoek` so it would dynamically update all instances.
- In the process, saw that we had a few more vulnerability warnings, so ran `npm audit fix` to update 9 other high vulnerabilities- (68 reported in all, low through high)


sdras:
Ok, this should do it. But please note that I'm also adding in node engine details here in order to get this done.

chrisvfritz:
Looks good to me! Thanks. 🙂

PR #1731, opened 2018-07-31
The keynote Github link has changed.

namanshenoy:
https://github.com/znck/vue-keynote seems to be the old github repo.
New one is: https://github.com/znck/keynote


chrisvfritz:
Thanks!

PR #1732, opened 2018-07-31
dockerize-vuejs-app: rewrite nginx in small letters

lex111:
Because nginx as the name of the server is written in small letters.


sdras:
Hmmm, looks like NGNIX is how they refer to themselves on [their site](https://www.nginx.com/), which is what we usually go by. I don't think we'll make this change.

lex111:
The documentation uses generally naming in lowercase, it looks more beautiful than writing in big letters, IMHO.

image

PR #1733, opened 2018-07-31
Proposal: change mutation name style

galvez:
The original text uses `REMOVE_TODO` (uppercase constant style, reminiscent of React and Redux) as mutation name in an example. I propose that it changes to `todoRemoved`, a style that will match the rest of the code, while maintaining a clear distinction from related actions.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1940,13 +1940,13 @@ export default {
     list: []
   },
   mutations: {
-    REMOVE_TODO (state, todoId) {
+    todoRemoved (state, todoId) {
       state.list = state.list.filter(todo => todo.id !== todoId)
     }
   },
   actions: {
     removeTodo ({ commit, state }, todo) {
-      commit('REMOVE_TODO', todo.id)
+      commit('todoRemoved', todo.id)
     }
   }
 }


chrisvfritz:
We're actually inconsistent about this in our current examples - sometimes using UPPER_CASE, sometimes camelCase. We're discussing it internally and right now, are leaning towards consolidating on camelCase for mutation names, which is more generally the standard for function names in JavaScript.

galvez:
@chrisvfritz awesome, very glad to hear! -- should I keep this PR open then? :)

galvez:
@chrisvfritz I'll insist then on my proposal for *past-tense mutations* then. I think having both `deleteTodo` and `removeTodo` would be weird and would contribute to confusion. Having past-tense mutations (`todoRemoved`) offers a clear and easy distinction between mutations and actions. We've adopted this standard in a number of projects at work and it's been pretty manageable.

I'll leave the PR open if this suggestion stands a chance of consideration.

chrisvfritz:
@galvez I think the chances are relatively low, but it's not impossible at this point. Personally, if I see something called `todoRemoved`, I'd think it was either a boolean to tell me whether or not a todo was removed, _or_ an event to log that the todo was removed. In the context of `commit('todoRemoved', todoId)`, the todo is _not_ technically removed yet, so we're calling the function to make that happen rather than reporting that it was already done.

Also more generally in JavaScript, imperative-style verbs are almost universally used for called functions, so we'd be breaking a big convention of the larger ecosystem.

Does that make sense? Also, I'd love to hear more about the confusions that the past-tense convention resolved for you. I'm wondering if there may be a 3rd option or additional convention that could resolve all our concerns. 🙂

chrisvfritz:
@galvez I think the chances are relatively low, but it's not impossible at this point. Personally, if I see something called `todoRemoved`, I'd think it was either a boolean to tell me whether or not a todo was removed, _or_ an event to log that the todo was removed. In the context of `commit('todoRemoved', todoId)`, the todo is _not_ technically removed yet, so we're calling the function to make that happen rather than reporting that it was already done.

Also more generally in JavaScript, imperative-style verbs are almost universally used for called functions, so we'd be breaking a big convention of the larger ecosystem.

Does that make sense? Also, I'd love to hear more about the confusions that the past-tense convention resolved for you. I'm wondering if there may be a 3rd option or additional convention that could resolve all our concerns. 🙂

galvez:
> we'd be breaking a big convention of the larger ecosystem.

That's a valid point. Adopting past tense for mutations also required creating the notions that mutations are really **special** functions. Since their purpose is to provoke actual state change, naming them in past tense simply created a way to identify them by interpreting them in a *_what happened after this ran_* narrative. It made sense enough to me, but I understand it may confuse some. Other ideas that come to my mind would involve even more complex and redundant semantics.

galvez:
One small thing: the past-tense _narrative_ also encourages data wrangling logic to happen outside the mutation, keeping them pure.

chrisvfritz:
> Since their purpose is to provoke actual state change, past tense simply created a way to identify them by interpreting them in a what happened after this ran narrative.

That makes sense. 🙂 It's what I was attempting to achieve with the suggestion to describe the specific nature of the mutation, but it sounds like it's not quite sufficient to avoid confusion. 😅 Do you think it would help to define an exhaustive list of verbs to recommend that all mutations begin with?

Browsing through my own apps, I believe every mutation could begin with one of the verbs below:

- __`push`__: when using `Array.prototype.push`.

- __`pop`__: when using `Array.prototype.pop`.

- __`unshift`__: when using `Array.prototype.unshift`.

- __`shift`__: when using `Array.prototype.shift`.

- __`reverse`__: when using `Array.prototype.reverse`.

- __`copyWithin`__: when using `Array.prototype.copyWithin`.

- __`fill`__: when using `Array.prototype.fill` or otherwise setting static or non-deterministic values for 2 or more items in an array.

- __`sort`__: when using `Array.prototype.sort` or otherwise sorting the items in an array (e.g. with lodash's `sortBy` or `orderBy`).

- __`delete`__: when removing 1 or more items from an array or object, whether using `Array.prototype.splice`, `Array.prototype.filter`, `Vue.delete`, or some other strategy.

- __`set`__: for all other cases using `=` to set a new value for an existing property, or `Vue.set` to create/set a property.

As a bonus, this strategy would also help people limit the scope of their mutations, keeping all business logic in actions where it belongs. In the apps I have access to, I did notice some cases where none of the above verbs were appropriate - but in every case, the mutation was doing much more than it should have.

And as _another_ bonus, this convention could speed up development with snippets and we might even be able to automate its enforcement in eslint-plugin-vue! 😻 (@michalsnik, do you think that might be feasible?)

What are your thoughts? In the teams you've worked with, do you think this would clarify the responsibilities of mutations for and reduce the cognitive overhead of naming them?

galvez:
@chrisvfritz No reason we can't do both, I guess? How about we add a paragraph as follow to the docs:

galvez:
```
When naming mutations, a good strategy is to mimic the list operations the mutation mainly performs, e.g., `pushUser`, `unshiftUser`. If that is too limiting an approach, you can also adopt a past-tense narrative style, where a mutation is named after the changes it provokes, e.g., `userAdded`, `userRemoved`. Either way promotes keeping mutations pure, with data wrangling logic best kept in actions.
```

I added it to the bottom of `style-guide/index.md`, but feel free to edit and relocate it if you think it's a good idea.

michalsnik:
I like this idea @chrisvfritz !

However in terms of eslint rule that might check it - I'm not so sure. Depending on a person or team, structure of the vuex module might differ drastically. Some might keep it in `store` folder, some not. One team prefers to have single file modules, other not. Same with keeping mutation names as constants in a separate file. I don't see how - except explicit comments detect whether given `mutation` is indeed a `mutation` using just an AST and filename from the context. In the end it might probably introduce too many false positives as well as false negatives.

I think that this kind of rule might be feasible but only per-project case as a team's internal agreement, so as from our plugin's perspective I must say: `no, it's not feasible`, unless I missed something.

manniL:
Maybe we should move this to https://github.com/vuejs/vuex? 🤔

tatoon187 (commented):
http-push

PR #1736, opened 2018-08-01
changed kebabCase to kebab-case

amadeann:
I think it makes more sense to show visually what is actually created. In the current form, it is somewhat confusing.


sdras:
Makes sense.

PR #1740, opened 2018-08-07
Add Firefox to Debugging in VS Code recipe

caugner:


phanan:
Personally, I feel it clumsy to add " / Firefox" everywhere. Maybe some `p.tip` at the beginning of the recipe works better.

sdras:
I agree with @phanan, I think it would be better to have one single declaration that it's both browsers and avoid the repetition, except where it's necessary to show two ways of working. Let me know if you need some more detailed suggestions on phrasing :)

caugner:
@sdras Thanks for the review. I have just updated the recipe, avoiding (or shortening) the repetition where possible.

Let me know if the PR needs further adjustments.

sdras:
Nicely done! Looks very clean. 👍Thank you @caugner.

PR #1741, opened 2018-08-08
Explicitly specify arguments

james-brndwgn:
Destructuring "args" throws an error. It is not clear that you need to destructure "arguments" therefore changing this to ...arguments is more explicit.


Justineo:
Thanks. It's a function signature, not real code. `args` is just the name of the parameter and has nothing to do with `arguments`.

PR #1745, opened 2018-08-10
Clarify that custom events do not bubble

amcsi:
This may be a bit wordy, but I remember when I was first working with VueJS, it confused me that custom events didn't bubble up. I thought I could listen to custom events on any grandparent element.

With these changes, I would like to have this pointed out to people if I'm not alone on this.

The text I used might be a bit wordy, so feel free to suggest changes.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -340,6 +340,9 @@ Then on our blog post, we can listen for this event with `v-on`, just as we woul
 ></blog-post>
 ```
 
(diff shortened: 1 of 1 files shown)


amcsi:
Would someone please provide some feedback about what they think about this PR?

sdras:
Thanks @amcsi! Sorry about the delay.

I think this a great addition! But where it is right now breaks up the flow of the tutorial, it really belongs in the caveats section at the bottom of the page. Here are the things we would need to accept the PR:

- Moved to the caveats section
- Take it out of a tip
- It would be good to mention `$listeners` as an alternative (let me know if you need an example, or I can follow it up with another PR to clarify)

If this all seems like too much to do after all this time, totally understandable, just ping me here and I'll close it out and open a new one.

amcsi:
@sdras thanks for the reply! That's fair enough. I don't think I will have time to do it any time soon, so feel free to close this PR for now.

PR #1747, opened 2018-08-11
Homepage – Fix GitHub Button's icon misalignment

Nirlah:


sdras:
Hi! Thanks for the PR. On review, I don't think this is actually misaligned, and we would also want it to be consistent with the left margin for visual consistency.

PR #1749, opened 2018-08-13
Use of computed property instead of watch method

awerlogus:
I think, use of computed property is more expressive, and if you're going to commit this pull request, please, don't forget to change the article part bellow the code and the codepen version of the code.


sdras:
I think we'll leave it this way, it's not _incorrect_ to use lifecycle and a watcher for this. I might think differently about it if the PR was more complete. Pinging @cfjedimaster to see if he would like to update it, and the subsequent demos. I'm open to reconsidering, but I'll close it for now.

cfjedimaster:
I'm sorry for the delay - but I don't think it's worth the change. (To be clear, not a bad change it just doesn't seem like it's critical enough?)

PR #1750, opened 2018-08-13
Added example of using $nextTick with async await

hitautodestruct:
Added an example of using $nextTick with the async await syntax

### DIFF
--- a/src/v2/guide/reactivity.md
+++ b/src/v2/guide/reactivity.md
@@ -118,3 +118,18 @@ Vue.component('example', {
   }
 })
 ```
+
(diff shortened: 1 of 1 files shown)


sdras:
Nice! This is going to be helpful to people. Thanks.

src/v2/guide/reactivity.md
```diff
@@ -118,3 +118,18 @@ Vue.component('example', {
   }
 })
 ```
+
+Since `$nextTick()` returns a promise, you can achieve the same as the above using the new [ES7 async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) syntax:
```
phanan: AFAIK there's no "ES7." The official name is "ES2016," which I'd suggest using instead.

phanan: Another PR should be fine I guess :)

Issue #1698, opened 2018-06-26
[Cookbook] Practical use of scoped slots with GoogleMaps


mateuszRybczonek:
The base idea is to make it more clear for the developers how/why we can use scoped slots. As I noticed the idea behind the scoped slot is know for many devs, but there is a lack of practical examples on how we could use it. This example helped me and few other persons to understand this concept deeper, so I would like to share it with others :)

Details about the Value
This example shows a bit more clearly how we can use the scoped slots to create a wrapper that initialises a specific object based on external API and expose that object, that is pre-configured and ready to be used.

Real-World Example:
- https://github.com/mateuszRybczonek/v-tracker-vue/blob/master/src/components/GoogleMapLoader.vue
- https://github.com/mateuszRybczonek/v-tracker-vue/blob/master/src/components/VesselDashboardGoogleMap.vue

The example extracts the map initialisation and configuration to a separate component and return the `google` and `map` that is ready to be used in the template to inject any further object into the map (Marker / Line, etc.)

When To Avoid This Pattern:
It might be tempting to create a very complex solutions based on the example, but at some point we can get to the situation where this abstraction becomes an independent part of the code living in our codebase. If we get to that point it might be worth considering extraction to an add-on.

sdras:
This is s a great use of scoped slots! We'd love an entry like this.

NataliaTepluhina:
Closing this as related PR was merged

PR #1754, opened 2018-08-14
[Cookbook] Practical use of scoped slots with GoogleMaps

mateuszRybczonek:
@sdras a PR related to https://github.com/vuejs/vuejs.org/issues/1698

### DIFF
--- a/src/v2/cookbook/practical-use-of-scoped-slots.md
+++ b/src/v2/cookbook/practical-use-of-scoped-slots.md
@@ -67,6 +67,8 @@ export default {
 };
 ```
 
+This is just part of a working example, you can find the whole example in the Codesandbox below.
+
 <iframe src="https://codesandbox.io/embed/1o45zvxk0q" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
 
 ## Real-World Example: Creating a Google Map Loader component
@@ -117,8 +119,7 @@ export default {
   },
 ```
 
-- on `mounted` hook we create an instance of `GoogleMapsApi` and `Map` object from the `GoogleMapsApi`
-- we set the values of `google` and `map` to the created instances
+- on `mounted` hook we create an instance of `GoogleMapsApi` and `Map` object from the `GoogleMapsApi` and we set the values of `google` and `map` to the created instances
 
 ```js
   async mounted() {
@@ -139,7 +140,7 @@ export default {
 };
 ```
 
-So far so good, with that done we could continue adding the other objects to the map (Markers, Lines, etc.) and using it as a ordinary map component. But we want to use our `GoogleMapLoader` component only as a loader that prepares the map, not renders anything on it.
+So far so good, with that done we could continue adding the other objects to the map (Markers, Polylines, etc.) and using it as a ordinary map component. But we want to use our `GoogleMapLoader` component only as a loader that prepares the map, not renders anything on it.
 
 To achieve that we need to allow parent component that will use our `GoogleMapLoader` to access `this.google` and `this.map` that are set inside the `GoogleMapLoader` component. That's where `scoped slots` really shine. Scoped slots allow us to expose the properties set in a child component to the parent component. It may sound like an inception, but bear with me one more minute.
 
@@ -151,18 +152,15 @@ In the template, we render the `GoogleMapLoader` component and pass props that a
 
 ```html
 <template>
-  <div class="google-map">
-    <GoogleMapLoader
-      class="google-map__map"
-      :mapConfig="mapConfig"
-      mapHeight="460px"
-      apiKey="yourApiKey"
-    />
-  </div>
+  <GoogleMapLoader
+    :mapConfig="mapConfig"
+    mapHeight="460px"
+    apiKey="yourApiKey"
+  />
 </template>
 ```
 
-Our script tag looks like this at this stage
+Our script tag will look like below at this stage.
 
 ```js
 <script>
@@ -220,7 +218,7 @@ So finally we can add a scoped slot that will do the job and allow us to access 
 </template>
 ```
 
-Now when we have the slot in the child component we need to receive and consume the exposed object in the parent component.
+Now when we have the slot in the child component we need to receive and consume the exposed props in the parent component.
 
 ### Receive exposed props in the parent component using `slot-scope` attribute.
 
@@ -230,7 +228,6 @@ To receive the props in the parent component we declare a template element and u
 
 ```html
 <GoogleMapLoader
-  class="google-map__map"
   :mapConfig="mapConfig"
   mapHeight="460px"
   apiKey="yourApiKey"
@@ -250,7 +247,7 @@ Scoped slots allow us to pass a template to the slot instead of passing a render
 
 ### Create factory components for Markers and Polylines
 
-We will now create two factory components that will be used to create elements in the `VesselsGoogleMap`
+Now when we have our map ready we will create two factory components that will be used to add elements to the `VesselsGoogleMap`.
 
 `GoogleMapMarker.vue`
 
@@ -319,7 +316,7 @@ export default {
 };
 ```
 
-- both of them receive `google` from which we extract required object (Marker or Polyline) and `map` which gives as a reference to the map on which we want to place our elements.
+- both of them receive `google` from which we extract required object (Marker or Polyline) and `map` which gives as a reference to the map on which we want to place our element.
 
 - each receive also a prop with data required to create a corresponding element.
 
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,410 @@
...
+
+### Add elements to map
+
+Now let's use our factory components to add elements to our map.
+
+`VesselsGoogleMap.vue`
+
+To add elements to our map we render the factory component and pass the `google` and `map` objects.
+
+We also need to provide data required by the element itself (in our case `marker` object with position of the marker, `path` object with polyline coordinates.
```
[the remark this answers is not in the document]
michalsnik: `and` keyword would probably sound a bit nicer: `of the marker and ...`. WDYT?

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,410 @@
...
+export default {
+  components: {
+    GoogleMapLoader,
+    GoogleMapMarker,
+    GoogleMapLine
+  },
+
+  data () {
+    return {
+      mapSettings,
```
michalsnik: `mapSettings` in data seems to be redundant, isn't it?

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,410 @@
...
+      apiKey: this.apiKey
+    });
+    this.google = googleMapApi;
+    this.initializeMap();
+  },
+
+  methods: {
+    initializeMap() {
+      const mapContainer = this.$el.querySelector("#map");
+      const { Map } = this.google.maps;
```
michalsnik: I'd be careful about destructuring `Map` as this is also a native constructor in ES6. I think we shouldn't avoid shadowing these kind of objects as it might have unintended consequences and might introduce confusion. Do you agree?

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,410 @@
...
+        ...mapSettings,
+        center: { lat: 0, lng: 0 }
+      }
+    },
+  },
+}
+</script>
+```
+
+Still no scoped slots yet. Let's add one.
```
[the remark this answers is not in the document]
mateuszRybczonek: yes, something is definitely wrong in this sentence. 
Maybe `Still no scoped slots, let's add one.` will be better.

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,410 @@
...
+order: 11
+---
+
+## Base Example
+
+There are situations when you want the template inside the slot to be able to access data from the child component that is responsible for rendering the slot content. This is particularly useful when you need freedom in creating custom templates that use the child component's data properties. That is a typical use case for scoped slots.
+
+Imagine a component that configures and prepares an external API to be used in another component, but is not tightly coupled with any specific template. Such a component could then be reused in multiple places rendering different templates but using the same base object with specific API.
+
+We'll create a component (`GoogleMapLoader.vue`) that initializes `Google Maps API`, creates a `google` and `map` objects and exposes those objects to the parent component in which the `GoogleMapLoader` is used.
```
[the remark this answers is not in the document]
mateuszRybczonek: `and exposes those objects to its potential children in the template of a parent component in which the `GoogleMapLoader` is used`

I think we should additionally point out that those children must exist in the parent template.

mateuszRybczonek:
@sdras 

Would you be able to take a look anytime soon? :)

sdras:
Yes, sorry about the delay, I'll take a look in the next day or so.

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,391 @@
...
+  props: {
+    mapConfig: Object,
+    apiKey: String,
+  },
+
+  data() {
+    return {
+      google: null,
+      map: null
+    };
```
phanan: Let's following [Vue's code style](https://github.com/vuejs/eslint-plugin-vue-libs/blob/master/config.js) by removing semicolons :)

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,391 @@
...
+    new this.google.maps.Polyline({
+      path: this.path,
+      map: this.map,
+      ...LINE_PATH_CONFIG
+    });
+  }
+};
+```
+
+- both of them receive `google` from which we extract required object (Marker or Polyline) and `map` which gives as a reference to the map on which we want to place our element
```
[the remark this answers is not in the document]
mateuszRybczonek: Bullets removed from here as well.

mateuszRybczonek:
@sdras @phanan, suggestions applied, please re-check 👁 when you find some time.

sdras:
This is looking really great! The order should actually be 14, I just looked and there is already a 13. That's probably why the netlify build is failing. I think we're almost there.

mateuszRybczonek:
@sdras 14 looks good 👍, thanks.

mateuszRybczonek:
@sdras any chance of getting another review round from you? :)

sdras:
Yes, great job! Congrats!

src/v2/cookbook/practical-use-of-scoped-slots.md
```diff
@@ -0,0 +1,389 @@
...
+    const googleMapApi = await GoogleMapsApiLoader({
+      apiKey: this.apiKey
+    })
+    this.google = googleMapApi
+    this.initializeMap()
+  },
+
+  methods: {
+    initializeMap() {
+      const mapContainer = this.$refs.googleMap)
```
alexdilley: You've got a parenthesis loitering at the end there.

sdras: Good catch! @mateuszRybczonek do you want to put in another PR for that or if not I can grab it sometime

mateuszRybczonek: PR with fix here - https://github.com/vuejs/vuejs.org/pull/1848, @alexdilley thanks for spotting that 👍

alexdilley: np! I'm selfishly only using merged commit notifications as a way to keep up to speed on new documentation hence not picking up sooner; your entry was an interesting read!

It's maybe not the remit of a recipe in a cookbook, but I, as a reader, was curious to understand why this approach is more appropriate than simply setting up the loader in the parent component, for example, or perhaps more globally in a Vuex store. I think the answer is "reusability", "encapsulation" and "separation of concerns", but I then question how many types of map-related workflows you'd typically find in a project whose requirements couldn't be satisfied by a single, general-purpose map component...splitting concerns and avoiding an unwieldy "god component", maybe? Please appreciate that I may be completely missing the point, for which I apologise!!

There's a lot of buzz around this pattern, with it being a common go-to for those "in the know", and it might be pertinent – with such an example as this as context – to explain _why_ this would be the tool to reach for in this case.

alexdilley: np! I'm selfishly only using merged commit notifications as a way to keep up to speed on new documentation hence not picking up sooner; your entry was an interesting read!

It's maybe not the remit of a recipe in a cookbook, but I, as a reader, was curious to understand why this approach is more appropriate than simply setting up the loader in the parent component, for example, or perhaps more globally in a Vuex store. I think the answer is "reusability", "encapsulation" and "separation of concerns", but I then question how many types of map-related workflows you'd typically find in a project whose requirements couldn't be satisfied by a single, general-purpose map component...splitting concerns and avoiding an unwieldy "god component", maybe? Please appreciate that I may be completely missing the point, for which I apologise!!

There's a lot of buzz around this pattern, with it being a common go-to for those "in the know", and it might be pertinent – with such an example as this as context – to explain _why_ this would be the tool to reach for in this case.

alexdilley: np! I'm selfishly only using merged commit notifications as a way to keep up to speed on new documentation hence not picking up sooner; your entry was an interesting read!

It's maybe not the remit of a recipe in a cookbook, but I, as a reader, was curious to understand why this approach is more appropriate than simply setting up the loader in the parent component, for example, or perhaps more globally in a Vuex store. I think the answer is "reusability", "encapsulation" and "separation of concerns", but I then question how many types of map-related workflows you'd typically find in a project whose requirements couldn't be satisfied by a single, general-purpose map component...splitting concerns and avoiding an unwieldy "god component", maybe? Please appreciate that I may be completely missing the point, for which I apologise!!

There's a lot of buzz around this pattern, with it being a common go-to for those "in the know", and it might be pertinent – with such an example as this as context – to explain _why_ this would be the tool to reach for in this case.

alexdilley: np! I'm selfishly only using merged commit notifications as a way to keep up to speed on new documentation hence not picking up sooner; your entry was an interesting read!

It's maybe not the remit of a recipe in a cookbook, but I, as a reader, was curious to understand why this approach is more appropriate than simply setting up the loader in the parent component, for example, or perhaps more globally in a Vuex store. I think the answer is "reusability", "encapsulation" and "separation of concerns", but I then question how many types of map-related workflows you'd typically find in a project whose requirements couldn't be satisfied by a single, general-purpose map component...splitting concerns and avoiding an unwieldy "god component", maybe? Please appreciate that I may be completely missing the point, for which I apologise!!

There's a lot of buzz around this pattern, with it being a common go-to for those "in the know", and it might be pertinent – with such an example as this as context – to explain _why_ this would be the tool to reach for in this case.

PR #1756, opened 2018-08-18
Default Slot content text was unclear

katerlouis:
I know it should be clear out of context, that the default content needs to be applied inside the template of this component.
Yet I got stuck while reading this– explicitly pointing to the template again prevents these hickups and doesn't bother people who dont need it.


katerlouis:
knock knock :)

chrisvfritz:
Sorry this took so long to review. Thanks for the contribution. 🙂

katerlouis:
Thanks!

– wait.. what is this feeling?!
I am a contributor now?

cet4

PR #1760, opened 2018-08-22
Security guide

chrisvfritz:


chrisvfritz:
@shentao I decided to leave out SSR (at least for now), since most people aren't using it and we already cover security concerns specific to SSR throughout the SSR docs rather than consolidated on a single page.

shentao:
What about a short "tip" like this:
> Building _Server-Side Rendered Applications_, requires additional care to make the application secure. Please visit the ssr.vuejs.org site to learn how to avoid common issues and vulnerabilities.

chrisvfritz:
@shentao I just added a note at the end. What do you think?

shentao:
Perfect! Couldn’t write it better myself. 👍

bwokich:
Thank you gentelemen!  We really appreciate this!

src/v2/guide/security.md
```diff
@@ -0,0 +1,166 @@
...
+</a>
+```
+
+let's assume that `sanitizedUrl` has been sanitized, so that it's definitely a real URL and not JavaScript. With the `userProvidedStyles`, malicious users could still provide CSS to "click jack", e.g. styling the link into a transparent box over the "Log in" button. Then if `https://user-controlled-website.com/` is built to resemble the login page of your application, they might have just captured a user's real login information.
+
+You may be able to imagine how allowing user-provided content for a `<style>` element would create an greater vulnerability, giving that user full control over how to style the entire page:
+
+```html
+<style>{{ userProvidedStyles }}</style>
+```
```
yyx990803: Should note here that Vue will refuse to render `<style>` tags inside templates.

yyx990803:
Looks great. There's one thing I would add: sometimes we receive vulnerability reports on how it's possible to do XSS in Vue templates - and we in general do not consider such cases actual vulnerabilities, because there are two possible cases for this to actual lead to a breach:

1. You are explicitly asking Vue to render user-provided, unsanitized content as Vue templates. This is inherently unsafe and it's impractical for Vue to protect you from that.

2. You are mounting Vue to an entire HTML page which happens to contain server-rendered & user-provided content. This is actually fundamentally the same with (1), but sometimes devs may do it without realizing it. This can lead to possible vulnerabilities where the attacker provides HTML which is safe as plain HTML but unsafe as a Vue template. The best practice is to never mount Vue on nodes that may contain server-rendered & user-provided content.

sdras:
Hey @chrisvfritz! If you have time, could you please let me know where this one is at? I also wouldn't mind finishing it up for you if you're busy. Thanks!

PhilipJohnBasile:
Looking to kick start this again @chrisvfritz .

Issue #1308, opened 2017-12-05
Problem with margins in hackernews-example


Alex-Sokolov:
https://vuejs.org/v2/examples/hackernews.html:
![image](https://user-images.githubusercontent.com/4497128/33594252-3ee27584-d9a4-11e7-83eb-dc0a8d9554cf.png)

sdras:
Thanks! That does look screwy. We'll take care of it soon.

honzie:
@sdras: I opened a PR with a fix for this margin issue. Please let me know if there are any additional steps to help out. Hopefully this solution jibes well with the rest of the theme.

chrisvfritz:
This is addressed now, so closing. 🙂

PR #1761, opened 2018-08-22
fix: removing margins from nested ul/ol, top avoid awkward spacing.

honzie:
This is to address the margins in the hacker news example (https://github.com/vuejs/vuejs.org/issues/1308).

By resetting the margins to 0, this undos the trick that allows for easier text highlighting and anchor positioning.


phanan:
Thanks @honzie!

PR #1763, opened 2018-08-22
Replace Webpack Template with Vue CLI in Tooling

joshua-s:
Since Vue CLI is an official project and very useful to VueJS developers, I think it should be added to the menu. Otherwise, it can be a bit tricky to discover unless one is looking for it. The logical place seems to be Ecosystem > Tooling.

### DIFF
--- a/themes/vue/layout/partials/ecosystem_dropdown.ejs
+++ b/themes/vue/layout/partials/ecosystem_dropdown.ejs
@@ -11,7 +11,7 @@
     <li>
       <ul>
         <li><a href="https://github.com/vuejs/vue-devtools" class="nav-link" target="_blank">Devtools</a></li>
-        <li><a href="https://vuejs-templates.github.io/webpack" class="nav-link" target="_blank">Webpack Template</a></li>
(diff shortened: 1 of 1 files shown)


themes/vue/layout/partials/ecosystem_dropdown.ejs
```diff
@@ -12,6 +12,7 @@
       <ul>
         <li><a href="https://github.com/vuejs/vue-devtools" class="nav-link" target="_blank">Devtools</a></li>
         <li><a href="https://vuejs-templates.github.io/webpack" class="nav-link" target="_blank">Webpack Template</a></li>
```
LinusBorg: I would even say we should remove the webpack template from the dropdown.

We won'T put much energy in its maintanance now that vue-cli 3 is stable, so while it will continue to be available, we should not put it on the same "level" as the CLIU here.

Can you remove the link?

joshua-s: That makes sense. Done!

joshua-s:
@LinusBorg your requested changes have been made

joshua-s:
Are we ready for merge? @LinusBorg

LinusBorg:
Yep.

PR #1767, opened 2018-08-24
Use keyed loop in guide in line with the style guide

hkraal:
Per the [Vue.js style guide](https://vuejs.org/v2/style-guide/#Keyed-v-for-essential) the loop should be keyed, I feel this should be reflected in the guide as well.

_Please note that I looked at Vue.js for the first time yesterday so I'm completely innocent and green on this one._

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -151,7 +151,9 @@ There are quite a few other directives, each with its own special functionality.
 ``` html
 <div id="app-4">
   <ol>
-    <li v-for="todo in todos">
+    <li v-for="todo in todos"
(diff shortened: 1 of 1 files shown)


Justineo:
Actually this `key` is not necessary here, though always adding key is suggested by the style guide. @chrisvfritz and @michalsnik are already working on some improvements on this for `eslint-plugin-vue`, see [vuejs/eslint-plugin-vue#522](https://github.com/vuejs/eslint-plugin-vue/issues/522).

phanan:
We should modify the style guide then, because for right now exactly the same example is being used.

PR #1769, opened 2018-08-26
Update W3C link about custom element naming

jsangradorp:
The link currently referenced goes to a warning about looking up the spec somewhere else.
I am linking to that page instead.


sdras:
Nice catch, thank you.

PR #1794, opened 2018-09-12
add style guide blockquote in v-if with v-for

CeceWall:
in the vue style guide, it's an essential rule to avoid use v-if and v-for together, i thought beginners should know this.

### DIFF
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -196,6 +196,6 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 
 ## `v-if` with `v-for`
 
-> Use v-if and v-for together is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.
+<p class="tip">Use v-if and v-for together is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.</p>
 
 When used together with `v-if`, `v-for` has a higher priority than `v-if`. See the <a href="../guide/list.html#V-for-and-v-if">list rendering guide</a> for details.
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -196,6 +196,6 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 
 ## `v-if` with `v-for`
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/conditional.md
```diff
@@ -196,4 +196,6 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 
 ## `v-if` with `v-for`
 
+> Use v-if and v-for together is **not recommended**. See the style guide for futher infomation [Avoid v-if with v-for](https://vuejs.org/v2/style-guide/#Avoid-v-if-with-v-for-essential)
```
phanan: I'd suggest something like this instead:

```md
> Use v-if and v-for together is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.
```

Notice the wording changes and the two spelling mistake corrections.

CeceWall: thanks for your change and correcting, your version is much better.

CeceWall:
sorry for didn't get your point, and I push a new commit use p.tip instead.

CeceWall:
does anybody here?

src/v2/guide/conditional.md
```diff
@@ -196,4 +196,6 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 
 ## `v-if` with `v-for`
 
+<p class="tip">Use v-if and v-for together is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.</p>
```
phanan: Please enclose `v-if` and `v-for` in backticks.

phanan:
@CeceWall Thanks for the ping. One extra little change and I think we're good to go :)

CeceWall:
thanks, and i have finished that

src/v2/guide/conditional.md
```diff
@@ -196,4 +196,6 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 
 ## `v-if` with `v-for`
 
+<p class="tip">Use `v-if` and `v-for` together is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.</p>
```
phanan: Sorry, but I miss this earlier 😅: It's "Using," not "Use."

phanan:
Thanks!

PR #1795, opened 2018-09-12
Clarify status of Web Components standard

TimvdLippe:
The standard has been finalized and is shipping in the next version of Firefox. I therefore updated the wording to reflect this status.


phanan:
Thanks for the PR! Can you link us to the current status, though? Google fails me this time.

phanan:
@TimvdLippe forgive my ignorance (I'm not exactly familiar with web components) but isn't HTMLImport still a working draft according to [MDN](https://developer.mozilla.org/en-US/docs/Web/Web_Components#Specifications)?

TimvdLippe:
@phanan No problem at all. The HTML Import standard has been abandoned, as Mozilla have stated they are not going to implement it.

TimvdLippe:
@phanan Is there anything else you would like to be clarified before this PR can be merged in?

phanan:
@TimvdLippe Not really – I was actually waiting for another opinion from the core members, but I guess this is good enough :) Thanks for the PR!

TimvdLippe:
Awesome thanks!

Issue #1787, opened 2018-09-03
The example given for "Disabling Attribute Inheritance" doesn't work


estyh:
The docs show an example for how to disable attribute inheritance (https://vuejs.org/v2/guide/components-props.html#Disabling-Attribute-Inheritance) - I tried, and it doesn't work!
In this example, the placeholder is passed to the input, but class is not, it is applied to the root-element, label. Are the docs wrong, or is something broken in Vue?
As a side note, the example uses a label prop for the child component, but doesn't pass it in as an attribute.
Here is my js-fiddle example, almost straight out of the docs (with some css for clarity)
https://jsfiddle.net/estherh/y12eb6xg/

leopiccionia:
From [docs](https://vuejs.org/v2/api/#inheritAttrs):
> Note: this option does **not** affect `class` and `style` bindings.

Perhaps this note could appear in guide, too?

LinusBorg:
Indeed. And the example should not use `class`, of course.

phanan:
@leopiccionia Any progress on this one?

leopiccionia:
@phanan I've closed my original PR, I'll open an improved one yet today.

PR #1796, opened 2018-09-14
Add clarification about inheritAttrs and class/style bindings. (#1787)

leopiccionia:
I've update the guide, adding the same note that appears on API docs.

I've also changed the given example for using `inheritAttrs: false` combined with base components, removing mentions to `class` to avoid confusion.

### DIFF
--- a/src/v2/guide/components-props.md
+++ b/src/v2/guide/components-props.md
@@ -330,12 +330,14 @@ Vue.component('base-input', {
 })
 ```
 
+<p class="tip">Note that `inheritAttrs: false` option does **not** affect `style` and `class` bindings.</p>
+
 This pattern allows you to use base components more like raw HTML elements, without having to care about which element is actually at its root:
 
 ```html
 <base-input
-  v-model="username"
-  class="username-input"
-  placeholder="Enter your username"
+  v-model="email"
+  type="email"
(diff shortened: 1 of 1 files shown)


src/v2/guide/components-props.md
```diff
@@ -330,12 +330,14 @@ Vue.component('base-input', {
...
+<p class="tip">Note that `inheritAttrs: false` option does **not** affect `style` and `class` bindings.</p>
+
 This pattern allows you to use base components more like raw HTML elements, without having to care about which element is actually at its root:
 
 ```html
 <base-input
-  v-model="username"
-  class="username-input"
-  placeholder="Enter your username"
+  v-model="email"
```
phanan: Why this change?

leopiccionia: Original example had two non-inherited attributes, so I'd want to keep this count. I think that adding a custom `type` attribute made sense in this specific example (and added small cognitive load).

I can partially revert it, removing problematic `class` and keeping only `placeholder` attribute.

leopiccionia: I'm replacing the example above, too, since it's plainly wrong (neither `class` nor `style` keys ever appear in `$attrs` binding).

Issue #1793, opened 2018-09-10
Cookbook/Client-side Storage example description doesn't correspond to the example


maxirozay:
At the end of the base example it says:
> You could also store a date representing when the values were stored. With that metadata, the mounted method could make a logical call on whether or not to store the values again, such as in this version below. You can try this version below.

But in the example the mounted method only do:
```
if(localStorage.name) this.name = localStorage.name;
if(localStorage.age) this.age = localStorage.age;
```
The mounted method only check if the values exist, it never "make a logical call on whether or not to store the values again".

LinusBorg:
Would you be so kind and link to the section that you are talking about? We don't haven't memorized very section of the documentation, and I have no idea what you are talking about right now.

maxirozay:
Sure it's in the cookbook/client-side storage https://vuejs.org/v2/cookbook/client-side-storage.html. Go at the last paragraph of the base example.

maxirozay:
@cfjedimaster should you change your CodePen example or should we change the description?

maxirozay:
Ok I can do that.

maxirozay:
Done in #1797

PR #1797, opened 2018-09-14
fix example description (#1793)

maxirozay:
The description is just a suggestion of what we can do and it's not actually in the "version below".


phanan:
Thanks!

PR #1803, opened 2018-09-17
replace indexOf with includes in props documentation

reed-jones:


phanan:
Thanks, but we consciously use ES5 in the main docs (except for the cookbook, where ES6 is preferred) :)

PR #1805, opened 2018-09-19
add max prop instructions for <keep-alive>

wuyuweixin:


phanan:
While the idea is good, I'm afraid the quality of the PR isn't optimal. Take a look at the Netlify build and you'll see what I mean :)

sdras:
Looks like the trouble is not closing the html triple backticks

src/v2/api/index.md
```diff
@@ -2516,6 +2516,16 @@ Used to denote a `<template>` element as a scoped slot, which is replaced by [`s
 
 - **See also:** [Dynamic Components - keep-alive](../guide/components.html#keep-alive)
 
+>New in 2.5.0+
+
+`max` maximum number of components allowed for caching (FIFO)
```
Justineo: We should add a corresponding description into the `props` section of `keep-alive`.

```md
  - `max` - number. The maximum number ... . (new in 2.5.0+)
```

src/v2/api/index.md
```diff
@@ -2516,6 +2516,16 @@ Used to denote a `<template>` element as a scoped slot, which is replaced by [`s
 
 - **See also:** [Dynamic Components - keep-alive](../guide/components.html#keep-alive)
 
+>New in 2.5.0+
```
Justineo: Missing a space after `>`.

Justineo: We can add a section similar to the previous “`include` and `exclude`”.

src/v2/api/index.md
```diff
@@ -2516,6 +2516,16 @@ Used to denote a `<template>` element as a scoped slot, which is replaced by [`s
...
+
+`max` maximum number of components allowed for caching (FIFO)
+   ``` html
```
Justineo: Indent with 2 spaces instead of 3 here.

src/v2/api/index.md
```diff
@@ -2516,6 +2516,16 @@ Used to denote a `<template>` element as a scoped slot, which is replaced by [`s
...
+  <!-- usage -->
+  <keep-alive max="10">
+    <component></component>
+  </keep-alive>
+   ```
```
Justineo: As above

wuyuweixin: OK

chrisvfritz:
Thanks for this and sorry for the delay in reviewing. I wanted to add more detail and different examples, so I just added this to the API doc in a separate commit. 🙂

PR #1814, opened 2018-09-30
Give a hint for older vuejs versions and slots

iskrid:


phanan:
Let's merge this with #1956.

PR #1830, opened 2018-10-07
Light method refactoring

NogrTL:
Light refactoring that uses method declaration and arrow function

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -35,7 +35,7 @@ var vm = new Vue({
   },
   computed: {
     // a computed getter
-    reversedMessage: function () {
+    reversedMessage() {
       // `this` points to the vm instance
       return this.message.split('').reverse().join('')
     }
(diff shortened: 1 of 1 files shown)


sdras:
Yes, it's intentional, we had talked about the possibility of changing it. (Informally, so perhaps not the final call).

NogrTL:
@Justineo 
As far as I know it's supported with all browsers with latest version, except the IE.

sdras:
The use of ES5 is indeed intentional. We have talked briefly about updating the docs but I think if we did, we'd have to do a large sweep of everything. Pining @chrisvfritz in case I'm mistaken.

NogrTL:
I'll be nice then to mention that you can use those(method shorthand and arrow function instead of binding this to variable) and warn users that IE doesn't support those features, for example.

phanan:
As everyone else said, ES5 usage in the docs (except for the cookbook) is intentional. Personally, I'd like to keep it as-is (as Vue will continue supporting IE11 even in 3.0) and even if we switch to ES2015, the scope will be much larger than just this `computed.md`. As such, I'm taking the liberty to close this PR :)

PR #1832, opened 2018-10-08
Update plugins.md

bk1973:


phanan:
Thanks, but I don't think the current example is incorrect. What you're doing is adding the method into the prototype, enabling it on every new instance, something we've already handled on line 80 (note that we even start the instance method with `$` as a best practice).

bk1973:
so, how to add the method into single instance in Plugin.install ?

PR #1834, opened 2018-10-10
Fix documentation on how to use production mode with webpack3

jsilvestri:
In order to get production mode to work with webpack 3, I needed to use vue.min, which is not part of the documentation.  The current documentation suggests you need to use vue.min only if you are *not* using a buildpack.  I have fixed the documentation below.


LinusBorg:
That's not correct, what makes you think that?

PR #1836, opened 2018-10-11
Add note in list.md on usage of v-for with v-if

robaxelsen:
Noticed that this was mentioned in the guide, but also mentioned in category "Essential" in the style guide as a pattern to avoid. Therefore suggesting this change where we warn that this is not best-practice and
should be avoided.

Could you please add the label "hacktoberfest" to this PR before reviewing? I would be very grateful :heart:

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vuejs.org",
   "private": true,
   "hexo": {
-    "version": "3.7.0"
+    "version": "3.7.1"
   },
   "scripts": {
     "start": "hexo server",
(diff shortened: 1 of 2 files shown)


phanan:
>Could you please add the label "hacktoberfest" to this PR before reviewing?

@robaxelsen Sure, but FWIW, you don't need the label to be eligible for the program :)

For the PR though, why do I feel like we've merged something like this before?

sdras:
Thanks for the PR! @phanan is correct, we already include this note earlier in the documentation. Appreciated, though :)

PR #1838, opened 2018-10-12
Fix #1802; Update docs to avoid misconceptions around emitting events

### LINKED ISSUE #1802: Misleading description on how $emit works in the guide
Noticed a few times already that there is a misconception around the way `this.$emit` works. The source of the problem might come from this line:
> To emit an event to the parent, we can call the built-in $emit method, passing the name of the event
found here: https://vuejs.org/v2/guide/components.html#Sending-Messages-to-Parents-with-Events

I understand that this might be there to make it slightly easier to grasp the idea of `props down, events up` but I’m not sure if this is worth it.

The issue is that it might mislead the user that it actually emits an event **to the parent component** even though the API docs clearly state:
> Trigger an event on the current instance. Any additional arguments will be passed into the listener’s callback function.

I think it might be a good idea to update the guide to highlight that fact that `$emit` only emits an event on itself and the parent component can choose to listen to those events by setting up listeners on the child instance through the template syntax: `@event-name="handler"`.

I can PR a change to the guide if we agree to introduce such change.

shentao:


src/v2/guide/components.md
```diff
@@ -268,7 +268,7 @@ Vue.component('blog-post', {
 
 Now, whenever a new property is added to `post` objects, it will automatically be available inside `<blog-post>`.
 
-## Sending Messages to Parents with Events
+## Listening to Child Components Events
```
sdras: Such a simple change and it's already easier to understand. Nicely done.

shentao: Thank you Sarah!

shentao:
I think that having a diagram (possibly animated) similar to the Vuex diagram could be very helpful, especially if we can show some code snippets on top of it and visualize how everything connects.
For example, how adding `@event-name` sets a listener on a child instance that will be triggered when the child calls `$emit`.

PR #1839, opened 2018-10-14
Changed ES7 to ES2016 as per comment

hitautodestruct:
Changed the ES7 nickname to ES2016 as per a comment request. Also added `Vue.component` so syntax highlighter will pick up on the code.

### DIFF
--- a/src/v2/guide/reactivity.md
+++ b/src/v2/guide/reactivity.md
@@ -119,10 +119,11 @@ Vue.component('example', {
 })
 ```
 
-Since `$nextTick()` returns a promise, you can achieve the same as the above using the new [ES7 async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) syntax:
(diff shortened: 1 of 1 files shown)


src/v2/guide/reactivity.md
```diff
@@ -119,10 +119,11 @@ Vue.component('example', {
 })
 ```
 
-Since `$nextTick()` returns a promise, you can achieve the same as the above using the new [ES7 async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) syntax:
+Since `$nextTick()` returns a promise, you can achieve the same as the above using the new [ES2016 async/await](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) syntax:
 
 ```
-...
+Vue.component('example', {
```
phanan: Hmm, can't you just use ```js for code syntax highlighting?

hitautodestruct: Don't know.. can you?
I am not familiar with the syntax highlighter the vuejs docs use.

phanan: Try it ;)

phanan:
Thanks!

PR #1842, opened 2018-10-17
Update computed.md in regards to computing objects

joelworsham:
Add documentation about how computing based on objects and arrays will require a specific implementation.


phanan:
I think this has been covered in Reactivity's [Change Detection Caveats](https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats). If it's really necessary (I don't think so), we can add a note that links to the caveats section instead.

joelworsham:
Oh! I did not see this. It took me a while to figure this out via the Google previously, so I thought perhaps it could go right into the computed/watcher docs. But you're right, no need to be redundant! I would love to at least see, as you mentioned, a note about it and pointing to that article. Great idea

joelworsham:
Although, it doesn't seem to quite touch on the exact same point. I mean, it's up to you guys in the end, but I personally would love to seem some documentation that makes explicit note of computed/watchers and objects/arrays.

sdras:
Thank you for taking the time to contribute!

However, this isn't accurate the way you're describing in the example, I think it's misleading. Here's a [codepen example](https://codepen.io/sdras/pen/387592157d58be32e83460c656271136?editors=1010) for reference. Please feel free to reopen if you think I'm missing something that needs more explaining.

PR #1844, opened 2018-10-18
preventing empty todo item addition

xeptore:
adds new todo item only if there is a title for it


phanan:
The examples are only meant for demonstration purpose _for the language feature under discussion_, so I'd prefer to keep them simple and concise instead of adding validations and such.

Issue #1765, opened 2018-08-23
Documentation advises to use deprecated keyCode attribute


deltragon:
The documentation on Event Handling, specifically on [key modifiers](https://vuejs.org/v2/guide/events.html#Key-Modifiers) advises to use the
[deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) keyCode attribute to detect pressed keys. Only as an afterthought/convenience method it mentions that the KeyboardEvent.key aliases can be used as well.

Ideally the documentation would present the .key alias as the default usage, and mention that usage of keyCode is deprecated.

mrmowji:
It'd be good to mention what's the alternative way to trigger events for keys with no [standard key values](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values), for example alphanumeric keys.

PR #1850, opened 2018-10-22
Update Key Modifiers to suggest KeyboardEvent.key

kokopelli314:
Fixes [#1765](https://github.com/vuejs/vuejs.org/issues/1765). 

I left the "define custom key modifier aliases" section at the bottom, below the deprecated key codes information. Definitely open to changing / re-arranging as needed!

### DIFF
--- a/src/v2/guide/events.md
+++ b/src/v2/guide/events.md
@@ -223,24 +223,26 @@ The `.passive` modifier is especially useful for improving performance on mobile
 
 ## Key Modifiers
 
-When listening for keyboard events, we often need to check for common key codes. Vue also allows adding key modifiers for `v-on` when listening for key events:
+When listening for keyboard events, we often need to check for specific keys. Vue allows adding key modifiers for `v-on` when listening for key events:
 
 ``` html
-<!-- only call `vm.submit()` when the `keyCode` is 13 -->
+<!-- only call `vm.submit()` when the `key` is PageDown -->
+<input @keyup.page-down="onPageDown">
+```
+
+In the above example, the handler will only be called if `$event.key === 'PageDown'`.
+
+You can directly use any valid key names exposed via [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) as modifiers by converting them to kebab-case.
+
+### Key Codes
+
+The usage of `keyCode` attributes is deprecated, but can be used if needed:
+
+``` html
 <input v-on:keyup.13="submit">
 ```
 
-Remembering all the `keyCode`s is a hassle, so Vue provides aliases for the most commonly used keys:
-
-``` html
-<!-- same as above -->
-<input v-on:keyup.enter="submit">
-
-<!-- also works for shorthand -->
-<input @keyup.enter="submit">
-```
-
-Here's the full list of key modifier aliases:
+Vue provides aliases for the most commonly used key codes when necessary for legacy browser support:
 
 - `.enter`
 - `.tab`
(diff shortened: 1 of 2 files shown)


src/v2/guide/events.md
```diff
@@ -223,24 +223,26 @@ The `.passive` modifier is especially useful for improving performance on mobile
...
 
 ``` html
-<!-- only call `vm.submit()` when the `keyCode` is 13 -->
-<input v-on:keyup.13="submit">
+<!-- only call `vm.submit()` when the `key` is PageDown -->
+<input @keyup.page-down="onPageDown">
 ```
 
-Remembering all the `keyCode`s is a hassle, so Vue provides aliases for the most commonly used keys:
+In the above example, the handler will only be called if `$event.key === 'PageDown'`.
```
phanan:
```suggestion
In the above example, the handler will only be called if `$event.key` is equal to `'PageDown'`.
```

src/v2/guide/events.md
```diff
@@ -223,24 +223,26 @@ The `.passive` modifier is especially useful for improving performance on mobile
...
-``` html
-<!-- same as above -->
-<input v-on:keyup.enter="submit">
+You can directly use any valid key names exposed via [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) as modifiers by converting them to kebab-case.
+
+### Key Codes
 
-<!-- also works for shorthand -->
-<input @keyup.enter="submit">
+The usage of `keyCode` attributes is deprecated, but can be used if needed:
```
phanan: A better approach here IMHO would be using some `p.tip` as a warning instead. Linking "deprecated" to the MDN article would also be nice.

phanan:
Very nice :) Just a bit of change and this should be golden (for me at least).

kokopelli314:
Just made the updates, let me know if there's any other adjustments you'd recommend.

src/v2/guide/events.md
```diff
@@ -223,24 +223,28 @@ The `.passive` modifier is especially useful for improving performance on mobile
...
 ## Key Modifiers
 
-When listening for keyboard events, we often need to check for common key codes. Vue also allows adding key modifiers for `v-on` when listening for key events:
+When listening for keyboard events, we often need to check for specific keys. Vue allows adding key modifiers for `v-on` when listening for key events:
 
 ``` html
-<!-- only call `vm.submit()` when the `keyCode` is 13 -->
-<input v-on:keyup.13="submit">
+<!-- only call `vm.submit()` when the `key` is PageDown -->
+<input @keyup.page-down="onPageDown">
```
Jinjiang: Sorry for late, why here the `key` becomes `PageDown` not still `Enter`? (The `keyCode` of `Enter` is `13` and the `keyCode` of `PageDown` is `34`. Ref: https://keycode.info, https://css-tricks.com/snippets/javascript/javascript-keycodes/)
I think it's better to remain the key `Enter` is this case to match the meaning of "submit".

Thanks.

phanan: That's a good point @Jinjiang, sorry I missed this. @kokopelli314 would you mind a subsequent PR?

src/v2/guide/events.md
```diff
@@ -252,27 +256,15 @@ Here's the full list of key modifier aliases:
...
 
-### Automatic Key Modifiers
-
-> New in 2.5.0+
-
-You can also directly use any valid key names exposed via [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) as modifiers by converting them to kebab-case:
-
-``` html
-<input @keyup.page-down="onPageDown">
-```
```
Jinjiang: I guess the key `PageDown` comes from here. Maybe we could rearrange these paragraphs to preserve this code example to show how kebab-case name converted. I will leave more advice for this to PR https://github.com/vuejs/vuejs.org/pull/1970. @phanan 😉

PR #1853, opened 2018-10-23
Added usage information about inline-templates and x-templates

MSCAU:
I wasted a whole day on this issue as it wasn't documented, AFAIK. StackOverflow people thankfully were able to help: https://stackoverflow.com/questions/52940350/vue-component-error-property-or-method-x-is-not-defined-on-the-instance-but-ref

### DIFF
--- a/src/v2/guide/components-edge-cases.md
+++ b/src/v2/guide/components-edge-cases.md
@@ -331,6 +331,8 @@ When the `inline-template` special attribute is present on a child component, th
 </my-component>
 ```
 
+<p>Your inline template needs to be defined outside the DOM element to which Vue is attached.</p>
+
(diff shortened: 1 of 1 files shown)


MSCAU:
Sorry, seems I was mistaken about the inline-templates which need to reside *inside* the attached DOM element.

sdras:
Ah, yeah, true. I wouldn't have thought of this as a place to get caught but that makes sense, we should be more clear here. Thanks!

src/v2/guide/components-edge-cases.md
```diff
@@ -331,6 +331,8 @@ When the `inline-template` special attribute is present on a child component, th
 </my-component>
 ```
 
+<p>Your inline template needs to be defined inside the DOM element to which Vue is attached.</p>
```
Jinjiang: The `<p>` tags here are not necessary because it's markdown. 🙂

Jinjiang: Just created a PR for this https://github.com/vuejs/vuejs.org/pull/2010.
Thanks

MSCAU: Great. Much appreciated.

PR #1855, opened 2018-10-25
Update custom-directive.md to link to VNode info where VNodes are first mentioned

### LINKED ISSUE #1799: VNode is used in the Guide before it is defined
Brand-new to Vue, and I'm reading the Guide front to back.  The term `VNode` is used in the [Custom Directives](https://vuejs.org/v2/guide/custom-directive.html) page.  However, it is not defined until later in the Guide, in the [Render Functions & JSX](https://vuejs.org/v2/guide/render-function.html) page.  I see that was mentioned in [#319 (comment)](https://github.com/vuejs/vuejs.org/issues/319#issuecomment-237098016).

Would someone who actually knows Vue :) be willing to update the Guide so that `VNode` is not used without explanation on the Custom Directives page?  I don't claim to know what would be the best way to accomplish that :) .  Thank you!

cxw42:
Otherwise, VNode is used without definition the first time you encounter it in the Guide.  

This PR implements the preferred change from @phanan in [#1799 (comment)](https://github.com/vuejs/vuejs.org/issues/1799#issuecomment-422380734), except that I added a direct link to the section as well as the link to the page.

Fixes #1799.

First PR in Vue.  Thanks for considering this request!

### DIFF
--- a/src/v2/guide/custom-directive.md
+++ b/src/v2/guide/custom-directive.md
@@ -66,6 +66,8 @@ A directive definition object can provide several hook functions (all optional):
 
 - `update`: called after the containing component's VNode has updated, __but possibly before its children have updated__. The directive's value may or may not have changed, but you can skip unnecessary updates by comparing the binding's current and old values (see below on hook arguments).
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks!

PR #1858, opened 2018-10-27
If `<navigation-link>` did **not** contain -> If template `<navigation-link>` did **not** contain

aidevio:


phanan:
Sorry, but I'm a bit lost. What's the problem (that we're trying to solve) here?

phanan:
Thanks for the explanation @aidevio :) But I'd say it reads better (when still clear enough) as it is now, so I'm closing it. Looking forward to your other PR's though!

PR #1862, opened 2018-10-31
Syntax: punctuation changes; link to Event Handling

jdhines:
- Minor punctuation changes
- In the Arguments section, link to the Event Handling page

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -54,7 +54,7 @@ new Vue({
 </script>
 {% endraw %}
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for your contribution! I'm not very sure if this is the best we could have, though:

* The change from hyphen to em dash, though valid (IMHO), should be made more consistent throughout the docs and not just on this one page.
* For the commas inside quotes: while I personally follow the same convention, I'd like to see it as a matter of taste and not a rule to follow, and even if we _did_ follow it, again, the change should be made globally.
* Please have separated PR's for different fixes.

sdras:
As this as been inactive for some time and we're only taking one change, we'll close it out. If you feel like opening another PR for that change, we'll be happy to take it, thanks.

PR #1878, opened 2018-11-13
Fix the async_function syntax in the example

lyswhut:
The example of using async_function syntax is wrong, fix it.

### DIFF
--- a/src/v2/guide/reactivity.md
+++ b/src/v2/guide/reactivity.md
@@ -123,7 +123,7 @@ Since `$nextTick()` returns a promise, you can achieve the same as the above usi
 
 ``` js
   methods: {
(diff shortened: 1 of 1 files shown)


src/v2/guide/reactivity.md
```diff
@@ -123,7 +123,7 @@ Since `$nextTick()` returns a promise, you can achieve the same as the above usi
 
 ``` js
   methods: {
-    async updateMessage: function () {
+    async updateMessage() {
```
phanan:
```suggestion
    updateMessage: async function () {
```

lyswhut:
They all work, the former is more concise, why choose the latter? Is this to maintain the code style?

PR #1882, opened 2018-11-15
Clarify that defineProperty is not supported below ES5

spurton:
This makes it clear that it is supported for ES5 and above and is less ambiguous.


Justineo:
Can you explain a little more about how is the current description ambiguous?

phanan:
Thanks for the PR, but I think the current version reads and explains better. Not to mention, your change has grammar and spelling errors ("backwards compatable").

PR #1883, opened 2018-11-15
Updated syntax

divyanshu-rawat:
Original:

`<div v-bind:class="[isActive ? activeClass : '', errorClass]"></div>`

Updated:

`<div v-bind:class="[isActive ? 'activeClass' : '', 'errorClass']"></div>`

Vue template will not compile if we will use the `Orginal syntax` that is mentioned in the doc, in order to apply classes and template to be compiled it should be replaced by `Updated syntax`

### DIFF
--- a/src/v2/guide/class-and-style.md
+++ b/src/v2/guide/class-and-style.md
@@ -100,7 +100,7 @@ Which will render:
 If you would like to also toggle a class in the list conditionally, you can do it with a ternary expression:
 
 ``` html
-<div v-bind:class="[isActive ? activeClass : '', errorClass]"></div>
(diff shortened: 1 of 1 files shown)


divyanshu-rawat:
Hey @phanan can you please review this :).

Justineo:
The original code implied there are `activeClass` and `errorClass` properties in the component instance, just like `isActive`.

divyanshu-rawat:
@Justineo That's fine 👍

phanan:
Like @Justineo explained above, I don't think we need to change in this case. Looking forward to your next contributions ;)

PR #1885, opened 2018-11-17
Adding addition example and explanations

GregariousAlex:
Making it easier to understand and providing another example of how scroll animations can be done.

### DIFF
--- a/src/v2/cookbook/creating-custom-scroll-directives.md
+++ b/src/v2/cookbook/creating-custom-scroll-directives.md
@@ -36,7 +36,7 @@ new Vue({
   }
 })
 ```
-
+When the element is inserted, a function, `f`, is created with the event, `evt`, as it's parameter. Then the scroll event listener is added, with the function `f`. When the user scrolls the `handleScroll` function checks if the `scrollY` of the window is greater than 50. If so, the element is displayed and the event listener is removed. 
 ```html
 <div id="app">
   <h1 class="centered">Scroll me</h1>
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -36,7 +36,7 @@ new Vue({
   }
 })
 ```
-
+When the element is inserted, a function, `f`, is created with the event, `evt`, as it's parameter. Then the scroll event listener is added, with the function `f`. When the user scrolls the `handleScroll` function checks if the `scrollY` of the window is greater than 50. If so, the element is displayed and the event listener is removed.
```
phanan: I don't think we should explain the code. Good code is self-explanatory, and what we have here is not an exception.

src/v2/cookbook/creating-custom-scroll-directives.md
```diff
@@ -96,6 +96,57 @@ new Vue({
 
 Though we would remove the previous CSS transition from this implementation because it's now handled with JavaScript.
 
+## Another Example
```
phanan: Call me conservative, but I don't find the need for another example here either.

phanan:
Thanks for the PR :) As explained in the review, though, I personally find this change unnecessary.

GregariousAlex:
I think that have a short explanation does not hurt, as it allows more people to understand it. I personally had to take a quick look, as I use a different set of variables for events. Untimely, it may be a bit redundant and I respect your decision.

PR #1886, opened 2018-11-18
Address self-closing component syntax

sheriffderek:
As a user coming from other frameworks - you might expect self-closing element syntax to be acceptable. Upon creating your first component this way - it will appear to work as intended. However, no elements after this component will be rendered and there isn't any warning or error. When inspecting the elements - the DOM seems to have it's closing tags. _So_, build a component: use a `<self-closing />` syntax and it'll work as you might expect: then put that in a v-for or use it twice and everything breaks with no information on why. It could lead users to question the directive - or the CLI - or any number of things that they shouldn't be thinking about.

I'd welcome better phrasing. : )

_Note: Vue Component syntax is never written with self-closing tags like `<br />` or `<component-name />` and must **always** be written with an accompanying closing tag. `<component-name></component-name>`_ .   - I thought about saying 'regardless of slot' - but didn't want to mention slots this early.

Should there be a link to why this is? That seems unnecessary - as it's not really open for debate due to how vue uses the browser parser.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -51,6 +51,8 @@ new Vue({ el: '#components-demo' })
 
 Since components are reusable Vue instances, they accept the same options as `new Vue`, such as `data`, `computed`, `watch`, `methods`, and lifecycle hooks. The only exceptions are a few root-specific options like `el`.
 
+Note: Vue Component syntax is never written with self closing tags like `<br />` or `<component-name />` and must **always** be written with an accompanying closing tag. `<component-name></component-name>`
+
 ## Reusing Components
 
 Components can be reused as many times as you want:


phanan:
> So, build a component: use a <self-closing /> syntax and it'll work as you might expect: then put that in a v-for or use it twice and everything breaks with no information on why.

This is incorrect – Vue components can totally be self-closing. I'd suggest that your error comes from elsewhere.

sheriffderek:
@phanan - Yeah... I guess I need to track down what **_MY_** problem is / because - I've been looking at some plugins and they seem to be using self-closing stuff just fine. If you have time - could you look at line 34 in this pen: https://codepen.io/pxlagency/pen/MzvboZ?editors=1010

phanan:
Oh, if you’re not using a build process (webpack) then yes, self-closing is
not supported. This is not a Vue problem per-se, but an HTML’s rule itself
(self-closing non-void elements are an error, but browsers choose to ignore
it). Build tools, on the other hands, will take care of self-closing tags
for you though.

sheriffderek <notifications@github.com> schrieb am Di. 20. Nov. 2018 um
01:48:

phanan:
See
https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5
.

An Phan <me@phanan.net> schrieb am Di. 20. Nov. 2018 um 08:55:

sheriffderek:
@phanan - Well, that's good to know! Maybe I can figure out the situations where you CAN and CANT use them - because I've tried in 3 or 4 different setups / and it's something that shouldn't be a pain point for new users. Maybe there are some combinations of using webpack and other frameworks that don't work as expected. I'll investigate.

sdras:
This isn't accurate so I'm going to close it out. I think there is probably a place to address something like this, but it would need a new location, new language, etc, so probably best just to start again.

PR #1887, opened 2018-11-20
Update render-function.md

hudarui:
Unnecessary escape character: \-， remove it also work.
it's error in eslint check


phanan:
Thanks!

PR #1897, opened 2018-12-03
Add a note that slot-scope expects a string value

speedpro:
I'm used to using the shorthand `:attr` for `v-bind:attr`, and I didn't notice that `v-bind` was missing for `slot-scope` in the docs. I spent several hours scouring these docs, the internet, and my code as I was trying to figure out why my slots were showing up in `vm.$slots` instead of `vm.$scopedSlots`.

I got confused because the _destructuring `slot-scope`_ section of this page said I could use any valid JS expression for `slot-scope`. I thought it might be useful to clarify that the actual attribute for `slot-scope` needs to be passed through as a literal string.

### DIFF
--- a/src/v2/guide/components-slots.md
+++ b/src/v2/guide/components-slots.md
@@ -227,3 +227,5 @@ The value of `slot-scope` can actually accept any valid JavaScript expression th
 ```
 
 This is a great way to make scoped slots a little cleaner.
+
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
Thank you for working on this PR!
I am not sure if it's still relevant considering the changes to scoped slots syntax in Vue 2.6 though. Closing it for now, please, feel free to reopen if you don't agree or if you find the issue persists with current scoped slots syntax.

PR #1900, opened 2018-12-05
Specify which types are allowed as values for 'key' of 'v-for'

gsmnv:
[List Rendering](https://vuejs.org/v2/guide/list.html#key) guide doesn't mention what types of values user can supply for `key` of `v-for`. There is a console warning for this already but it's hard to point people to it. This pull request fixes the issue.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -233,6 +233,8 @@ It is recommended to provide a `key` with `v-for` whenever possible, unless the 
 
 Since it's a generic mechanism for Vue to identify nodes, the `key` also has other uses that are not specifically tied to `v-for`, as we will see later in the guide.
 
+<p class="tip">Don't use non-primitive values like objects and arrays as `v-for` keys. Use string or numeric values instead.</p>
+
 ## Array Change Detection
 
 ### Mutation Methods


src/v2/guide/list.md
```diff
@@ -221,7 +221,7 @@ When Vue is updating a list of elements rendered with `v-for`, by default it use
 
 This default mode is efficient, but only suitable **when your list render output does not rely on child component state or temporary DOM state (e.g. form input values)**.
 
-To give Vue a hint so that it can track each node's identity, and thus reuse and reorder existing elements, you need to provide a unique `key` attribute for each item. An ideal value for `key` would be the unique id of each item. This special attribute is a rough equivalent to `track-by` in 1.x, but it works like an attribute, so you need to use `v-bind` to bind it to dynamic values (using shorthand here):
+To give Vue a hint so that it can track each node's identity, and thus reuse and reorder existing elements, you need to provide a unique `key` attribute for each item. An ideal value for `key` would be the unique id of each item, also avoid using non-primitive values as key, use string/number value instead. This special attribute is a rough equivalent to `track-by` in 1.x, but it works like an attribute, so you need to use `v-bind` to bind it to dynamic values (using shorthand here):
```
phanan: The added text feels pretty out of place. I'd suggest adding a `p.note` at the end of the section instead.

gsmnv: @phanan Good point, but did you mean `p.tip`? I've updated this PR.

src/v2/guide/list.md
```diff
@@ -233,6 +233,8 @@ It is recommended to provide a `key` with `v-for` whenever possible, unless the
 
 Since it's a generic mechanism for Vue to identify nodes, the `key` also has other uses that are not specifically tied to `v-for`, as we will see later in the guide.
 
+<p class="tip">Avoid using non-primitive values as key, use string/number value instead.</p>
```
phanan:
```suggestion
<p class="tip">Don't use non-primitive values like objects and arrays as `v-for` keys. Use string or number values instead.</p>
```

I know the message is taken directly from Vue's warning, but for a documentation purpose, it doesn't need to be that brief ;)

src/v2/guide/list.md
```diff
@@ -233,6 +233,8 @@ It is recommended to provide a `key` with `v-for` whenever possible, unless the
 
 Since it's a generic mechanism for Vue to identify nodes, the `key` also has other uses that are not specifically tied to `v-for`, as we will see later in the guide.
 
+<p class="tip">Don't use non-primitive values like objects and arrays as `v-for` keys. Use string or numeric values instead</p>
```
phanan: You miss a punctuation here :)

george-semenov: @phanan Fixed, sorry. :man_facepalming:

gsmnv: @phanan Fixed, sorry. :man_facepalming:

phanan:
Thanks!

PR #1909, opened 2018-12-15
Specified accurate url.

numb86:
Because the specification of link was not accurate, Multiple times redirect was occurring.
Since we finally transit to `https://vuejs.org/v2/guide/components-registration.html#Local-Registration`, so I specified that URL.


phanan:
Thanks!

PR #1910, opened 2018-12-15
Update components.md

alexeydemin:


phanan:
Thanks for your contribution, but I don't think the current text is incorrect. We're referring to the example right _above_ the text:

```js

<blog-post
  v-for="post in posts"
  v-bind:key="post.id"
  v-bind:title="post.title"></blog-post>
```

PR #1912, opened 2018-12-17
explain what 'trim' does

smilingkite:
'This trims out white-space from the input. '

### DIFF
--- a/src/v2/guide/forms.md
+++ b/src/v2/guide/forms.md
@@ -378,6 +378,8 @@ If you want user input to be trimmed automatically, you can add the `trim` modif
 <input v-model.trim="msg">
 ```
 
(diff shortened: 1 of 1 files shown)


phanan:
Hmm. I believe the sentence right before the example explains the functionality quite well already:

>  If you want user input to be trimmed automatically, you can add the `trim` modifier to your `v-model` managed inputs.

Can you elaborate why/how it doesn't for you?

smilingkite:
It doesn't explain that to trim means to get rid of white space. I had to
look it up online.

Katinka on mobile

chrisvfritz:
Thanks for this! I agree it's better not to assume familiarity with `String.prototype.trim`. 🙂 I just made a small tweak to the language.

PR #1924, opened 2018-12-29
Move the footnote about using v-for with v-if to list.md

karolsw3:
The footnote about using v-for with v-if shouldn't be present in conditional rendering docs, because list rendering has not been introduced yet.
Also, v-for with v-if is also described in the list rendering docs, which can mislead people.

It would be better if the whole information about it was present in the list rendering docs.

### DIFF
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -194,8 +194,3 @@ In comparison, `v-show` is much simpler - the element is always rendered regardl
 
 Generally speaking, `v-if` has higher toggle costs while `v-show` has higher initial render costs. So prefer `v-show` if you need to toggle something very often, and prefer `v-if` if the condition is unlikely to change at runtime.
 
-## `v-if` with `v-for`
-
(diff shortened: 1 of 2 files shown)


aroue2:
Hello,

I would move the tip paragraph just above because the recomandation is not about wrapping a v-for with a v-if but about using v-for and v-if on the same element.
Moreover, I would replace "Using `v-if` and `v-for` together" with "Using `v-if` and `v-for` on the same element"

Like that :
> The above only renders the todos that are not complete.
> 
> <p class="tip">Using `v-if` and `v-for` on the same element is **not recommended**. See the [style guide](/v2/style-guide/#Avoid-v-if-with-v-for-essential) for further information.</p>
> 
> If instead, your intent is to conditionally skip execution of the loop, you can place the `v-if` on a wrapper element (or [`<template>`](conditional.html#Conditional-Groups-with-v-if-on-lt-template-gt)). For example:

sdras:
Thanks for the PR! However, I disagree with this change. This is a really important caveat and it should be introduced as soon as possible in the docs as it does create confusion. Furthermore, there's some detail missing in your refactor. Appreciated, though!

PR #1933, opened 2019-01-11
Change Vuetron link

artn:
Since vuetron.io domain was expired and redirects to some news or malware sites, I changed Vuetron link to it's github site.

### DIFF
--- a/src/v2/cookbook/debugging-in-vscode.md
+++ b/src/v2/cookbook/debugging-in-vscode.md
@@ -102,7 +102,7 @@ There are other methods of debugging, varying in complexity. The most popular an
 
 ### Vuetron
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks @artn! I took a look at the domain and indeed it's redirecting to a malware host. However, I'm kind of tempted to remove Vuetron altogether, as the project doesn't look very active (the fact that the domain expired at least more than a month ago and [no one seems to care](https://github.com/vuetwo/vuetron/issues/94) is among the indications). @sdras What do you think?

phanan:
Pinging @yyx990803.

yyx990803:
Agreed, let's remove it.

season19840122:
调试不了小程序

PR #1937, opened 2019-01-14
Clarify null and undefined values on typed props

### LINKED ISSUE #1874: Clarify behavior with null values on prop validation
Hi! 👋 This issue (https://github.com/vuejs/vue/issues/9023) reminded me of a discussion we had in our team regarding the [Prop Validation](https://github.com/vuejs/vuejs.org/blob/53b11fbbef932f3ba2cf0596fe198d6be8d70e23/src/v2/guide/components-props.md#prop-validation) doc, where people who started working with Vue found hard to understand the following statement:

```js
Vue.component('my-component', {
  props: {
    // Basic type check (`null` matches any type)
    propA: Number,
//...
```

Does the comment mean that defining the prop as `propA: null` will match any type of prop?
Or that passing `:prop-a=null` in the template would work, no matter the type we define in the prop declaration?

I know it is a small nuance :)

If you find this is an issue worth solving, I'd happily submit a PR.

Thanks for your work!

afontcu:
Hi! 👋 This closes #1874.

### DIFF
--- a/src/v2/guide/components-props.md
+++ b/src/v2/guide/components-props.md
@@ -188,7 +188,7 @@ To specify prop validations, you can provide an object with validation requireme
 ``` js
 Vue.component('my-component', {
   props: {
-    // Basic type check (`null` matches any type)
+    // Basic type check (`null` and `undefined` values will pass any type validation)
     propA: Number,
     // Multiple possible types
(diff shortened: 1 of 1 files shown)


sdras:
Great, thank you, I think this is more clear

PR #1939, opened 2019-01-15
[In Review] 2.6 Updates

yyx990803:


src/v2/api/index.md
```diff
@@ -1973,6 +2008,8 @@ type: api
   </div>
   ```
 
+  Since 2.6, `v-for` can also work on values that implements the [Iterable Protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol), including native `Map` and `Set`. However it should be noted that Vue 2.x currently does not support reactivity detection on `Map` and `Set` values.
```
Justineo:
```suggestion
  Since 2.6, `v-for` can also work on values that implement the [Iterable Protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterable_protocol), including native `Map` and `Set`. However it should be noted that Vue 2.x currently does not support reactivity detection on `Map` and `Set` values.
```

src/v2/guide/installation.md
```diff
@@ -68,12 +82,12 @@ Vue provides an [official CLI](https://github.com/vuejs/vue-cli) for quickly sca
 
 In the [`dist/` directory of the NPM package](https://cdn.jsdelivr.net/npm/vue/dist/) you will find many different builds of Vue.js. Here's an overview of the difference between them:
 
-| | UMD | CommonJS | ES Module |
+| | UMD | CommonJS | ES Module (for bundlers) | ES Module (for browsers)
```
Justineo:
```suggestion
| | UMD | CommonJS | ES Module (for bundlers) | ES Module (for browsers) |
```

src/v2/guide/installation.md
```diff
@@ -68,12 +82,12 @@ Vue provides an [official CLI](https://github.com/vuejs/vue-cli) for quickly sca
...
 | --- | --- | --- | --- |
```
Justineo:
```suggestion
| --- | --- | --- | --- | --- |
```

src/v2/guide/render-function.md
```diff
@@ -536,6 +536,7 @@ Everything the component needs is passed through `context`, which is an object c
 - `props`: An object of the provided props
 - `children`: An array of the VNode children
 - `slots`: A function returning a slots object
+- `scopedSlots`: (2.6.0+) An object that exposes passed-in scoped slots
```
Justineo: Shall we add note about this includes `slots` as functions?

src/v2/guide/syntax.md
```diff
@@ -152,6 +152,18 @@ The `v-` prefix serves as a visual cue for identifying Vue-specific attributes i
...
+> New in 2.6
+
+In 2.6 a separate shorthand for explicit DOM property bindings (with the `.prop` modifier) have been introduced:
+
+``` html
+<!-- full syntax -->
+<video v-bind:muted.prop="isMuted"> ... </video>
+
+<!-- shorthand -->
+<video .muted="isMuted"> ... </video>
```
Justineo: TBH I think this shorthand somewhat breaks the existing unified syntax of Vue's directives:

```
v-directive:argument.modifier="value"
```

Where `:` always follows with an argument and `.` always follows with a modifier. `.muted` along may look like a modifier without directive/argument part and might be somewhat confusing if users haven't got to this part of the documentation yet.

yyx990803: A modifier would only be a modifier if it has something to modify, right? So a standalone modifier doesn't even make sense.

src/v2/guide/render-function.md
yyx990803: yes

src/v2/guide/components-slots.md
```diff
@@ -85,145 +115,288 @@ For these cases, the `<slot>` element has a special attribute, `name`, which can
...
 ```
 
-If the slot is provided content by the parent, it will replace the default content.
+Here you can name `slotProps` anything you like, just like declaring a function argument in JavaScript. The declared variable will become available inside the scope of the slot's `<template>`.
 
-## Compilation Scope
+Slots that expect props are also known as "scoped slots", as they create a "scope" within which the received props can be used.
 
-When you want to use data inside a slot, such as in:
+<p class="tip">The `v-slot` syntax is introduced in 2.6.0 and is not supported in older versions. The legacy `slot` and `slot-scope` syntax is still supported in 2.6, and is documented [here](#Legacy-Syntax).</p>
```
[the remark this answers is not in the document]
chrisvfritz: Done. 👍

src/v2/api/index.md
```diff
@@ -2389,6 +2377,40 @@ Used to denote a `<template>` element as a scoped slot, which is replaced by [`s
   - [Dynamic Components](../guide/components.html#Dynamic-Components)
   - [DOM Template Parsing Caveats](../guide/components.html#DOM-Template-Parsing-Caveats)
 
+### slot <sup>replaced</sup>
```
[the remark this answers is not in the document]
chrisvfritz: Done. 👍

yyx990803:
@chrisvfritz I've reverted it now in this PR and hidden it behind a flag in core (disabled by default).

yyx990803:
@Justineo added clarification for that case.

chrisvfritz:
@yyx990803 Sounds good. 👍

PR #1941, opened 2019-01-16
Automatic Global Registration of Base Components example does not work…

twickstrom:
Automatic Global Registration of Base Components example does not work if you wanted to include subfolders. This solution has been tested and confirmed. https://stackoverflow.com/questions/53676837/registering-components-globally-in-vuejs-in-subfolders/53678945#53678945

### DIFF
--- a/src/v2/guide/components-registration.md
+++ b/src/v2/guide/components-registration.md
@@ -195,9 +195,9 @@ const requireComponent = require.context(
   // The relative path of the components folder
   './components',
   // Whether or not to look in subfolders
-  false,
+  true,
   // The regular expression used to match base component filenames
-  /Base[A-Z]\w+\.(vue|js)$/
+  /[A-Z]\w+\.(vue|js)$/
 )
 
 requireComponent.keys().forEach(fileName => {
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for your contribution, but please focus on only one feature/fix per PR while possible. Yours have 3: one code change with comment still kept, one (incorrect) regex modification, and one unnecessary style fix.

phanan:
Sorry, I make a mistake :) Reopening.

chrisvfritz:
@twickstrom I think it's best to just demonstrate the simplest and most common solution necessary, since all we're trying to do is demonstrate the technique so that people can modify it as they'd like for their own projects. It's quite rare from what I've seen for base components to be spread throughout nested subfolders and what you're doing is globally registering _all_ components, rather than just base components, which we don't recommend.

twickstrom:
Chris, I do not think anyone would think that registering all components globally would be a good idea.

This update to the docs was meant to show an example of the JS needed that would work for either including subfoldersor or not. The current example fails and then gives the perception that the docs are buggy to the community. See the stack overflow link for reference.

In large enterprise applications it is very common to have many components that are used through out the system. Putting them in a folder called components/base and then organizing them based on functional grouping or application purposes seems perfectly reasonable to me.

Thanks for considering this commit

PR #1942, opened 2019-01-16
Update links to new JSX plugin

nickmessing:
At the moment documentation points to old JSX plugin, however new plugin was release and is the default choice in Vue CLI.


phanan:
👍

PR #1944, opened 2019-01-17
Update dockerize cookbook

DRoet:
Node 9 has been EOL for a while, this will make sure people will always use the latest LTS version, instead of a hardcoded version number

motivation: https://github.com/sass/node-sass/issues/2564#issuecomment-451254080


phanan:
Thanks!

PR #1946, opened 2019-01-17
Add comments about $watch behavior with a function

posva:
I realised we are not explicit about this behaviour so I added some comments. I wonder if we should add a full example instead saying it in a comment. We could say that if we got `a === 2` and `b === 0` and we change both variables to `b = 2; a = 0`, the watcher won't get triggered

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1486,6 +1486,9 @@ type: api
   // function
   vm.$watch(
     function () {
+      // everytime the expression `this.a + this.b` yields a different result,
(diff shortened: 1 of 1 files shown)


phanan:
I think with the API doc already saying

> Watch an expression or a computed function on the Vue instance for changes. The callback gets called with the new value and the old value. The expression only accepts dot-delimited paths. **For more complex expressions, use a function instead.**

your added comments should do the job well enough. Taking the liberty to merge now :)

PR #1952, opened 2019-01-21
Update debugging-in-vscode.md

PSausM:
I think there is an error. It works only with the dot


phanan:
Pinging @sdras, as I'm not familiar with this topic.

sdras:
This is a duplicate of a PR that was already merged, but thanks!

PR #1953, opened 2019-01-22
Reword heading in slot docs to be a bit more specific

psalaets:
I was reading through the slot documentation from top to bottom and when I got to the heading [Default Slot Content](https://vuejs.org/v2/guide/components-slots.html#Default-Slot-Content) I had to pause for a second because I wasn't sure if it was talking about "content of the default slot" vs "default content of slots".

I think this confusion happened because
- I had just read about the Default Slot in the previous paragraph
- Heading could be interpreted either way 

My change is an attempt to remove the small ambiguity.

### DIFF
--- a/src/v2/guide/components-slots.md
+++ b/src/v2/guide/components-slots.md
@@ -132,7 +132,7 @@ There can still be one unnamed slot, which is the **default slot** that serves a
 </div>
 ```
 
-## Default Content of Slots
+## Default Content for Slots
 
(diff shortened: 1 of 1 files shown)


phanan:
👍 I agree, the title is a bit ambiguous. I'd suggest using "Default Content _for_ Slots" though, as it would better reflect the intent of the section. Does it make sense?

psalaets:
> I'd suggest using "Default Content for Slots" though, as it would better reflect the intent of the section. Does it make sense?

Yep, I pushed another commit as per your comment

phanan:
Thanks!

PR #1956, opened 2019-01-23
Clarify that "slot-scope" is named "scope" before 2.5.0

### LINKED ISSUE #1955: Clarify in documentation that "slot-scope" is named "scope" before 2.5.0
The [guide](https://vuejs.org/v2/guide/components-slots.html#Scoped-Slots) about Scoped Slots says:

> In 2.5.0 , `slot-scope` is no longer limited to the `<template>` element, but can instead be used on any element or component in the slot.

But it is not says, that until 2.5.0, `slot-scope` doesn't work at all, even inside `<template>` element. To make it work, It must be `scope` inside `<template>` element.

It will be good to mention in the same warning element in documentation, that `slot-scope` not works until 2.5.0 and `scope` must be used instead.

Something, like:

> Until 2.5.0, `slot-scope` was named `scope` and was limited to the `<template>` element.
> In 2.5.0 , `slot-scope` is no longer limited to the `<template>` element, but can instead be used on any element or component in the slot.

I spend several hours trying to understand why it says that this functionality should work from 2.1.0+, but it doesn't work until 2.5.0. And the reason: it works! Just `slot-scope` should be `scope`. :)

Related issue: #1223 but it fixed it only in API docs (not in the guide).
Related commit: https://github.com/vuejs/vue/commit/dae173d96d15f47de6ce6961354d5c05e4273005

xak2000:
Fixes #1955

### DIFF
--- a/src/v2/guide/components-slots.md
+++ b/src/v2/guide/components-slots.md
@@ -211,9 +211,7 @@ Now when we use the `<todo-list>` component, we can optionally define an alterna
 </todo-list>
 ```
 
-> Until 2.5.0, `slot-scope` was named `scope` and was limited to the `<template>` element.
-
-> In 2.5.0+, `slot-scope` is no longer limited to the `<template>` element, but can instead be used on any element or component in the slot.
+> Before 2.5.0, `slot-scope` was named `scope` and was limited to the `<template>` element. Since 2.5.0, `slot-scope` doesn't have this limitation and can instead be used on any element or component in the slot.
 
 ### Destructuring `slot-scope`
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/components-slots.md
```diff
@@ -211,6 +211,8 @@ Now when we use the `<todo-list>` component, we can optionally define an alterna
 </todo-list>
 ```
 
+> Until 2.5.0, `slot-scope` was named `scope` and was limited to the `<template>` element.
```
phanan: Instead of using 2 quotes, I'd suggest merging them into something like:

> Before 2.5.0, `slot-scope` was named `scope` and was limited to the `<template>` element. Since 2.5.0, `slot-scope` doesn't have this limitation and can instead be used on any element or component in the slot.

phanan:
Thanks

PR #1959, opened 2019-01-25
Globally register components example

twickstrom:
Updated regex in global component registration code example to allow setting subfolder to true


twickstrom:
Simplified the changes to address the concerns of @chrisvfritz in https://github.com/vuejs/vuejs.org/pull/1941 while still providing a solution that will work if someone has base components within subfolders of /components.

phanan:
Pinging @chrisvfritz.

sdras:
I agree this is an improvement. If Chris disagrees we can always revert :)

PR #1968, opened 2019-01-29
Use highlight.js

### LINKED ISSUE #1954: Avoid translation of sample code fragments with Google Translate
The sample code fragments enclosed with `<pre>` tags are translated along with the rest of the document when using translation tools such as Google Translate.

Please, it's important to solve this for people whose languages ​​are not available in the documentation. It's easily solved by enclosing the code with a `<code>` tag in place or together with the `<pre>` tag. This would also give you correct semantics. If for some reason you do not want to use the `<code>` tag, the alternative option would be to use the class `notranslate` or `skiptranslate` on the `<pre>` tag, like this: `<pre class="notranslate">` or `<pre class="skiptranslate">`. Both solutions work correctly, avoiding translation.

(Sorry for my english...)

phanan:
This fixes #1954 by enabling the use of highlight.js. The generated code will be more semantics e.g. 

image

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -39,6 +39,7 @@ highlight:
   enable: true
   line_number: false
   tab_replace:
+  hljs: true
 
 # Category & Tag
 default_category: uncategorized
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vuejs.org",
   "private": true,
   "hexo": {
-    "version": "3.7.1"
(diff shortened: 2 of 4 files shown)


phanan:
@espagueti Can you take a look at the Netlify deploy and confirm that this works as expected? Cheers!

espagueti:
Perfect! now it works as it should and has the correct semantics

PR #1969, opened 2019-01-29
Add clarification about inheritAttrs and class/style bindings (#1787)

### LINKED ISSUE #1787: The example given for "Disabling Attribute Inheritance" doesn't work

leopiccionia:
It closes #1787, and overrides #1796.

It adds a note to the guide, clarifying that `inheritAttrs: false` doesn't affect `class` and `style` bindings, and updates a related misleading example (`class` doesn't ever appear in `$attrs`, [see fiddle](https://jsfiddle.net/leopiccionia/1zdtvbjx/)).

cc @phanan

### DIFF
--- a/src/v2/guide/components-props.md
+++ b/src/v2/guide/components-props.md
@@ -313,7 +313,7 @@ This can be especially useful in combination with the `$attrs` instance property
 
 ```js
 {
-  class: 'username-input',
+  required: true,
   placeholder: 'Enter your username'
 }
 ```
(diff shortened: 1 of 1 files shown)


phanan:
This looks great, thanks!

PR #1970, opened 2019-01-30
Use Enter and submit in keyboardEvent example

phanan:


Jinjiang:
Below is my total version suggestion from loc 229 (better?):

----

```
<!-- only call `vm.submit()` when the `key` is `Enter` -->
<input v-on:keyup.enter="submit">
```

You can directly use any valid key names exposed via [`KeyboardEvent.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values) as modifiers by converting them to kebab-case.

``` html
<input @keyup.page-down="onPageDown">
```

In the above example, the handler will only be called if `$event.key === 'PageDown'`.

### Key Codes

...

PR #1972, opened 2019-01-31
Remove Handlebars reference (fixes 1780)

### LINKED ISSUE #1780: Conditional Rendering starts with comparison to unrelated technology
From: https://vuejs.org/v2/guide/conditional.html#v-if

Beginning an explanation of functionality by comparison with unrelated, possibly unfamiliar technology may lead to more mystification than it alleviates.
Direct description of the operating principle, perhaps followed by a salient comparison with a related technology to mention any distinguishing behaviour may be more helpful

phanan:
Following up and fixes #1780. This removes the (mildly sudden) reference to Handlebars by rewriting the directive's explanation.

### DIFF
--- a/src/v2/guide/conditional.md
+++ b/src/v2/guide/conditional.md
@@ -6,26 +6,17 @@ order: 7
 
 ## `v-if`
 
-In string templates, for example [Handlebars](https://handlebarsjs.com/), we would write a conditional block like this:
+The directive `v-if` is used to conditionally render a block. The block will be rendered if the directive's expression returns a truthy value and vice versa.
 
 ``` html
-<!-- Handlebars template -->
-{{#if ok}}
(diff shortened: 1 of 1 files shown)


src/v2/guide/conditional.md
```diff
@@ -6,26 +6,17 @@ order: 7
...
+<h1 v-if="awesome">Vue is awesome!</h1>
 ```
 
 It is also possible to add an "else block" with `v-else`:
 
 ``` html
-<h1 v-if="ok">Yes</h1>
-<h1 v-else>No</h1>
+<h1 v-if="awesome">Vue is awesome!</h1>
+<h1 v-else>Oh no 😢</h1>
```
sdras: Nice, I like this example better :)

PR #1974, opened 2019-01-31
Clarified v-model props/events for native elements

### LINKED ISSUE #1965: Clarify v-model in the docs
### What problem does this feature solve?
On multiple occasions, the docs clearly stated that `v-model on a component uses value as the prop and input as the event` ([here](https://vuejs.org/v2/guide/components-custom-events.html#Customizing-Component-v-model) and [there](https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components)).

On standards inputs, the documentation ([here](https://vuejs.org/v2/guide/forms.html)) don't explain what can be expected from `v-model`, and how to replace its default behavior (when writting wrappers, for example). I have found that:

* on `input[type=text]` and `textarea`, v-model uses `value` as the prop and `input` as the event
* on `input[type=radio]` and `input[type=checkbox]`, v-model uses `checked` as the prop and `input` as the event — but seems to collide with `value` ([#8379](https://github.com/vuejs/vue/issues/8379))
* on `select`, v-model uses `value` as the prop and `change` as the event

Some issues are related to this:
* dynamic input value with radio buttons don't work in IE11 [#8379](https://github.com/vuejs/vue/issues/8379) — related to [this section of the docs](https://vuejs.org/v2/guide/forms.html#Radio-1) which breaks the default IE9+ compatibility
* <select @input="..."> doesn't work in IE11 but works with v-model [#4701](https://github.com/vuejs/vue/issues/4701)

May you please consider detailing for each control (input, radio, checkbox, select, textarea) what is the  expected behaviour of v-model? This would be of great help!

### What does the proposed API look like?
--

Related to [#9370](https://github.com/vuejs/vue/issues/9370)

NataliaTepluhina:
Added description for `v-model` using different props and emitting different event depending on input type. 

Closes #1965

### DIFF
--- a/src/v2/guide/forms.md
+++ b/src/v2/guide/forms.md
@@ -10,7 +10,10 @@ You can use the `v-model` directive to create two-way data bindings on form inpu
 
 <p class="tip">`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the Vue instance data as the source of truth. You should declare the initial value on the JavaScript side, inside the `data` option of your component.</p>
 
-<p class="tip">`v-model` internally uses different properties and emits different events for different input elements. For text and textarea elements it's `value` property and `input` event; for checkboxes and radiobuttons it's `checked` property and `change` event; and for select fields  `v-model` uses `value` as a prop and `change` as an event.</p>
+`v-model` internally uses different properties and emits different events for different input elements:
+- text and textarea elements use `value` property and `input` event;
+- checkboxes and radiobuttons use `checked` property and `change` event;
+- select fields use `value` as a prop and `change` as an event.
 
 <p class="tip" id="vmodel-ime-tip">For languages that require an [IME](https://en.wikipedia.org/wiki/Input_method) (Chinese, Japanese, Korean etc.), you'll notice that `v-model` doesn't get updated during IME composition. If you want to cater for these updates as well, use `input` event instead.</p>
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/forms.md
```diff
@@ -10,6 +10,8 @@ You can use the `v-model` directive to create two-way data bindings on form inpu
 
 <p class="tip">`v-model` will ignore the initial `value`, `checked` or `selected` attributes found on any form elements. It will always treat the Vue instance data as the source of truth. You should declare the initial value on the JavaScript side, inside the `data` option of your component.</p>
 
+<p class="tip">`v-model` internally uses different properties and emits different events for different input elements. For text and textarea elements it's `value` property and `input` event; for checkboxes and radiobuttons it's `checked` property and `change` event; and for select fields  `v-model` uses `value` as a prop and `change` as an event.</p>
```
sdras: Two things here:
1) I'm not sure we want to do two tips in a row- we've talked about how we should either join them and use bullets in this case or take one out of tips. In this case, this might be beyond a tip and could be refactored into text with bullet points for each case
2) `'it's` is wrong in this case, we would just use `its` :)

sdras:
This is great! Thanks Natalia :) I added a quick review.

src/v2/guide/forms.md
NataliaTepluhina: @sdras thank you so much for the review! I added a commit to fix mentioned issues

sdras:
Awesome, thanks!

PR #1982, opened 2019-02-02
Prop Validation - add Array default value

ramiy:
Add prop validation for an Array type with a default value using an anonymous function

### DIFF
--- a/src/v2/guide/components-props.md
+++ b/src/v2/guide/components-props.md
@@ -212,8 +212,6 @@ Vue.component('my-component', {
     // Object with a default value
     propF: {
       type: Object,
-      // Object or array defaults must be returned from
-      // a factory function
       default: function () {
         return { message: 'hello' }
       }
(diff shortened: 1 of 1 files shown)


Justineo:
https://github.com/vuejs/vuejs.org/blob/5e1911b73ca0b7866436ae3c0daae4a73ee8fb3b//src/v2/guide/components-props.md#L208-L209

ramiy:
@Justineo This comments should not be displayed in the code example, it should be part of the content, maybe a "Tip" style.

ramiy:
@Justineo The code was updated.

The PR adds an example for Array defaults and moves the code comment to the content area.

ramiy:
@Justineo ?

sdras:
Thanks for the PR! I don't really think this change is necessary, it's pretty straightforward now, and this means a lot more reading and scanning. We would also then have two "tips" next two eachother which we don't do. Appreciated, though!

Issue #863, opened 2017-04-07
Arabic Translation


chadidi:
We've just started the translation of `vuejs.org`

https://github.com/LaravelArab (Laravel Framework Arab Community)
https://github.com/LaravelArab/vuejs.org (VueJs Repo)
We'll keep you up-to-date when the project has more updates.

Thank you. (contributors are always welcome).
contact : @chadidi

kazupon:
Welcome!

phanan:
I'm closing this issue as an effort to tidy our issue list. Feel free to open another if you think it's a mistake (and sorry for that 🙂).

PR #1983, opened 2019-02-03
Translation to Arabic : update file - README.md

celyes:
Translation to Arabic : update file - README.md

@Aissaoui-ahmed @InterstellarClub

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# vuejs.org
+﻿# vuejs.org
 
 This site is built with [hexo](http://hexo.io/). Site content is written in Markdown format located in `src`. Pull requests welcome!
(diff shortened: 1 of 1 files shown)


phanan:
Thanks. How is this different from https://github.com/LaravelArab/vuejs.org (taken from #863)?

celyes:
> 
> 
> Thanks. How is this different from https://github.com/LaravelArab/vuejs.org (taken from #863)?
Thank you for your response.
the Arabic translation process is currently not active. the process might take a long period of time if it continues in such a progress and that's why we've decided to do the job.

sdras:
Ok! Thank you!

PR #1987, opened 2019-02-06
Fix typo in new slots shorthand syntax

btoo:


sdras:
Nice catch, thank you

PR #1990, opened 2019-02-11
Add negative example warning

### LINKED ISSUE #1986: Negative example of a Custom Event Name might cause confusion
I'm working with Vue and hit a problem that had me stumped for a few hours. I was using the documentation as a loose guide to emit and receive custom events, however the first example under Event Names [here](https://vuejs.org/v2/guide/components-custom-events.html#Event-Names) illustrates how _not_ to name both the emitter and the receiving handler. Skim reading the documentation caused me to name the emitter using camel case, and the receiver as kebab case, just working off what I thought was the example.

Perhaps it would be better to illustrate naming these correctly initially, then mention the lack of automatic case transformation afterwards as a common pitfall? Appreciate that fully reading the paragraphs surrounding the example make this clear, however I would suspect a proportion of developers would skim over that like I did.

afontcu:
This closes #1986.

I suggest a simple HTML comment to warn readers about the negative example. I saw a similar usage in the [Slots](https://vuejs.org/v2/guide/components-slots.html#Abbreviated-Syntax-for-Lone-Default-Slots) page.

However, the whole section could be rearranged differently, by providing a better context before showing the example. We could even use a tip.

Any thoughts?
Thanks!

### DIFF
--- a/src/v2/guide/components-custom-events.md
+++ b/src/v2/guide/components-custom-events.md
@@ -17,6 +17,7 @@ this.$emit('myEvent')
 Listening to the kebab-cased version will have no effect:
 
 ```html
+<!-- Won't work -->
 <my-component v-on:my-event="doSomething"></my-component>
 ```
 


sdras:
Interesting, I wouldn't have thought this would be an issue, but I appreciate the reference to the issue and also an existing example. If it helps people grok things faster, I don't see the harm. Thanks!

PR #1996, opened 2019-02-13
Updated el and $el from HTMLElement to Element

robinchow:
As per the Typescript definitions plus this merged PR: https://github.com/vuejs/vue/pull/8809


sdras:
Yes, thanks for this!

PR #2002, opened 2019-02-16
Update onUpdate callback.

atjason:
Seems the code in page is out of date, and is different from the real code in html source.


sdras:
Thanks!

PR #2012, opened 2019-02-18
Fixed grid component example with different prop name

### LINKED ISSUE #1739: Grid component example uses a confusing prop naming scheme.
The [grid component example](https://vuejs.org/v2/examples/grid-component.html) has a `demo-grid` component that accepts a prop named as `data` which could be quite confusing, as witnessed in the forum post: https://forum.vuejs.org/t/help-understanding-example-grid-component-data/40161

NataliaTepluhina:
The [grid component example](https://vuejs.org/v2/examples/grid-component.html) had a `demo-grid` component that accepts a prop named as `data` and if was probably confusing to new users. I changed a fiddle, here is a link to [new one](https://jsfiddle.net/Tertia/vbyon64p/6)

Close #1739

### DIFF
--- a/src/v2/examples/grid-component.md
+++ b/src/v2/examples/grid-component.md
@@ -6,4 +6,4 @@ order: 3
 
 > This is an example of creating a reusable grid component and using it with external data.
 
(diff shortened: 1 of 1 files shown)


phanan:
Love it!

PR #2014, opened 2019-02-19
Placed download button on the home page

### LINKED ISSUE #1782: Suggestion: place download link on introduction page
Downloading Vue requires navigating through three pages: from the home page to _Get Started_, and from _Get Started_ to _Installation_.

It would be nice to have a download link and the URL of a CDN right there on the home page, similar to what other JS library hosts are doing (e.g., Bootstrap, JQuery, Rivets).

NataliaTepluhina:
As mentioned in #1782, installing Vue requires navigating through three pages: from the home page to _Get Started_, and from _Get Started_ to _Installation_.

This PR adds `Install` button on the home page and moves GitHub link to the menu.

@phanan could you please have a look at this draft? 🙏 It will probably need some style fixes

Close #1782


themes/vue/layout/partials/main_menu.ejs
```diff
@@ -10,3 +10,7 @@
 </li>
 <%- partial('partials/support_vue_dropdown') %>
 <%- partial('partials/language_dropdown') %>
+<a class="button nav-button gray has-icon" href="https://github.com/vuejs/vue" target="_blank">
```
phanan: As this a direct child of `<ul>`, I believe it should be wrapped inside a `<li>`.

NataliaTepluhina: Thank you for spotting it 😅 pushed a visual fix (I still don't like how it looks though, suggestions welcome!)

phanan:
Much better! Do you think we can achieve something like this with pure CSS though? IMHO this is good enough for the mobile layout.

NataliaTepluhina:
@phanan sure, I will tweak CSS not to override `button` styles then

NataliaTepluhina:
@phanan fixed GitHub link styles on small screen resolutions

phanan:
It's near perfect, but 😅 the link still has a gray background and is without a proper hover status (even though this is meant to be for mobile/tablet devices).

NataliaTepluhina:
@phanan refactored `nav-button` styles :sweat_smile:

phanan:
@sdras @chrisvfritz I think we can try this. WDYT?

yyx990803:
I'm not sure about this. Having a prominent link to GitHub also feels quite important.
I think a more reasonable workaround is adding a more prominent link to the installation page on the introduction page.

sdras:
Yeah, I agree with Evan- I think we do want to keep Github on the first page, but move a green button for installation to the top of the introduction page: (like the buttons in this section) https://vuejs.org/v2/guide/installation.html#Direct-lt-script-gt-Include

IMO it would be great right under the heading for **Getting Started**
https://vuejs.org/v2/guide/#Getting-Started

Thanks for your effort Natalia!

PR #2024, opened 2019-02-21
fixed camelCase in dynamic HTML attribute example and noted constraint

procyclinsur:
I was running through the documentation performing the examples, and noticed that the Dynamic HTML Attribute examples were not working due to camelCase in the `v-bind:[camelCase]="something"` examples.

on further examination I noticed that kebab case similarly does not work, but that snake_case does.

I changed the examples to snake_case, and added a blip in the constraint section for Dynamic HTML Attribute.

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -135,18 +135,18 @@ Here the argument is the event name to listen to. We will talk about event handl
 Starting in version 2.6.0, it is also possible to use a JavaScript expression in a directive argument by wrapping it with square brackets:
 
 ``` html
-<a v-bind:[attributeName]="url"> ... </a>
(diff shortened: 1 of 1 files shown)


phanan:
@procyclinsur Thanks for the PR! The issue [has just been solved](https://github.com/vuejs/vuejs.org/commit/50589c52d89e06a4ee14c950d550c6f235125940) by @yyx990803, though, so I'm closing it.

procyclinsur:
@phanan Thank you for informing me of the change!

I am glad to see that the issue has been resolved, and the new documentation by @yyx990803 mentioned the issue exists with in-DOM templates, which I only assume means that there is another kind of template that allows for camelCase which I look forward to discovering.

PR #2026, opened 2019-02-22
Add Function type to props

posva:
Hi, I realized we are not showing we can accept a Function which is a basic type in JS, it has misled people to believe they cannot use Functions.

However, I believe we can do better than just adding the line with callback: Function because we can also specify custom constructors or new ones like Promise

Looking forward to hear your feedback


phanan:
👍 Even though a Function or a Promise is just an Object (which is already there), being explicit here is a good idea.

PR #2028, opened 2019-02-23
Added options for object-based props definition in API

### LINKED ISSUE #2027: Document all options for object-based props definition in API
This issue is a consequence of a conversation I had on the forum:

https://forum.vuejs.org/t/where-are-prop-options-documented/56692/3?u=linusborg

The user raised valid points, namely that the API docs in that area don't acually document the options that are avaliable except for the brief example that's given. For details about these options, the reader is forwarded to the Props Guide, but information here is spread across multiple sections in the guide, making it hard to get the full picture of what options are available and what their syntax is.

So we should clearly and still briefly document all options that are available in the API docs section:

* type
* default
* required
* validator

NataliaTepluhina:
I created a draft API entry describing all possible options we can have with Object-syntax prop. Also, I fixes a link @phanan mentioned in #2027.

Close #2027

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -514,7 +514,7 @@ type: api
     - **default:** `any`
     Specifies a default value for the prop. If the prop is not passed, this value will be used instead. Object or array defaults must be returned from a factory function.
     - **required:** `Boolean`
-    Defines if prop is required. When set to `true`, will produce a console warning if prop is not passed.
+    Defines if the prop is required. In a non-production environment, a console warning will be thrown if this value is truthy and the prop is not passed.
     - **validator:** `Function`
     Custom validator function that takes the prop value as the sole argument. In a non-production environment, a console warning will be thrown if this function returns a falsy value (i.e. the validation fails). You can read more about prop validation [here](../guide/components-props.html#Prop-Validation).
 
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -516,7 +516,7 @@ type: api
     - **required:** `Boolean`
     Defines if prop is required. When set to `true`, will produce a console warning if prop is not passed.
     - **validator:** `Function`
-    Custom validator function that takes the prop value and returns a boolean. When returns `false` (i.e. when prop fails validation), throws a console warning. You can read more about props validation [here](../guide/components-props.html#Prop-Validation).
+    Custom validator function that takes the prop value as the sole argument. In a non-production environment, a console warning will be thrown if this function returns a falsy value (i.e. the validation fails). You can read more about prop validation [here](../guide/components-props.html#Prop-Validation).
 
 - **Example:**
 
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -510,15 +510,13 @@ type: api
   A list/hash of attributes that are exposed to accept data from the parent component. It has an Array-based simple syntax and an alternative Object-based syntax that allows advanced configurations such as type checking, custom validation and default values.
 
   With Object-based syntax, you can use following options:
-    - **type:** can be one of the following native constructors: `String`, `Number`, `Boolean`, `Array`, `Object`, `Date`, `Function`, `Symbol` or any custom constructor function. Will check if a prop has a given type, and will throw a warning if it doesn't.
+    - **type:** can be one of the following native constructors: `String`, `Number`, `Boolean`, `Array`, `Object`, `Date`, `Function`, `Symbol`, any custom constructor function or an array of those. Will check if a prop has a given type, and will throw a warning if it doesn't. [More information](../guide/components-props.html#Prop-Types) on prop types.
     - **default:** `any`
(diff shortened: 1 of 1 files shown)


src/v2/api/index.md
```diff
@@ -509,6 +509,17 @@ type: api
 
   A list/hash of attributes that are exposed to accept data from the parent component. It has an Array-based simple syntax and an alternative Object-based syntax that allows advanced configurations such as type checking, custom validation and default values.
 
+  With Object-based syntax, you can use following options:
+    - **type:** can be one of the following native constructors: `String`, `Number`, `Boolean`, `Array`, `Object`, `Date`, `Function`, `Symbol` or any custom constructor function. Will check if a prop has a given type, and will throw a warning if it doesn't.
```
LinusBorg: It can also be an array of those, like
```javascript
type: [String, Number]
```

Consider adding a `"More"` link to: https://vuejs.org/v2/guide/components-props.html#Prop-Types

src/v2/api/index.md
```diff
@@ -509,6 +509,17 @@ type: api
...
+    - **default:** `any`
+    Specifies a default value for the prop. If prop is not passed, the `default` value will be used instead. Object or array defaults must be returned from a factory function.
+    - **required:** `Boolean`
+    Defines if prop is required. When set to `true`, will produce a console warning if prop is not passed.
+    - **validator:** `Function`
+    Custom validator function that takes the prop value and returns a boolean. When returns `false` (i.e. when prop fails validation), throws a console warning.
+
+  You can read more about props validation [here](../guide/components-props.html#Prop-Validation).
```
LinusBorg: I think this sentence should be part of the paragraph above, about the validator. The way it is right now it seems to be a sentence detatched from the list of options we saw before, and not related to the last point of that list (`validator`)

src/v2/api/index.md
NataliaTepluhina: Fixed! 👍

NataliaTepluhina:
Thanks for the blazing fast review, @LinusBorg! I've added some requested changes

src/v2/api/index.md
```diff
@@ -509,6 +509,15 @@ type: api
 
   A list/hash of attributes that are exposed to accept data from the parent component. It has an Array-based simple syntax and an alternative Object-based syntax that allows advanced configurations such as type checking, custom validation and default values.
 
+  With Object-based syntax, you can use following options:
+    - **type:** can be one of the following native constructors: `String`, `Number`, `Boolean`, `Array`, `Object`, `Date`, `Function`, `Symbol`, any custom constructor function or an array of those. Will check if a prop has a given type, and will throw a warning if it doesn't. [More information](../guide/components-props.html#Prop-Types) on prop types.
+    - **default:** `any`
+    Specifies a default value for the prop. If prop is not passed, the `default` value will be used instead. Object or array defaults must be returned from a factory function.
```
```suggestion
    Specifies a default value for the prop. If the prop is not passed, this value will be used instead. Object or array defaults must be returned from a factory function.
```

src/v2/api/index.md
```diff
@@ -509,6 +509,15 @@ type: api
...
+    - **required:** `Boolean`
+    Defines if prop is required. When set to `true`, will produce a console warning if prop is not passed.
```
```suggestion
    Defines if the prop is required. In a non-production environment, a console warning will be thrown if this value is truthy and the prop is not passed.
```

src/v2/api/index.md
```diff
@@ -509,6 +509,15 @@ type: api
...
+    - **validator:** `Function`
+    Custom validator function that takes the prop value and returns a boolean. When returns `false` (i.e. when prop fails validation), throws a console warning. You can read more about props validation [here](../guide/components-props.html#Prop-Validation).
```
```suggestion
    Custom validator function that takes the prop value as the sole argument. In a non-production environment, a console warning will be thrown if this function returns a falsy value (i.e. the validation fails). You can read more about prop validation [here](../guide/components-props.html#Prop-Validation).
```

NataliaTepluhina:
Thanks @phanan! Yes, I agree, I forgot to mention warnings are thrown only on dev environment
Suggestions applied :bowing_woman:

phanan:
Looks great! I think it's good enough to merge now 👍

src/v2/api/index.md
```diff
@@ -509,6 +509,15 @@ type: api
 
   A list/hash of attributes that are exposed to accept data from the parent component. It has an Array-based simple syntax and an alternative Object-based syntax that allows advanced configurations such as type checking, custom validation and default values.
 
+  With Object-based syntax, you can use following options:
+    - **type:** can be one of the following native constructors: `String`, `Number`, `Boolean`, `Array`, `Object`, `Date`, `Function`, `Symbol`, any custom constructor function or an array of those. Will check if a prop has a given type, and will throw a warning if it doesn't. [More information](../guide/components-props.html#Prop-Types) on prop types.
```
Jinjiang: Is it better to wrap the object keys **type**, **default**, **required** and **validator** with backquotes? Because they are kinds of actual code somehow. 🤔
I will give a PR if you all don't mind.
Thanks.

PR #2029, opened 2019-02-23
to reduce misleading, add a note for [v-for with v-if] section

zdsfwy:
When reading the Vue document, I found section [v-for with v-if](https://vuejs.org/v2/guide/list.html#v-for-with-v-if) doesn't include any information about the fact that `v-for` and `v-if` are **not recommended** to be used together. I see in some other [section](https://vuejs.org/v2/guide/conditional.html#v-if-with-v-for) it's kindly mentioned. I think it's necessary to also mention it here to not mislead readers.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -441,6 +441,8 @@ Similar to template `v-if`, you can also use a `<template>` tag with `v-for` to 
 
 ## `v-for` with `v-if`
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks! Do you think we should put the tip at the beginning of the section, in the same fashion as "`v-if` with `v-for`"?

zdsfwy:
Hi @phanan , thanks for the good point! I have updated the PR correspondingly.

sdras:
This looks good! Thank you :)

PR #2035, opened 2019-02-27
add how to construct plugin object

andy840119:
add how to construct plugin object


phanan:
Hmm, I think the current example is fine as it is now. A plugin author should already know about constructing an object in JS :) Not to mention, if we go this route, why not do

```js
var MyPlugin = {
  install = function () {}
}
```

Thanks for the PR, though!

PR #2038, opened 2019-02-28
Fixed smooth scrolling time

### LINKED ISSUE #2037: API navigation is slow due to smoothScroll configuration
Navigating to different [API](https://vuejs.org/v2/api/) entries via the anchor links on the left-hand table of contents (TOC) can be extremely slow.

For example, try clicking one of the top links in the left column, e.g. [#Global-Config](https://vuejs.org/v2/api/#Global-Config), and then scroll to the bottom of the TOC and click a link near the bottom, e.g. [#Server-Side-Rendering](https://vuejs.org/v2/api/#Server-Side-Rendering). I clocked the scroll time at nearly _30 seconds_.

Personally I'd favor just dropping smoothScroll entirely. However, if you wish to keep it, [according to the docs](https://github.com/cferdinandi/smooth-scroll#scroll-speed) you might reconfigure it with the option `speedAsDuration: true` so all scrolls will take the same amount of time.

https://github.com/vuejs/vuejs.org/blob/0d9b754b2f31eb77b2c7767aac2b943a2f8775bd/themes/vue/source/js/common.js#L412

NataliaTepluhina:
Fixing smooth scrolling time as described in #2037 

Close #2037

### DIFF
--- a/themes/vue/source/js/common.js
+++ b/themes/vue/source/js/common.js
@@ -410,6 +410,7 @@
 
       new SmoothScroll('a[href*="#"]', {
         speed: 400,
+        speedAsDuration: true,
         offset: function (anchor, toggle) {
           let dataTypeAttr = anchor.attributes['data-type']
           if(dataTypeAttr && dataTypeAttr.nodeValue === 'theme-product-title') {


phanan:
Biutiful, thanks!

PR #2044, opened 2019-03-04
Clarify computed properties reactive dependencies

Eido95:
It is not obvious that computed properties dependencies **must be reactive** in order to trigger
their referring computed properties.

It is indirectly mentioned a few lines bellow (using the date example),
that's why I classify this commit as "clarification" rather than "addition".


phanan:
Thanks!

PR #2046, opened 2019-03-05
fix Composing with Components

sot528:
It's more understandable than before for me.


phanan:
Thanks but I don't think it's necessary. Not to mention, if we go for this route, we'll need to change in a lot of other examples as well for consistency.

PR #2051, opened 2019-03-08
Use immediate watcher in serverless blog cookbook post

andreasvirkus:
Since the same method gets called in the created hook, we can
add the immediate property to the watcher and remove the hook.


phanan:
I like the idea, but feel the explanation is unnecessary.

andreasvirkus:
Good point. I wasn't sure what the skill level is for these recipes, so thought it safer to add an explanation.

sdras:
Looks good now! Thanks!

PR #2053, opened 2019-03-09
(maybe) improved the builds introduction

LongTengDao:


Justineo:
Please preview your modification locally first. (You can also see the rendered version [here](https://deploy-preview-2053--vuejs.netlify.com/v2/guide/installation.html).)

LongTengDao:
> Please preview your modification locally first. (You can also see the rendered version [here](https://deploy-preview-2053--vuejs.netlify.com/v2/guide/installation.html).)

Thank you! I previewed it in GitHub, maybe here is some differences between the render engines, I will check it.

phanan:
Would you care to explain the improvement(s)?

LongTengDao:
@phanan Oh, thanks.

When I saw this doc, I was confused, though I'm familiar with Vue for three year use.

I guess it's just because of the old document structure didn't take into account the various builds that were added later.

The best solution is to show this naming rule: `vue(.runtime if no need for compiler)(.umd-use|.esm-use|.cjs-pack|esm-pack)(.min if production).js`, but this need react rich HTML to support, I think.

Since that's not supported, I tried to understand the builds deeper -- what does these builds built for, not just the tech differences.

I think there are two major category, directly use and for bundlers. Under that, there are the minor tech different builds, which is more understandable. So I changed the explanation.

However, since you still need to explain after reading the rendered version ([old](https://vuejs.org/v2/guide/installation.html) and [new](https://deploy-preview-2053--vuejs.netlify.com/v2/guide/installation.html)), I may have failed. :)

PR #2061, opened 2019-03-14
Refactor Themes section

phanan:
As discussed, this PR refactors the Themes section out of the current Examples page and into its own page. 

We'll also need the contact email.


phanan:
Ignore commit aa9b567, I mistakenly included it into the PR, but it shouldn't have any bad effect.

yyx990803:
Should the "new" dot be shown for the top level Theme link instead?

phanan:
@yyx990803 Oops, totally missed that. Fixed.

Also, what do you reckon the contact email should be, and how can we make it clearer that this is an affiliation-only page (if necessary)?

bencodezen:
Nice job with theme page!

My only feedback is that putting it in the main navigation is giving it too much prominence especially since it's an affiliation-only page. It almost feels like we're selling ad space on the site.

Instead, I would recommend that we put it in the "Ecosystem" dropdown instead.

phanan:
@bencodezen I thought about exactly that: the ecosystem dropdown. Problem is, this dropdown is already so long that having the Themes item there would be not much better than the current Examples when it comes to exposure, thus defeat the purpose of this refactor in the first place.

bencodezen:
@phanan What if we surface the entire subsection of "Resources" instead since that includes other curated links that has a more open-source vibed rather than affiliates only? That would simultaneously reduce the Ecosystem dropdown list length while helping to better surface everything at once.

What do you think?

yyx990803:
Looks good - can you resolve the conflicts?

phanan:
@yyx990803 Done. I've also moved the Partners page to under Resources as we discussed and fixed some menu/sidebar issues.

PR #2062, opened 2019-03-15
Update custom-directive.md

bk-dev:
Add a tip promoting and explaining dynamic directive arguments. I spent several hours trying to figure out how I could dynamically send arguments to a directive. I stumbled upon a git issue which had syntax which I'd never seen before. This led to me discovering about dynamic arguments. I think this should be within the Custom Directive guide so others do not have to face the same issue as me.

### DIFF
--- a/src/v2/guide/custom-directive.md
+++ b/src/v2/guide/custom-directive.md
@@ -143,6 +143,7 @@ new Vue({
 </script>
 {% endraw %}
 <br>
+
 Directive arguments can be dynamic. For example, in `v-mydirective:argument=[dataproperty]`, `argument` is the string value assigned to the *arg* property in your directive hook *binding* parameter and `dataproperty` is a reference to a data property on your component instance assigned to the *value* property in the same *binding* parameter. As directive hooks are invoked, the *value* property of the *binding* parameter will dynamically change based on the value of `dataproperty`.
 
 An example of a custom directive using a dynamic argument:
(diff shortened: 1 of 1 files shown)


src/v2/guide/custom-directive.md
```diff
@@ -89,6 +89,8 @@ Directive hooks are passed these arguments:
 - `vnode`: The virtual node produced by Vue's compiler. See the [VNode API](../api/#VNode-Interface) for full details.
 - `oldVnode`: The previous virtual node, only available in the `update` and `componentUpdated` hooks.
 
+<p class="tip">Directive arguments can be dynamic. For example, in `v-my-directive:[foo]`, the arg would be the value of the data property `foo` on your component instance. As directive hooks are invoked, the value of `arg` within the `binding` argument will dynamically change based on the value of `foo`.</p>
```
sdras: This is great! I think it would be more clear to use an example like `v-mydirective:arg="[dataproperty]"`

The two differences there being:
- `v-my-directive` might be more confusing for people just learning
- Using more "real" names and a case that would work can be a little clearer.

It also needs to be taken out of a tip, as there's a tip right next to it, and we try not to do two in a row like that. 

I've made a quick example that can be included as well. If you don't feel like including, I can just follow up with another PR after yours.

https://codepen.io/sdras/pen/62a58fc1a893bc88c7a6795d212b6a8c?editors=1010

Great work! Thanks!

bk-dev: You're welcome. I made some changes based on your comments.

sdras:
Great job, thank you so much 🎉

Issue #1248, opened 2017-11-01
Clarify all use cases for `key` in the guide


Justineo:
Currently when I search for `key` on vuejs.org I'll only get the “List Rendering” section. And this section only tells us `key` is useful when rendering lists, as a hint for element reuse. 

> Since it’s a generic mechanism for Vue to identify nodes, the `key` also has other uses that are not specifically tied to `v-for`, as we will see later in the guide.

I looked into the later part of the guide and found nothing about `key`. In fact it's mentioned earlier in [“Conditional Rendering - Controlling Reusable Elements with `key`”](https://vuejs.org/v2/guide/conditional.html#Controlling-Reusable-Elements-with-key). The problem here is this section only tells us `key` will affect element reuse but fail to mention this may happen to components as well. Later in the same section, in “Conditional Rendering - `v-if` vs `v-show`”, it says child components are destroyed and re-created during toggles, which may make this more confusing:

> v-if is “real” conditional rendering because it ensures that event listeners and child components inside the conditional block are properly destroyed and re-created during toggles.

According to my current understanding, `key` is an essential hint for Vue across the whole system. Components and elements are all possible to be reused during conditional or list rendering. I think it would be better to give `key` a separate section to clarify all possible situations when we need to use `key` because of the mechanisms under the hood, in one place.

myst729:
@Justineo I think you should post a JSFiddle link to the issue we talked of yesterday. IMHO it's a bug.

Generally, I'm against the proposal. How `key` is working in Vue is an implementation detail, which ordinary developers wouldn't care about. It's OK for list rendering, because this use case of reusing is very easy to understand and does make sense. But why a developer should be aware of reusing mechanism when `v-if` and `v-else` are rendering the same component? Especially when the docs have claimed that child components are `properly destroyed and re-created during toggles`.

Justineo:
@myst729 

OK. See https://jsfiddle.net/j7dopc99/.

chrisvfritz:
I wouldn't say how `key` works is an implementation detail, because there are many cases (like with `<keep-alive>`) where I would only know when/why I'd use `key` _if_ I understand how it works. I can see a page on object constancy in general being useful.

sdras:
Does the link to the API reference merged in #2064 aid this understanding enough to close this @Justineo? Or do you think the user needs further explanation inline?

Justineo:
[...]

PR #2064, opened 2019-03-16
Explain key uniqueness, including when generated elements are siblings.

drojf:
The required uniqueness of keys for `v-for` isn't documented (whether it should be globally unique, unique with some particular scope, or just unique within each `v-for`).

On the Vue forum, there is a post where [the answer was it must be "unique within each `v-for`"](https://forum.vuejs.org/t/v-for-key-does-the-key-need-to-be-unique-across-the-entire-component/28040/3), however I have found that if the generated nodes are the same type and are siblings, their keys are treated as in the same scope, and can conflict even if they were generated with different `v-for` statements.

This case could happen if you want to generate a list from two groups of items, where some items (with the same key) appear in both groups, and the lists are adjacent to each other in the DOM.

I have added a subsection to to the List Rendering->`key` documentation which explains this. As I have only started using Vue recently, I hope I haven't misunderstood the problem.

I have explained it by saying "it's unique within each v-for, with this one exception", but if there's some global rule which can be explained just as easily, that would be nicer.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -235,6 +235,8 @@ Since it's a generic mechanism for Vue to identify nodes, the `key` also has oth
 
 <p class="tip">Don't use non-primitive values like objects and arrays as `v-for` keys. Use string or numeric values instead.</p>
 
+For detailed usage of the `key` attribute, please see the [`key` API documentation](https://vuejs.org/v2/api/#key).
+
 ## Array Change Detection
 
 ### Mutation Methods


Justineo:
Thank you for your contribution! But the content of this PR is not accurate enough IMO. As I understand, `key`s are not only required by `v-for` (and are actually not always required by `v-for`). When you have conditionally generated nodes among siblings, you should provide unique `key`s for those stateful ones, ie. input elements, focusable elements or components with local state or even scrollable elements if you want to keep their local state. See also #1248.

drojf:
That's...basically what I would have wanted to know from the documentation - "conditionally generated nodes among siblings should have unique keys".

I feel like keys deserves its own section in the Guide, as it's core to properly using Vue with generated nodes.

I only just found this page which documents it very well: https://vuejs.org/v2/api/#key  ...but it's in the full api documentation, so you are unlikely to find it unless you're specifically looking for it. 

Would it be possible to include this information directly in the Guide (with a bit of tweaking)? or does that sound like a bad idea? I'm happy to try and action it if you guys trust me enough :S.

phanan:
> I only just found this page which documents it very well: https://vuejs.org/v2/api/#key ...but it's in the full api documentation, so you are unlikely to find it unless you're specifically looking for it.

Indeed, and that's why I'd suggest linking to this API section instead of repeating the information here, especially because the existing documentation is more precise and concise (IMO) in explaining the topic under discussion:

> Children of the same common parent must have unique keys. Duplicate keys will cause render errors.

drojf:
I've updated the commit to link to the API documentation as you requested.

sdras:
This looks good now and I agree with @phanan 👍

PR #2065, opened 2019-03-16
Replace deprecated slot syntax in render-function.md

jhnns:
Hey there 👋 

I found one example that was still using the deprecated slot syntax. Hope my correction is correct 😁


phanan:
Thanks!~

PR #2068, opened 2019-03-19
feat: Add Annotating Props to Typescript page

pikax:
Adding some examples on how to annotate props with typescript

### DIFF
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -217,4 +217,4 @@ const Component = Vue.extend({
   }
 })
 ```
-If you find validator not getting type inference or member completion isn't working, annotating the argument the the expected type may help address these problems;
+If you find validator not getting type inference or member completion isn't working, annotating the argument the expected type may help address these problems.
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -217,4 +217,4 @@ const Component = Vue.extend({
   }
 })
 ```
-If you find validator not getting type inference or member completion isn't working, annotating the argument the expected type may help address these problems.
+If you find validator not getting type inference or member completion isn't working, annotating the argument with the expected type may help address these problems.
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -213,15 +213,8 @@ const Component = Vue.extend({
       validator (message: ComplexMessage) {
         return !!message.title;
       }
-    },
-    notificationMessage: {
-      type: Object,
-      validation: (message) => {
-        return !!message.title;
-      } 
-    } as PropValidator<ComplexMessage>
+    }
   }
 })
 ```
-If you find type inference or member completion isn’t working, annotating prop with `PropValidator<T>` may help address it.
-
+If you find validator not getting type inference or member completion isn't working, annotating the argument the the expected type may help address these problems;
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -205,12 +205,12 @@ const Component = Vue.extend({
     name: String,
     success: { type: String },
     callback: { 
-      type: Function as PropType<()=>void>
+      type: Function as any as PropType<() => void>
     },
     message: {
       type: Object as PropType<ComplexMessage>,
       required: true,
-      validation(message: ComplexMessage){
+      validator (message: ComplexMessage) {
         return !!message.title;
       }
     },
--- a/src/v2/guide/typescript.md
+++ b/src/v2/guide/typescript.md
@@ -187,3 +187,43 @@ const Component = Vue.extend({
 ```
 
 If you find type inference or member completion isn't working, annotating certain methods may help address these problems. Using the `--noImplicitAny` option will help find many of these unannotated methods.
+
+
+
+## Annotating Props
+
+```ts
+import Vue, { PropType } from 'vue'
+
+interface ComplexMessage { 
+  title: string,
+  okMessage: string,
+  cancelMessage: string
+}
+const Component = Vue.extend({
+  props: {
+    name: String,
+    success: { type: String },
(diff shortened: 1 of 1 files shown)


src/v2/guide/typescript.md
```diff
@@ -187,3 +187,43 @@ const Component = Vue.extend({
...
+  props: {
+    name: String,
+    success: { type: String },
+    callback: { 
+      type: Function as PropType<()=>void>
+    },
+    message: {
+      type: Object as PropType<ComplexMessage>,
+      required: true,
+      validation(message: ComplexMessage){
```
KaelWD:
```suggestion
      validator (message: ComplexMessage) {
```

KaelWD: You can probably use arrow functions here too. I think the main docs don't use them because of browser support but TS has to be transpiled anyway.

src/v2/guide/typescript.md
```diff
@@ -187,3 +187,43 @@ const Component = Vue.extend({
...
+        return !!message.title;
+      }
+    },
+    notificationMessage: {
+      type: Object as PropType<ComplexMessage>,
+      required: true,
+      // without argument type
+      validation: function(message){
+        return !!message.title;
+      } as (arg: ComplexMessage) => boolean
```
KaelWD: I don't think anyone would ever do this, and it's very similar to the previous example. A better one would be `notificationMessage: {...} as PropValidator<ComplexMessage>` with no annotation on the validator or type.

src/v2/guide/typescript.md
```diff
@@ -187,3 +187,43 @@ const Component = Vue.extend({
...
+  title: string,
+  okMessage: string,
+  cancelMessage: string
+}
+const Component = Vue.extend({
+  props: {
+    name: String,
+    success: { type: String },
+    callback: { 
+      type: Function as PropType<()=>void>
```
KaelWD:
```suggestion
      type: Function as any as PropType<() => void>
```

Not sure if PropValidator would help with this one.

src/v2/guide/typescript.md
KaelWD: See https://github.com/vuejs/vue/issues/8679#issuecomment-414437327

src/v2/guide/typescript.md
pikax: Makes sense

NataliaTepluhina:
The only issue I can see is if you try to do 
```js
  callback: {
    type: Function as PropType<() => void>
  },
```
You will have an error
```
Type 'FunctionConstructor' cannot be converted to type 'PropType<() => void>'.
  Type 'FunctionConstructor' is not comparable to type 'Prop<() => void>[]'.
```

pikax:
That issue is: https://github.com/vuejs/vue/issues/9692

It will be fixed on the next release: 
https://github.com/vuejs/vue/commit/df4af4bd1906b9f23b62816142fdfbd6336d3d2f

KaelWD:
Ah nice. @pikax you can probably remove that `as any` then.

pikax:
I didn't even notice that `as any` thought the suggestion was only for code style.

src/v2/guide/typescript.md
```diff
@@ -187,3 +187,41 @@ const Component = Vue.extend({
...
+      validator (message: ComplexMessage) {
+        return !!message.title;
+      }
+    },
+    notificationMessage: {
+      type: Object,
+      validation: (message) => {
+        return !!message.title;
+      } 
+    } as PropValidator<ComplexMessage>
```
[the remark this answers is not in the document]
pikax: Just wanted to provide other ways to allow forcing the PropValidator, when you don't send the validation parameter type typescript has some problems to get type explicity, it will complain about implicit 'any'.
Other way to fix this would be annotate the validation with `as (message: ComplexMessage)=>boolean` or simply set the type on the parameter.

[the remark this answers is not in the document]
ktsn: Then let's just annotate validator parameter 🙂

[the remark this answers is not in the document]
pikax: there is no need for that example then 🙂

src/v2/guide/typescript.md
```diff
@@ -187,3 +187,34 @@ const Component = Vue.extend({
...
+      type: Object as PropType<ComplexMessage>,
+      required: true,
+      validator (message: ComplexMessage) {
+        return !!message.title;
+      }
+    }
+  }
+})
+```
+If you find validator not getting type inference or member completion isn't working, annotating the argument the the expected type may help address these problems;
```
KaelWD: `the the`

src/v2/guide/typescript.md
```diff
@@ -187,3 +187,34 @@ const Component = Vue.extend({
...
+      type: Object as PropType<ComplexMessage>,
+      required: true,
+      validator (message: ComplexMessage) {
+        return !!message.title;
+      }
+    }
+  }
+})
+```
+If you find validator not getting type inference or member completion isn't working, annotating the argument the expected type may help address these problems.
```
phanan: I think the sentence is broken. What exactly does it mean by "annotating the argument the expected type may help address these problems?"

pikax: If you don't add the argument type to `validator`, typescript won't resolve the correct type for the prop, leading to not have autocompletion, for example this won't work as expected: 
```ts
  message: {
      type: Object as PropType<ComplexMessage>,
      required: true,
      validator (message) {
        return !!message.title;
      }
    } // typescript won't be able to resolve this type correctly
```

I admit the wording is not one of my strengths

phanan: I think what you mean is "annotating the argument _with_ the expected type" then.

jagged3dge:
This ever gonna make it in?

PR #2069, opened 2019-03-19
Fix typo

YourSouLi5Mine:
data was written as tada


YourSouLi5Mine:
My bad: it was an Animate.css animation

PR #2075, opened 2019-03-20
Fix the npm command and add a tip about ports

phanan:
Came across this cookbook article while triaging another issue. Since Vue CLI's version is now officially 3 with some changes (duh), I'm fixing the serve command and adding a tip about ports.

### DIFF
--- a/src/v2/cookbook/debugging-in-vscode.md
+++ b/src/v2/cookbook/debugging-in-vscode.md
@@ -41,6 +41,8 @@ module.exports = {
 
 ### Launching the Application from VS Code
 
+<p class="tip">We're assuming the port to be `8080` here. If it's not the case (for instance, if `8080` has been taken and Vue CLI automatically picks another port for you), just modify the configuration accordingly.</p>
+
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/debugging-in-vscode.md
```diff
@@ -81,7 +83,7 @@ Click on the Debugging icon in the Activity Bar to bring up the Debug view, then
 2.  Open your favorite terminal at the root folder and serve the app using Vue CLI:
 
   ```
-  npm start
+  npm run serve
```
NataliaTepluhina: Just a minor comment: should we consider `yarn serve` as an alternative here?

phanan: As much as I love yarn, I think we're using `npm` across the docs. Also, Vue CLI generated projects are not yarn specific (no `yarn.lock` for example).

NataliaTepluhina: Ok, just ignore the comment above then 😅

PR #2079, opened 2019-03-21
Fixed rollup config

romanslonov:
Current rollup config is not working now. According to this https://github.com/vuejs/rollup-plugin-vue/commit/fd3dfb9a4440b13b220fb2dc535506274280fa0c commit, it requires 'rollup-plugin-commonjs' plugin as well. Without that, we got "Error: 'default' is not exported by node_modules/vue-runtime-helpers/dist/normalize-component.js" error.

### DIFF
--- a/src/v2/cookbook/packaging-sfc-for-npm.md
+++ b/src/v2/cookbook/packaging-sfc-for-npm.md
@@ -112,11 +112,12 @@ There is no need to write your module multiple times. It is possible to prepare 
     "build:unpkg": "rollup --config build/rollup.config.js --format iife --file dist/my-component.min.js"
   },
   "devDependencies": {
-    "rollup": "^0.57.1",
-    "rollup-plugin-buble": "^0.19.2",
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/packaging-sfc-for-npm.md
```diff
@@ -171,13 +172,15 @@ With the package.json `scripts` section ready and the SFC wrapper in place, all
 ```js
 import vue from 'rollup-plugin-vue'; // Handle .vue SFC files
 import buble from 'rollup-plugin-buble'; // Transpile/polyfill with reasonable browser support
+import commonjs from 'rollup-plugin-commonjs';
```
phanan: We need an empty line after this.

phanan (changes requested):

romanslonov:
@phanan ?

znck:
@phanan There is an example configs page on rollup docs. Maybe we can link it here.

PR #2084, opened 2019-03-24
commits from examples

iamalejandroga:
commits.md traducido


phanan:
I'm assuming this is by mistake.

Issue #2067, opened 2019-03-18
Inaccuracy in comparisons guide


MickaMx:
In the paragraph https://vuejs.org/v2/guide/comparison.html#Flexibility-and-Modularity you are using webpack template but it's deprecated thanks to Vue CLI 3. 
 
Maybe talking about Vue CLI 3 is best.

phanan:
Indeed, thanks for pointing out. Are you interested in working on a PR?

MickaMx:
Why not, that could be interesting.

phanan:
@MickaMx Any updates?

MickaMx:
I've been busy the last week, i'll do it today or tomorrow.

sdras:
Fixed! Closing it out.

PR #2088, opened 2019-03-26
update Comparisons-guide Flexibility and Modularity

MickaMx:
Fix for the issue #2067

### DIFF
--- a/src/v2/guide/comparison.md
+++ b/src/v2/guide/comparison.md
@@ -146,7 +146,7 @@ Vue is much simpler than AngularJS, both in terms of API and design. Learning en
 
 AngularJS has strong opinions about how your applications should be structured, while Vue is a more flexible, modular solution. While this makes Vue more adaptable to a wide variety of projects, we also recognize that sometimes it's useful to have some decisions made for you, so that you can just start coding.
 
-That's why we offer a [webpack template](https://github.com/vuejs-templates/webpack) that can set you up within minutes, while also granting you access to advanced features such as hot module reloading, linting, CSS extraction, and much more.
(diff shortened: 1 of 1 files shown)


src/v2/guide/comparison.md
```diff
@@ -146,7 +146,7 @@ Vue is much simpler than AngularJS, both in terms of API and design. Learning en
 
 AngularJS has strong opinions about how your applications should be structured, while Vue is a more flexible, modular solution. While this makes Vue more adaptable to a wide variety of projects, we also recognize that sometimes it's useful to have some decisions made for you, so that you can just start coding.
 
-That's why we offer a [webpack template](https://github.com/vuejs-templates/webpack) that can set you up within minutes, while also granting you access to advanced features such as hot module reloading, linting, CSS extraction, and much more.
+That's why we offer a full system for rapid Vue.js development. [Vue CLI](https://github.com/vuejs/vue-cli) aims to be the standard tooling baseline for the Vue ecosystem. It ensures the various build tools work smoothly together with sensible defaults so you can focus on writing your app instead of spending days wrangling with configurations. At the same time, it still offers the flexibility to tweak the config of each tool without the need for ejecting.
```
phanan: Sounds good to me overall. Some suggestions for possible improvements:

* "_hours_ wrangling" instead of days (we know that it can take days, but let's assume for the better case here)
* "configuration" instead of "config"
* I feel like the part of "without the need for _ejecting_ (emphasized)" can be written in a way so that it doesn't feel too specifically technical. WDYT?

MickaMx: I fixed the first 2 points, for the last time, I don't understand what you mean. I am French and I may have missed a hidden meaning.

phanan: How about "to tweak the configuration of each tool to specific needs?"

MickaMx: That sound great !

phanan:
Thanks a lot!

PR #2092, opened 2019-03-27
Removed this.errors = [];

pebutler3:
Maybe I am missing something, but the this.errors = []; seem to be redundant.


phanan:
Thank you, but the current implementation is correct. `this.error` is reset so that we don't have duplicated error messages.

pebutler3:
ah yes, that makes sense.

PR #2094, opened 2019-03-29
Improvement Suggestion

KadoBOT:
Having UI status as boolean is prone to bugs. In this case, it's correct, but in other cases, the user can end up with `error` and `loading` turned on due to bad code, causing a bug. By having status as a string and not boolean, we can be safe against this type of bug.

### DIFF
--- a/src/v2/cookbook/using-axios-to-consume-apis.md
+++ b/src/v2/cookbook/using-axios-to-consume-apis.md
@@ -109,8 +109,7 @@ new Vue({
   data () {
     return {
       info: null,
-      loading: true,
-      errored: false
+      status: 'loading',
     }
   },
   filters: {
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR. I'm not sure what you mean by "the user can end up with `error` and `loading` turned on due to bad code" though. Do you have an example to share with us?

phanan:
> Sure, let's say that I can't use finally for a reason (i.e. can't transpile it), so I just remove it

I'd say this "problem" applies for any kind of code modification/removal logically. When one alters some code, they should be aware of the effect(s). I don't see this as the code's problem, but the developer's. 

Also, I'd personally frown upon using strings as logical flags, and when it _must_ be the case, I'd always turn the strings into constants.

As such, I'm closing this PR. Thanks a lot for your contribution though, and looking forward to the next ones!

PR #2098, opened 2019-04-02
In relation to issues #9299 and #9777

bbsimonbb:
Make clear that mobile keyboards share the IME behaviour.


phanan:
Thanks, but I don't think this is necessary. Both mentioned mobile keyboards (Swype and Gboard) have already been included in the wiki link, and the information of "v-model is updated after each word, as determined by the entry of a space, or when focus leaves the field" isn't 100% correct.

bbsimonbb:
It's up to you. I got really tricked by this, and was completely persuaded the mobile behaviour of v-model was a bug. Then 7 others piled in behind me and thought it was a bug too. In 18 months of fairly intense Vue use, this is the first time I've misunderstood/been led astray by the incredibly excellent docs.

PR #2099, opened 2019-04-04
Adds a Video Courses link to the Learn drop down

Gregg:
Adding "Video Courses" link under Learn in the nav bar.  This links to the Vue Mastery courses page.  

skitched-20190404-114055

### DIFF
--- a/themes/vue/layout/partials/learn_dropdown.ejs
+++ b/themes/vue/layout/partials/learn_dropdown.ejs
@@ -2,17 +2,20 @@
   <a class="nav-link<%- page.path.match(/\/(guide(?!\/team.html$)|api|style-guide|examples|cookbook)(\/|$)/) ? ' current' : '' %>">Learn</a><span class="arrow"></span>
   <ul class="nav-dropdown">
     <li><h4>Documentation</h4></li>
-    <li><ul>
-      <li><a href="<%- url_for("/v2/guide/") %>" class="nav-link<%- page.path.match(/\/guide(\/|$)/) ? ' current' : '' %>">Guide</a></li>
-      <li><a href="<%- url_for("/v2/api/") %>" class="nav-link<%- page.path.match(/\/api(\/|$)/) ? ' current' : '' %>">API</a></li>
-      <li><a href="<%- url_for("/v2/style-guide/") %>" class="nav-link<%- page.path.match(/\/style-guide(\/|$)/) ? ' current' : '' %>">Style Guide</a></li>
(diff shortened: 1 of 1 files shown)


themes/vue/layout/partials/learn_dropdown.ejs
```diff
@@ -4,6 +4,7 @@
     <li><ul>
       <li><a href="<%- url_for("/v2/guide/") %>" class="nav-link<%- page.path.match(/\/guide(\/|$)/) ? ' current' : '' %>">Guide</a></li>
       <li><a href="<%- url_for("/v2/api/") %>" class="nav-link<%- page.path.match(/\/api(\/|$)/) ? ' current' : '' %>">API</a></li>
+      <li><a href="https://www.vuemastery.com/courses/" class="nav-link" target="_blank">Video Courses</a></li>
```
yyx990803: Let's move it to the bottom so it doesn't disrupt existing 1st party content.

hootlex:
I don’t think `Vue Courses` is a proper title for the link, how about `Vue Mastery`?
Naming the link `courses` makes the content look official for many people and grants authority to vm over other learning platforms like Vue School. This is unfair and can be misleading since there are many great courses out there. What do you think @yyx990803 ?

Gregg:
Sorry for the delayed response, I've been out of town.

Thanks for speaking up @hootlex, the last thing I want to do is anything that discourages teachers from creating Vue courses.  That being said, it's probably important that we know how and why Vue Mastery is getting this link.  Is it a "Sponsored" link?  If so, does a sponsor (with Evan's approval) get to suggest what text goes here?  @yyx990803 what do you think? 

I'm not sure about verbiage.  Adam and I spoke a great deal about what should go here.  I don't think simply putting "Vue Mastery" communicates what is behind the link.  We could write "Vue Mastery Courses" but that's a lot of text compared to the other links on this Nav.  "Video Courses" was the shortest and most understandable we came up with.  I'd like to hear @yyx990803 's thoughts.

themes/vue/layout/partials/learn_dropdown.ejs
Gregg: No problem, done!

yyx990803:
My suggestion: how about a sub section (like the ones in the Ecosystem dropdown) titled "Video Courses", then with VueMastery link and potentially other course links?

Gregg:
I added a subsection (see screenshot below).  I do think it will need to be clear how learning resources get a link added onto this subsection, as I'm sure others will ask.  Since they will make money from this link, Vue.js should get some sort of sponsorship $ up front.  

subsection

hootlex:
@yyx990803 sounds good 👍

sdras:
Looks good now, thanks everyone!

PR #2109, opened 2019-04-08
Fix team distance sorting for no location coordinates

jillztom:
- Add missing coordinates for cities. 
- Fix city names to match `cityCoordsFor` lookup table values
- Pass `null` values for `distanceInKm` in case location is not provided
- Add sort condition to add `null` value vuers to the end of the sorted list.

### DIFF
--- a/src/v2/guide/team.md
+++ b/src/v2/guide/team.md
@@ -225,6 +225,7 @@ order: 803
     'Annecy, France': [45.899247, 6.129384],
     'Alicante, Spain' : [38.346543, -0.483838],
     'Amsterdam, Netherlands': [4.895168, 52.370216],
+    'Atlanta, GA, USA': [33.7676338, -84.5606888],
     'Bangalore, India': [12.971599, 77.594563],
     'Beijing, China': [39.904200, 116.407396],
     'Bordeaux, France': [44.837789, -0.579180],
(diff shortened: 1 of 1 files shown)


src/v2/guide/team.md
```diff
@@ -225,15 +225,18 @@ order: 803
     'Annecy, France': [45.899247, 6.129384],
     'Alicante, Spain' : [38.346543, -0.483838],
     'Amsterdam, Netherlands': [4.895168, 52.370216],
+    'Atlanta, GA, USA': [33.7676338, -84.5606888],
```
phanan: Are these the correct coordinates? It should be `33.749000, -84.388000` based on what I found on Google.

src/v2/guide/team.md
```diff
@@ -225,15 +225,18 @@ order: 803
...
     'Bangalore, India': [12.971599, 77.594563],
     'Beijing, China': [39.904200, 116.407396],
     'Bordeaux, France': [44.837789, -0.579180],
     'Bucharest, Romania': [44.426767, 26.102538],
     'Chengdu, China': [30.572815, 104.066801],
     'Chongqing, China': [29.431586, 106.912251],
     'Denver, CO, USA': [39.739236, -104.990251],
+    'Dublin, Ireland': [53.3239919, -6.5258808],
```
phanan: Same as above: `53.349800, -6.260300`.

src/v2/guide/team.md
```diff
@@ -225,15 +225,18 @@ order: 803
...
     'Dubna, Russia': [56.732020, 37.166897],
     'East Lansing, MI, USA': [42.736979, -84.483865],
+    'Fort Worth, TX, USA': [32.800501, -97.5695075],
```
phanan: And `32.755500, -97.330800`.

src/v2/guide/team.md
```diff
@@ -1379,19 +1382,25 @@ order: 803
...
-              cityCoords[1]
-            )
+            distanceInKm: null
           })
         })
         vuersWithDistances.sort(function (a, b) {
           return (
-            a.distanceInKm -
-            b.distanceInKm
+            a.distanceInKm === null) - (b.distanceInKm === null) ||
```
phanan: Are you sure `(a.distanceInKm === null) - (b.distanceInKm === null)` is the expected statement?

jillztom: Definitely :) I'll update it.

jillztom: Changed the condition here: https://github.com/vuejs/vuejs.org/pull/2109/commits/db5f3156c6e256bd5cc0bee2512c17b6073e2f89

phanan:
LGTM, thanks!

PR #2115, opened 2019-04-11
Add private property names example using module scoping

### LINKED ISSUE #1789: Styleguide: Priority A Rules: Private Property Names
https://vuejs.org/v2/style-guide/#Private-property-names-essential

This rule isn’t completely clear to me, probably because it revolves around the term **custom private properties**, for which no defintion can be found on vuejs.org

So basically it’s straight forward. Applying mixins is simply merging two objects so we need to take care that properties/methods are not overwritten unpredictably. Therefore it advises us to prefix it.

But what exactly is the method ```$_myGreatMixin_update ``` in the example?

a) Is it one to be used by the component consuming the mixin?
So somewhere in the component’s code I'll be writing ```_myGreatMixin_update````

b) Or is ```_myGreatMixin_update``` a computed property which is only used in the mixin itself?

afontcu:
Hi! 👋  This intends to close #1789.

I added an example to show how to leverage modules to make internal methods private, but without removing the one using `$_mixinName`, as I believe is still a recommended strategy. I also reworded the first paragraph accordingly.

As usual, I'm totally open to suggestions, rewording and modifications :)

Thank you very much!

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -564,7 +564,7 @@ Beyond the `scoped` attribute, using unique class names can help ensure that 3rd
 
 ### Private property names <sup data-p="a">essential</sup>
 
-**Always use the `$_` prefix for custom private properties in a plugin, mixin, etc. Then to avoid conflicts with code by other authors, also include a named scope (e.g. `$_yourPluginName_`).**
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
Thank you for this PR, Adri!  Looks good to me, pinging @sdras for the final review

sdras:
This looks great to me and I think it's an improvement. Thank you for submitting it.

afontcu:
Thank you both!! Enjoy Cluj :)

PR #2119, opened 2019-04-12
Clarify multi-word naming with built-in components

### LINKED ISSUE #2114: Multi-word built-in components
Hi! The docs state that:

_"Component names should always be multi-word, except for root App components."_ ([source](https://vuejs.org/v2/style-guide/#Multi-word-component-names-essential)).

However, built-in components such as `<component>` or `<transition>` break that rule. 

Is it something worth mentioning in the docs? My team had some doubts about it (the docs are enforcing a rule that Vue itself isn't following).

However, given the reason to apply this rule (avoid name collision with existing and future HTML elements), one could argue that built-in components should become multi-worded. If that's the case, I'd be more than happy to open an issue to the Vue repo.

Thanks! 🙌

afontcu:
This PR closes #2114.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -43,7 +43,7 @@ Some features of Vue exist to accommodate rare edge cases or smoother migrations
 
 ### Multi-word component names <sup data-p="a">essential</sup>
 
-**Component names should always be multi-word, except for root `App` components.**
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
Thank you for this fix! Looks good to me, merging 👍

PR #2121, opened 2019-04-12
Fix link of Installation

lex111:
Just like in 37 line of same file.

https://github.com/vuejs/vuejs.org/pull/2121/files#diff-35e56c59f0169d94de1eaec69d5d507bR37


phanan:
Thanks!

PR #2122, opened 2019-04-13
Sync url with source in list.md

bartduisters:
The title in list.md is 'v-for with v-if', not 'V-for and v-if'.


phanan:
Thanks a lot!

PR #2123, opened 2019-04-14
Remove smoothScroll from example

bartduisters:
It gives an error in the console when trying to execute the example: example1.items.push({ message: 'Baz' })

This might confuse new people.

Perhaps it is better not to have it in the code at all, since it does nothing because of it being undefined.


phanan:
Agreed. This seems to be some leftover.

PR #2127, opened 2019-04-16
Use npm ci for docker deployments

zzyyxxww:
`npm ci` is much more suited for deploy environments.


phanan:
For `npm ci`, don't we need `package-lock.json` which may not be available in this context?

zzyyxxww:
@phanan that's true, now that I think about I'm not seeing a real benefit for this, since docker caches layers and adding packages doesn't happen on every deploy. Feel free to close this :+1:.

phanan:
👍 Closing. Looking forward to your next PR!

PR #2131, opened 2019-04-20
Fix instructions by replacing `npm` by `yarn`

MickaelBergem:
There is no `package-lock.json` but there is a `yarn.lock`, so I assume `yarn` is preferred?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,9 @@ This site is built with [hexo](http://hexo.io/). Site content is written in Mark
 ## Developing
 
 ``` bash
-$ npm install
-$ npm start # dev server at http://localhost:4000
+# This should also work with npm instead of yarn
+$ yarn install
+$ yarn start # dev server at http://localhost:4000
 ```
 
 ## Deploying


phanan:
Though I personally use (and prefer yarn), npm is still the default package manager for Node. Hence, instead of replacing `npm` commands with `yarn`, maybe we should add the missing `package-lock.json` instead. WDYT?

MickaelBergem:
@phanan I don't have a strong opinion about this. Not blaming/rejecting users who have never used/installed yarn - or who have never heard of it seems important, especially for projects like this one.

Also not sure how other similar projects are handling this issue? https://github.com/vuejs/vue has a `yarn.lock` file.

MickaelBergem:
@sdras I'd be happy to do it, I just have one question: how can we ensure both lockfiles remain in sync over time? 

Should we run the CI pipeline twice, one installing with npm and the other one with yarn (probably `yarn install --frozen-lockfile` to make sure the lockfile is up-to-date, I haven't seen any equivalent)? This option feels really heavy and unconventional :/

sdras:
Closing this out. We're migrating to a new build on VuePress anyway so this isn't a priority for the team.

Issue #2132, opened 2019-04-20
nextTick: Changed the order of compatibility


Mayness:
page: https://vuejs.org/v2/guide/reactivity.html#Async-Update-Queue
code: https://github.com/vuejs/vue/blob/dev/src/core/util/next-tick.js
the compatibility order in nextTick is changed：`Promise.then`、`MutationObserver`、`setImmediate`、`setTimeout`

phanan:
I'm not sure I get what you mean. What exactly is wrong with the current doc?

Mayness:
yeah,the previous compatibility priority were `Promise.then` and `MessageChannel`. But now, the `MutationObserver` replaced `MessageChannel`, and `MutationObserver` is a micro-task.

phanan:
Oh right, sorry I missed that yesterday. Do you want to send a PR over? Something like this will work fine I guess:

>  Internally Vue tries native `Promise.then`, `MutationObserver`, and `setImmediate` for the asynchronous queuing and falls back to setTimeout(fn, 0).

Mayness:
Ok,thank you

NataliaTepluhina:
Closing this issue as #2134 is merged.

PR #2134, opened 2019-04-21
Update reactivity.md

Mayness:
Change the compatibility priority of nextTick
https://github.com/vuejs/vuejs.org/issues/2132


src/v2/guide/reactivity.md
```diff
@@ -77,7 +77,7 @@ There are technical reasons behind this restriction - it eliminates a class of e
...
+In case you haven't noticed yet, Vue performs DOM updates **asynchronously**. Whenever a data change is observed, it will open a queue and buffer all the data changes that happen in the same event loop. If the same watcher is triggered multiple times, it will be pushed into the queue only once. This buffered de-duplication is important in avoiding unnecessary calculations and DOM manipulations. Then, in the next event loop "tick", Vue flushes the queue and performs the actual (already de-duped) work. Internally Vue tries native Promise.then, MutationObserver, and setImmediate for the asynchronous queuing and falls back to setTimeout(fn, 0).
```
Justineo:
```suggestion
In case you haven't noticed yet, Vue performs DOM updates **asynchronously**. Whenever a data change is observed, it will open a queue and buffer all the data changes that happen in the same event loop. If the same watcher is triggered multiple times, it will be pushed into the queue only once. This buffered de-duplication is important in avoiding unnecessary calculations and DOM manipulations. Then, in the next event loop "tick", Vue flushes the queue and performs the actual (already de-duped) work. Internally Vue tries native `Promise.then`, `MutationObserver`, and `setImmediate` for the asynchronous queuing and falls back to `setTimeout(fn, 0)`.
```

NataliaTepluhina: @Mayness, thank you for your contribution!

 In order to merge your pull request could you please apply @Justineo's suggestion or allow commits from maintainers so we can apply it ourselves?

Mayness: Oh,of course I accept Justineo's suggestion.

NataliaTepluhina:
@Mayness thank you for this contribution! :+1:

PR #2135, opened 2019-04-21
Extended watch API documentation

hitautodestruct:
Added another example of how to use watch handler and elaborated on `deep` comment.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -674,14 +674,14 @@ type: api
       },
       // string method name
       b: 'someMethod',
-      // deep watcher
+      // the callback will be called whenever any of the watched object properties change regardless of their nested depth
       c: {
         handler: function (val, oldVal) { /* ... */ },
         deep: true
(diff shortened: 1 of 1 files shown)


phanan:
I think this change makes a lot of sense, thanks!

Justineo:
Why change `d` into using a method name here?

hitautodestruct:
It's an example of using a method name inside of a watcher definition.
It just shows there is an option to do so and how to do it.

Justineo:
Oh, it makes sense. Maybe we can add some explanation on this too? The original comment focused on the `immediate` part.

hitautodestruct:
Two options then:
1) We make another example that includes just the handler as a method name.
2) We add a sub comment to the immediate property.

Since there were two examples of functions as method handlers I simply
changed one. But could have easily added another example.

What do you think?

Justineo:
To me, a sub comment should suffice.

PR #2136, opened 2019-04-22
Added sub comment to method name handler

hitautodestruct:


phanan:
Please fix the merge conflict.

src/v2/api/index.md
```diff
@@ -674,14 +674,15 @@ type: api
...
-      // deep watcher
+      // the callback will be called whenever any of the watched object properties change regardless of their nested depth
       c: {
         handler: function (val, oldVal) { /* ... */ },
         deep: true
       },
       // the callback will be called immediately after the start of the observation
       d: {
-        handler: function (val, oldVal) { /* ... */ },
+        // use a method name, instead of an anonymous function
```
Justineo: This looks a bit misleading that people may assume using method name here is required. Maybe we can use something like:

```suggestion
        // string method name also works here
```

phanan: Or "a method name also works here" because "name" already implies a string.

sdras:
This has no movement for a while and is stuck in merge conflict since April. I'm closing it out, please feel free to do a fresh PR

PR #2137, opened 2019-04-22
Update typescript.md to describe how to use typescript types without module bundler (closes #2042)

### LINKED ISSUE #2042: Improve typescript documentation to show how to use vue without imports that would require module bundler
Since currently type definitions for vue need to be exported, you have to use import:

```typescript
//ItemComponentl.ts
import Vue from 'vue'

const ItemComponent = Vue.extend({
})
```

To get them work without import you can create custom type definition with declare global:

```typescript
//global.d.ts - name does not matter
import { VueConstructor } from 'vue'

declare global {
    const Vue: VueConstructor
}
```

realmerx:


src/v2/guide/typescript.md
```diff
@@ -70,6 +70,18 @@ const Component = {
 }
 ```
 
+To use vue types in environment without module bundler you can create custom type definition file containing declare global (see https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
```
phanan:
```suggestion
To use Vue types in an environment without a module bundler, you can create a custom type definition file containing `declared global` (see [Declaration Merging
](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).
```

src/v2/guide/typescript.md
```diff
@@ -70,7 +70,8 @@ const Component = {
 }
 ```
 
-To use vue types in environment without module bundler you can create custom type definition file containing declare global (see https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
+To use Vue types in an environment without a module bundler, you can create a custom type definition file containing `declared global` (see [Declaration Merging
```
phanan: The markup is broken here.

realmerx: Not sure I understood exactly what is broken in markup? Found a typo and added link to url label as well.

phanan: Preview your markdown and you'll see what I mean.

realmerx: Interesting. I am probably missing something obvious because I did preview and looks correct to me. Here is what I am testing:

https://github.com/vuejs/vuejs.org/blob/6e6891eeb139c6dc712169359677e1bf72dd3c4b/src/v2/guide/typescript.md

And here is what I am seeing:

image

src/v2/guide/typescript.md
```diff
@@ -70,6 +70,18 @@ const Component = {
 }
 ```
 
+To use Vue types in an environment without a module bundler, you can create a custom type definition file containing `declare global` (see [Declaration Merging https://www.typescriptlang.org/docs/handbook/declaration-merging.html](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).
```
ktsn: It would make more sense to explicitly describe what the definition file does.

> To use Vue types

I feel we cannot use Vue's types (like `ComponentOptions` etc.) without this definition which is not correct from this phrase.

What about the below change?

```suggestion
To use Vue in an environment without a module bundler, you can create a custom type definition file containing `declare global` to declare global `Vue` object type (see [Declaration Merging https://www.typescriptlang.org/docs/handbook/declaration-merging.html](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).
```

realmerx: Thank you all for reviewing. Yes, wording needs some work to make it clear what is the purpose of this. You can use Vue fine in typescript without importing types or creating custom type definition file with declare global if you can live with errors (javascript still gets generated correctly) and without intellisense and type recognition in IDE.

realmerx:
```suggestion
To compile Vue without errors and have intellisense and type recognition in IDE in environment without a module bundler you can create a custom type definition file containing `declare global` to declare global `Vue` object type (see [Declaration Merging https://www.typescriptlang.org/docs/handbook/declaration-merging.html](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)).
```

ktsn: Wait, is this code really work on your environment? I just try it in my local but there is a compile error on it. I'm using the latest Vue (v2.6.10).

```
Subsequent variable declarations must have the same type.  Variable 'Vue' must be of type 'typeof import("(...)/node_modules/vue/types/index")', but here has type 'VueConstructor<Vue>'.ts(2403)
```

ktsn: OK, I realized there is already this kind of type declaration in Vue's typings. https://github.com/vuejs/vue/blob/dev/types/index.d.ts#L5

But this declaration is actually not correct as `Vue` object type is declared under `Vue.default`. This causes the compilation error which I showed above.

I already created the PR to fix this bug. Once this is merged, you can use global `Vue` without any additional set up which means we would no longer require this section in docs.
https://github.com/vuejs/vue/pull/9912

realmerx: Yes I think something has changed in latest typescript version that makes this kind of declaration break - I need to investigate more. But if we can make it work in global context without tricks then even better.

phanan:
Closing in favor of vuejs/vue#9912. Thanks guys!

davidfstr:
As far as I can tell the PR that replaced this one ( vuejs/vue#9912 ) did not actually update any documentation that explains how to use Vue in a global context, which is what this PR was originally intended to do. Therefore I'd advocate reopening this PR.

I'm actively having trouble trying to typecheck a JS file with TypeScript that uses Vue through the `window.Vue` global, as covered in [my forum post here](https://forum.vuejs.org/t/using-typescript-to-typecheck-js-code-that-uses-the-vue-global/68237), so having updated documentation would be quite useful!

PR #2138, opened 2019-04-24
npm run serve is dependent on package.json

cptully:
Following this howto page assumes that you have a serve script in your package.json file.

There is no such script in the project I have inherited so this howto fails miserably....


NataliaTepluhina:
@cptully as stated in the mentioned file: `This recipe shows how to debug Vue CLI applications in VS Code as they run in the browser.`

Vue CLI 3 application has `serve` npm task by default.

Thank you for the PR anyway!

PR #2139, opened 2019-04-24
Make it clear that cookbook isn't meant to be a step-by-step guide

phanan:
As some of our users followed the cookbook tutorials literally step-by-step and were surprised when things like `npm` commands didn't work out of the box, we feel the need of adding this note.


sdras:
The copy is ok but I'm not super down with making this a tip, it should be incorporated into the flow of the document. I'll follow up with another PR.

phanan:
Oops.

Sarah Drasner <notifications@github.com> schrieb am Mi. 24. Apr. 2019 um
16:30:

PR #2140, opened 2019-04-24
docs: add a tip on using appear attribute and v-on:appear hook

DaKoala:
Based on [issue #9320](https://github.com/vuejs/vue/issues/9320) in the issue list of vue, I added a tip to clarify the usage of `appear` attribute and `v-on:appear` hook.

### DIFF
--- a/src/v2/guide/transitions.md
+++ b/src/v2/guide/transitions.md
@@ -569,6 +569,8 @@ By default, this will use the transitions specified for entering and leaving. If
 
 and custom JavaScript hooks:
 
+<p class="tip">Both `appear` attribute and `v-on:appear` hook will cause an appear transition. Compared to `appear` attribute, `v-on:appear` hook has a higher priority. If `v-on:appear` hook is specified, using `:appear="false"` does not cancel the transition.</p>
+
 ``` html
 <transition
   appear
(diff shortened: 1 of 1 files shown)


src/v2/guide/transitions.md
```diff
@@ -569,6 +569,8 @@ By default, this will use the transitions specified for entering and leaving. If
 
 and custom JavaScript hooks:
 
+<p class="tip">Both `appear` attribute and `v-on:appear` hook will cause an appear transition. Compared to `appear` attribute, `v-on:appear` hook has a higher priority. If `v-on:appear` hook is specified, using `:appear="false"` does not cancel the transition.</p>
```
phanan: This `p.tip` here breaks the flow though. Do you think you can rewrite this content in a way that doesn't require a `p.tip`?

DaKoala: Do you think it is okay to keep `p.tip` but move this part below the code snippet?

sdras: Please do heed An's advice here and rewrite it into the flow of the document. Thanks!

DaKoala: I have changed the layout and removed the `p.tip`

sdras:
Looks good now, thanks!

PR #2144, opened 2019-04-28
Fix typo

faliure:


NataliaTepluhina:
Thank you for spotting and fixing it! :+1:

PR #2149, opened 2019-05-03
Updated documentation for custom events.

andrewckor:
I updated the custom events section, to using a positive example, instead of a negative/exception. To make it easier for people to copy paste the example, instead of having to read the whole section in order to understand that the shown example is not recommended.

### DIFF
--- a/src/v2/guide/components-custom-events.md
+++ b/src/v2/guide/components-custom-events.md
@@ -8,20 +8,26 @@ order: 103
 
 ## Event Names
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks, but I find the current version to be better and more fluent.

andrewckor:
I do not agree, but it's fine. 

At least we can include the DO and DO NOT, because this is the only page that is against the whole documentation, that is using a recommended example and not an exception.

PR #2152, opened 2019-05-08
Update list.md

Vbahole:
https://jsfiddle.net/Vbahole22/erco78a5/10/


Vbahole:
something is not right about that example. maybe the first slot on the v-for is the name and not the value?

Vbahole:
had a bad vue version in fiddle i guess

PR #1666, opened 2018-06-03
Improve text accessibility

robcresswell:
This patch makes a few changes:

- Alter the "light" text colour so that it is WCAG AAA compliant for legibility
- Adjust code colours to be at least WCAG AA compliant.
- Darken the "medium" and "dark" colours, so as to preserve significance of
  different shades
- Make text use the font-size set by the browser, rather than hardcoding 15px.
  Most browsers default to 16px, in my understanding. Also adjust the max-width
  of the body, to account for the increased font-size and prevent aggressive
  line wrapping.
- Add VIM swap file extension to .gitignore.

PR #2153, opened 2019-05-08
fix: minor style fixes

robcresswell:
This is a followup to https://github.com/vuejs/vuejs.org/pull/1666/commits/0bfc43c970f02771084c24897916a8ca1f66ad5b

I noticed a couple of small things:
- The lilac color used in code blocks was overlooked, and failed contrast testing, so I've updated that
- The "demo" blocks looked overiszed sized after the font-size change. I'm not completely sure what the intention is there, but I removed the line-height and font size adjustments, and kept the "card style" outline

**Before**

Code color:
image

Demo block:
image

**After**

Code color:
image

Demo block:
image

### DIFF
--- a/themes/vue/source/css/_demo.styl
+++ b/themes/vue/source/css/_demo.styl
@@ -4,8 +4,6 @@
   padding: 25px 35px
   margin-top: 1em
   margin-bottom: 40px
-  font-size: 1.2em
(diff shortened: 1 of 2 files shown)


sdras:
Ok, just had a chance to take a look and wow, that new purple is much better. Thanks!

Issue #2155, opened 2019-05-09
Example of dynamic argument in custom-directive.md is incorrect


oohira:
In #2062, an example of dynamic argument for custom directive was added. However, the intension is not clear because 

1. if it means [dynamic arguments introduced in v2.6](https://vuejs.org/v2/guide/syntax.html#Dynamic-Arguments), the example should be `mydirective:[argument]=dataproperty` instead of `mydirective:argument=[dataproperty]`
2. if it means just passing the argument dynamically, the example should be `mydirective:argument=dataproperty` instead of `mydirective:argument=[dataproperty]`. But this is already described in the previous example `v-demo:foo.a.b="message"`

I think the following example may be close to the original intention.

* https://jsfiddle.net/y0ftjLxs/

cc: @bk-dev @sdras

phanan:
Pinging @sdras

bk-dev:
If I'm totally honest, I was confused as well. Mainly because in my code base the syntax was different than what was outlined by @sdras. An example usage in my code base is `<div v-my-directive:[dynamicArg]></div>`. That said, I have never had a need to use `=dataproperty` to create statement.

My original goal was to pass a data property to a directive and when that data property changed the directive update hook would get invoked so the directive could react to the change. Doing something like 2) from @oohira's post would not have worked (at least not in my testing).

sdras:
Yep, this was my fault for not reading through more thoroughly. I rewrote this section with clearer examples (thank you for the fiddle @oohira), but also rewrote the explanation to be more conversational in our docs style, and explain the problem space/use case. Let me know if that clears things up for you. Thanks for the issue!

PR #2161, opened 2019-05-11
Dynamic Arguments in Custom Directives fixes

### LINKED ISSUE #2160: Issue with Dynamic Arguments
Hi All,

I tried the examples in https://vuejs.org/v2/guide/syntax.html#Dynamic-Arguments but I was not successful as I always get the below warnings in the console.

_[Vue warn]: Property or method "attribname" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property._

My HTML:
```
<!-- Dynamic Arguments-->
<h4 class="comment">Dynamic Arguments</h4>
<a v-bind:[attribName]="urlNamecheap">Visit Namecheap.com</a>
<button v-on:[eventName]="doSomething">Click me</button>
```

My Script:
```
<script>
      var vm = new Vue({
        el: '#app',
        data: {
          msg: 'Hello Vue!',
          rawHtml: '<span style="color:  red">This should be red.</span>',
          dynamicId: 'mainDiv',
          isButtonDisabled: true,
          number: 1000,
          ok: true,
          message: 'I love my Razan',
          id: 'Rzn248',
          seen: true,
          url: 'https://www.google.com',
          urlNamecheap: 'https://www.namecheap.com',
          attribName: 'href',
          eventName: 'click'

        },
        methods: {
          doSomething: function() {
            console.log('I just got fired up, who was it?');
          }
        }
      });
    </script>
```

Am I doing something wrong or is there indeed an issue?

sdras:
Fixes #2160 and #2155

The previous example was inaccurate, and the explanation was a bit jargony. In this PR, I put in a more conversational explanation of the use case, and added in a few examples with results.

### DIFF
--- a/src/v2/guide/custom-directive.md
+++ b/src/v2/guide/custom-directive.md
@@ -159,14 +159,14 @@ Let's say you want to make a custom directive that allows you to pin elements to
 ```js
 Vue.directive('pin', {
   bind: function (el, binding, vnode) {
-    el.style.position = 'fixed';
-    el.style.top = binding.value + 'px';
+    el.style.position = 'fixed'
+    el.style.top = binding.value + 'px'
   }
-}); 
+})
 
 new Vue({
   el: '#baseexample'
-});
+})
 ```
 
 This would pin the element 200px from the top of the page. But what happens if we run into a scenario when we need to pin the element from the left, instead of the top? Here's where a dynamic argument that can be updated per component instance comes in very handy:
@@ -182,11 +182,11 @@ This would pin the element 200px from the top of the page. But what happens if w
 ```js
 Vue.directive('pin', {
   bind: function (el, binding, vnode) {
-    el.style.position = 'fixed';
-    var s = (binding.arg == 'left' ? 'left' : 'top');
-    el.style[s] = binding.value + 'px';
+    el.style.position = 'fixed'
+    var s = (binding.arg == 'left' ? 'left' : 'top')
+    el.style[s] = binding.value + 'px'
   }
-});
+})
 
 new Vue({
   el: '#dynamicexample',
--- a/src/v2/guide/custom-directive.md
+++ b/src/v2/guide/custom-directive.md
@@ -147,7 +147,7 @@ new Vue({
 
 Directive arguments can be dynamic. For example, in `v-mydirective:[argument]="value"`, we can update the argument in a custom directive based on data properties in our component instance! This makes our custom directives extremely flexible for use throughout our application.
 
-Let's say you want to make a custom directive that allows you to easily tack elements to your page in fixed positioning. We could pass a value in that would allow it to be tacked 200px from the top of the page like so:
+Let's say you want to make a custom directive that allows you to easily tack elements to your page using fixed positioning. We could create a custom directive so that we can pass in a value that would update its vertical positioning in pixels, like this:
 
 ```html
 <div id="baseexample">
@@ -169,7 +169,7 @@ new Vue({
 });
 ```
 
-This would tack the element with fixed positioning 200px down the page. But what happens if we run into a scenario when we need to tack the element from the left, instead of the top? Here's where a dynamic directive that you could update per component instance would come in very handy:
+This would tack the element with fixed positioning 200px from the top of the page. But what happens if we run into a scenario when we need to tack the element from the left, instead of the top? Here's where a dynamic argument that you could update per component instance would come in very handy:
 
 
 ```html
--- a/src/v2/guide/custom-directive.md
+++ b/src/v2/guide/custom-directive.md
@@ -164,7 +164,6 @@ Vue.directive('tack', {
   }
 });
 
-// start app
 new Vue({
   el: '#dynamicexample',
   data() {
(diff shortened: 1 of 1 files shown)


src/v2/guide/custom-directive.md
```diff
@@ -143,14 +143,38 @@ new Vue({
 </script>
 {% endraw %}
 
-Directive arguments can be dynamic. For example, in `v-mydirective:argument=[dataproperty]`, `argument` is the string value assigned to the *arg* property in your directive hook *binding* parameter and `dataproperty` is a reference to a data property on your component instance assigned to the *value* property in the same *binding* parameter. As directive hooks are invoked, the *value* property of the *binding* parameter will dynamically change based on the value of `dataproperty`.
+### Dynamic Directive Arguments
 
-An example of a custom directive using a dynamic argument:
+Directive arguments can be dynamic. For example, in `v-mydirective:[argument]="value"`, the `argument`  can be updated based on data properties in our component instance! This makes our custom directives  flexible for use throughout our application.
+
+Let's say you want to make a custom directive that allows you to tack elements to your page using fixed positioning. We could create a custom directive where the value updates the vertical positioning in pixels, like this:
```
phanan: I'll be honest here: I've never used/seen the word "tack" before, and (therefore?) find it weird in a UI context. I even did some googling and…

image

Is it possible to replace the word with something more familiar, like "stick," which is used right in the following example?

sdras: It's not a common a UI element, the tack name is meant more as "tacking something to a board". Or "tacking a note". Would it be more clear if I used the term "pin"? I believe that waypoints library uses "pin" as a term for a UI behavior.

phanan: Yeah, "pin" sounds perfect to me.

sdras: Great! Updated. :)

src/v2/guide/custom-directive.md
```diff
@@ -143,37 +143,71 @@ new Vue({
 </script>
 {% endraw %}
 
-Directive arguments can be dynamic. For example, in `v-mydirective:argument=[dataproperty]`, `argument` is the string value assigned to the *arg* property in your directive hook *binding* parameter and `dataproperty` is a reference to a data property on your component instance assigned to the *value* property in the same *binding* parameter. As directive hooks are invoked, the *value* property of the *binding* parameter will dynamically change based on the value of `dataproperty`.
+### Dynamic Directive Arguments
 
-An example of a custom directive using a dynamic argument:
+Directive arguments can be dynamic. For example, in `v-mydirective:[argument]="value"`, the `argument`  can be updated based on data properties in our component instance! This makes our custom directives flexible for use throughout our application.
```
Justineo:
```suggestion
Directive arguments can be dynamic. For example, in `v-mydirective:[argument]="value"`, the `argument` can be updated based on data properties in our component instance! This makes our custom directives flexible for use throughout our application.
```

sdras: Thanks for catching that :)

src/v2/guide/custom-directive.md
```diff
@@ -143,37 +143,71 @@ new Vue({
...
+<div id="baseexample">
   <p>Scroll down the page</p>
-  <p v-tack:left="[dynamicleft]">I’ll now be offset from the left instead of the top</p>
+  <p v-pin="200">Stick me 200px from the top of the page</p>
+</div>
+```
+
+```js
+Vue.directive('pin', {
+  bind(el, binding, vnode) {
```
phanan: Sorry, missed this in my previous review. Shouldn’t we use ES5 syntax per our doc convention?

sdras: Ah, yep. I think I left a const in there as well 😅

src/v2/guide/custom-directive.md
```diff
@@ -143,37 +143,71 @@ new Vue({
...
     el.style[s] = binding.value + 'px';
   }
-})
+});
 
-// start app
 new Vue({
-  el: '#app',
+  el: '#dynamicexample',
   data() {
```
phanan: This as well? :)

sdras: Done!

src/v2/guide/custom-directive.md
```diff
@@ -143,37 +143,71 @@ new Vue({
...
+  <h3>Scroll down inside this section ↓</h3>
+  <p v-pin:[position]="200">I am pinned onto the page at 200px to the left.</p>
+</div>
+```
+
+```js
+Vue.directive('pin', {
+  bind: function (el, binding, vnode) {
+    el.style.position = 'fixed';
+    var s = (binding.arg == 'left' ? 'left' : 'top');
```
NataliaTepluhina:
```suggestion
    var s = (binding.arg === 'left' ? 'left' : 'top');
```
@sdras what do you think about changing abstract equality to strict equality here?

sdras: Yep!

src/v2/guide/custom-directive.md
```diff
@@ -143,37 +143,71 @@ new Vue({
...
+<div id="dynamicexample">
+  <h3>Scroll down inside this section ↓</h3>
+  <p v-pin:[position]="200">I am pinned onto the page at 200px to the left.</p>
+</div>
+```
+
+```js
+Vue.directive('pin', {
+  bind: function (el, binding, vnode) {
+    el.style.position = 'fixed';
```
phanan: Two last things (I swear 😅):

- I'm afraid a new user may get confused at first by `position` as the dynamic argument's name, since it has a different meaning in CSS (as shown by `el.style.position = 'fixed'`). I couldn't find a better alternative though, so maybe this isn't a big deal after all.
- Did you run the code through a linter which added all these extra semicolons 😄 ? Same with the snippet above.

Justineo: Maybe `side`?

sdras: - How about `direction`
- I'm a no-semi person, so all the semis are too much for me- we really need to get prettier in here because this is the stuff that doesn't really matter :) I'll run it just the snippet through prettier and change my settings and adjust back to the spaces after parens like we have in the docs. We should make this more automated with a `.prettierrc` in the future.

phanan: * `direction` sounds good to me
* Totally agreed re. `.prettierrc`, and I asked the question exactly because normally we don't have semi-colons in Vue docs.

sdras: Oh man, I don't know why I thought we did... working in too many repos at once. I'll remove them, that's easier anyways. Nice catch! Thanks An :)

sdras: Alright, both should be fixed

PR #2163, opened 2019-05-13
Added meta keywords for Better SEO

shivaylamba:


phanan:
Thanks, but meta keywords aren't considered by search engines for SEO anymore.

PR #2167, opened 2019-05-17
Updated computed.md to make the examples accessible

superuntitled:
I changed the paragraph tags to labels. Wrapping the input in a <label> rather than a <p> element promotes semantic markup that would pass an accessibility audit.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -189,10 +189,10 @@ For example:
 
 ``` html
 <div id="watch-example">
-  <p>
+  <label>
     Ask a yes/no question:
     <input v-model="question">
-  </p>
+  </label>
   <p>{{ answer }}</p>
 </div>
 ```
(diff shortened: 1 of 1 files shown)


phanan:
Closing for the same/similar reason as #2168.

superuntitled:
@phanan The reasons you have given for rejecting this update are brevity and that semantic html does not matter here. 

I agree that brevity is important, and I do appreciate the terseness of the code examples in the Vue Guide. The update I have recommended here changes the `<p>` tags to `<label>` tags. I would argue that the update has no effect on brevity.

The idea that semantic html does not matter is absurd. To a web development community, I think that incorrect html markup in it's documentation should be embarrassing. 

```
<p>
  Ask a yes/no question:
  <input v-model="question">
</p>
```
to
```
<label>
  Ask a yes/no question:
  <input v-model="question">
</label>
```

This update not makes the code semantically correct, but also accessible. 

Non-semantic/non-accessible code in documentation gives the impression that the Vue community is not concerned with these things.

phanan:
The problem is though, if the code is changed here, it should be changed
everywhere else (e.g. the other PR) as well or we’ll end up with
inconsistencies, and that’s why I mentioned the same reason. Also, simply
replacing <p> with <label> has the potential of breaking our examples’
layout, knowing that <p> is a block element and <label> is an inline,
unless we add an additional stylesheet to fix it, which I don’t think is
worth the effort.

superuntitled <notifications@github.com> schrieb am Fr. 17. Mai 2019 um
18:39:

superuntitled:
Thanks for the clarification. I would venture to say that there is not consistency in the examples across the guide, but I understand your concerns. Thanks for you reply.

PR #2168, opened 2019-05-17
Updates forms.md to make examples consistent/semantic

superuntitled:
Added labels to form elements and changed some <span> to <p> (to get rid of <br>). Changed the order of some of the examples so that the form inputs are consistently first and the reactive elements are second.

The checkbox with the dynamic true/false label is a semantic catch 22. I changed it a bit. It is not quite as concise, but I think it still gets the point across.

### DIFF
--- a/src/v2/guide/forms.md
+++ b/src/v2/guide/forms.md
@@ -20,13 +20,13 @@ You can use the `v-model` directive to create two-way data bindings on form inpu
 ### Text
 
 ``` html
-<input v-model="message" placeholder="edit me">
+<label>Add a message: <input v-model="message" placeholder="edit me"></label>
 <p>Message is: {{ message }}</p>
 ```
 
 {% raw %}
 <div id="example-1" class="demo">
-  <input v-model="message" placeholder="edit me">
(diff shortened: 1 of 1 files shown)


phanan:
Thanks, but I'd prefer the brevity of the current examples. Semantic HTML is definitely a good thing, just not our main concern here.

superuntitled:
Thanks for looking at this @phanan! I agree that brevity is certainly important in online instruction.

When reading through the Vue Guide for the first time, I was really impressed. It is clear and concise and well organized. However, I was confused at some of the code in the examples: why not use semantically correct markup when it makes the examples clearer and in some cases more succinct.

For example:
```
<br>	
<span>Checked names: {{ checkedNames }}</span>
```
to
`<p>Checked names: {{ checkedNames }}</p>`

This change not only makes the example more semantically correct, but it also eliminates the need for the `<br>` tag. 

I believe that clarity is important when designing instructional examples. I think that the updates I have recommended here have made the concepts easier to understand, and in some cases, made the code less verbose.

I am really very new to vue community, but have been developing front-end applications for an online university for 10 years. This is my first commit to an open source project, and I do think that these updates would be valuable to people who are new to Vue.

PR #2169, opened 2019-05-17
Added example code for each lifecycle hook

superuntitled:
Adding an example of how to code each hook in the diagram. The diagram is awesome, but I have always wished the code for the hooks was available here as well.


phanan:
Thanks for the PR, but I don't think the change is necessary. All lifecycle hooks have been documented extensively on [the API page](https://vuejs.org/v2/api/#Options-Lifecycle-Hooks).

superuntitled:
@phanan When first reading through the Vue Guide for the Lifecycle hooks, I really appreciated the Lifecycle diagram, it gave me a clear visual image of the Lifecycle hooks. While I did find the diagram very helpful, I had hoped supplemental code to reinforce the diagram. When looking for example code around the web, I noticed comments from other developers lamenting the lack of example code for these hooks on the Vue site. 

I would have benefitted from code similar to what I have added here, and I think that others that are new to Vue would as well.

PR #2172, opened 2019-05-17
Update cookbook npm recipe with more up-to-date SSR/package.json info

mgdodge:
The SSR suggestions included in the original are out of date (and a little inaccurate, TBH.) A little over a year later, nuxt.js and quasar have enough of a following that the suggested changes align with their usage as a default. More specifically:

- package.json `main` references the SSR build, created by official vue-template-compiler with `optimizeSSR` flag
- package.json `browser` is actively ignored by nuxt, so is not recommended as the way to get raw `.vue` file
- dropping `umd` in favor of `cjs` since that is the format expected for SSR
  - servers could continue using `umd`, but the SSR optimizations are not compatible with browser usage, which is implied by `umd`

### DIFF
--- a/src/v2/cookbook/packaging-sfc-for-npm.md
+++ b/src/v2/cookbook/packaging-sfc-for-npm.md
@@ -61,35 +61,38 @@ dist/
 
 ### How does npm know which version to serve to a browser/build process?
 
-The package.json file used by npm really only requires one version (`main`), but as it turns out, we aren't limited to that. We can address the most common use cases by specifying 2 additional versions (`module` and `unpkg`), and provide access to the `.vue` file itself using the `browser` field. A sample package.json would look like this:
+The package.json file used by npm really only requires one version (`main`), but as it turns out, we aren't limited to that. We can address the most common use cases by specifying 2 additional versions (`module` and `unpkg`). The `.vue` file can also be included in the package for those who need access to it. A sample package.json would look like this:
 
 ```json
 {
   "name": "my-component",
   "version": "1.2.3",
-  "main": "dist/my-component.umd.js",
+  "main": "dist/my-component.ssr.js",
   "module": "dist/my-component.esm.js",
   "unpkg": "dist/my-component.min.js",
-  "browser": {
-    "./sfc": "src/my-component.vue"
-  },
+  "files": [
+    "dist/*",
+    "src/**/*.vue"
+  ],
   ...
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mgdodge:
*bump*

Looks like I neglected to mention that the motivation for the above update was someone trying to follow the existing recipe (via the utility mentioned in the _acknowledgements_ section) and was unable to get SSR working. The result was a bug report, some addl. research, and ultimately this PR. I've also seen a few mentions in the vue forums about similar issues.

mgdodge:
@sdras Are you still in charge of the cookbook submissions? Would love a review of this, I hate to have inaccurate information out there.

mgdodge:
Any action on this? I still receive notifications when people read comments on blog posts which link to this cookbook recipe, so the outdated and inaccurate information contained is still being read.

mgdodge:
A couple more issues recently that could benefit from this documentation being updated. The new, more accurate information would have possibly prevented those issues from being filed. What can be done to get this reviewed/merged?

mgdodge:
For clarification, the `browser` property is only actively ignored in server mode, but is still used if present on client mode. Since the portion of this recipe that suggests the `browser` property is specifically tied to server use, it is wrong and will never be used in nuxt.

[here](https://github.com/nuxt/nuxt.js/issues/6361) is an issue on nuxt related to this fact. The important bit is where it links to [this line](https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/config/server.js#L48) which shows nuxt does not even look for `browser` in server mode.

adamdehaven:
For what it's worth, I just tried the existing documentation and, as @mgdodge suggested previously, ended up going down a rabbit hole that inevitably lead me to this PR.

I'd like to formally bump this PR to be merged for all the reasons already listed above, and in addition, the simple fact that the current documentation references options that are no longer valid.

As a side note, any idea what has to be done to get `stylus` blocks to be processed correctly by rollup? I keep getting an "unknown word" error that only disappears when I remove the style block from my `.vue` component file.

bencodezen:
Hey @mgdodge! Sorry for the delay in response. Love the changes you made and would love to get this merged. Can we get the last merge conflict resolved? After that, happy to help merge this.

adamdehaven:
@mgdodge one other change that may be needed:

The current npm package for `rollup-plugin-vue` is serving up the v6 beta (linked to Vue version 3.x) by default, which causes errors if you're using Vue 2.x (like I am). 

The documentation on this page should (maybe) reflect that users should explicitly install version 5.x of `rollup-plugin-vue` if they are using Vue 2.x, otherwise, go with the default if using Vue 3.x.

mgdodge:
Fixed the merge conflict. Also, since we're so close to Vue 3, I reworked a few of the examples to indicate considerations when compiling for either version of Vue. I realize that might require some further review, but it took so long to get things this far, the ecosystem changed on me...again...

alessiomaffeis (approved):

PR #2175, opened 2019-05-28
Added dynamic slot name for shortand syntax example

rigwild:
Hello,
I think it would be useful to be mentioned that the [Dynamic Slot Name](https://vuejs.org/v2/guide/components-slots.html#Dynamic-Slot-Names) feature can be used with the [Name slot shortand](https://vuejs.org/v2/guide/components-slots.html#Named-Slots-Shorthand) syntax.

I also moved the `Dynamic Slot Names` chapter under it, for it to be consistent with my change.
Have a nice day !

### DIFF
--- a/src/v2/guide/components-slots.md
+++ b/src/v2/guide/components-slots.md
@@ -345,20 +345,6 @@ You can even define fallbacks, to be used in case a slot prop is undefined:
 </current-user>
 ```
 
-## Dynamic Slot Names
-
-> New in 2.6.0+
-
-[Dynamic directive arguments](syntax.html#Dynamic-Arguments) also work on `v-slot`, allowing the definition of dynamic slot names:
-
-``` html
-<base-layout>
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR. Personally, I find this addition to be unnecessary (slots and dynamic arguments are for immediate or advanced users who shouldn't have a problem concluding this information from other details), but I'll leave the PR open in case others find it better to be explicit this way. 

Pinging @NataliaTepluhina, @sdras, and @chrisvfritz.

rigwild:
It could be under there too. https://vuejs.org/v2/guide/syntax.html#v-on-Shorthand

NataliaTepluhina:
I'd rather agree with @phanan 
Thank you for your work on this PR anyway!

rigwild:
This issue has not received an update in a while, should it be closed or thinked again ?

PR #2179, opened 2019-06-01
added ES6 syntax to improve readability

karimdaghari:


phanan:
Thanks, but we intentionally use ES5 throughout of the doc (with cookbook as the sole exception), as Vue officially supports all ES5 browsers.

PR #2180, opened 2019-06-03
Start with a simple $listeners example

Meekohi:
It took me a few passes through this documentation to understand the simple/common use case for `$listeners`, since it rolls right into the more complicated use for support `v-model`. Starting with a simple explicit example will hopefully make it easier for new developers to catch this use case.


NataliaTepluhina:
@Meekohi thank you for working on this!
@chrisvfritz could you please give your review on this PR?

PR #2186, opened 2019-06-07
Add details about the Filters.

Koooooo-7:
Add details about the  Filters when the global filter has the same name as the local filter.

when the global filter has the same name as the local filter, 
the local filter will be preferred ,based on Proximity Principle.


NataliaTepluhina:
Thank you for working on this PR! :+1:

PR #2187, opened 2019-06-08
Ignore editor's directories in .gitignore file

negarjf:
This pull request will add `idea` and `vscode` directories to `.gitignore` file.


phanan:
Thanks, but these entries belong in the global `.gitignore` file instead.

PR #2196, opened 2019-06-12
Add Truthiness/Falsiness with v-bind

starkovsky:


NataliaTepluhina:
@starkovsky thank you for your work! This section definitely should have beed added to docs, not only to Migration guide 😅 

My only concern is maybe we should place this addition either to `v-bind` description or [here](https://vuejs.org/v2/guide/components-props.html#Passing-a-Boolean). It seems more logical to me than `Conditional rendering`.

@phanan what do you think?

src/v2/guide/conditional.md
```diff
@@ -192,10 +192,8 @@ Generally speaking, `v-if` has higher toggle costs while `v-show` has higher ini
 When used together with `v-if`, `v-for` has a higher priority than `v-if`. See the <a href="../guide/list.html#v-for-with-v-if">list rendering guide</a> for details.
 
 
-### Truthiness/Falsiness with `v-bind`
+### Truthiness/Falsiness with `v-show`
```
iNerV: `Note that for other directives (e.g. `v-if` and `v-show`), JavaScript's normal truthiness still applies.`
previously version was correct

starkovsky:
I agree, in directives more correctly

src/v2/api/index.md
```diff
@@ -2197,6 +2197,11 @@ type: api
   ```
 
   `.camel` is not needed if you are using string templates, or compiling with `vue-loader`/`vueify`.
+  
+  ## Truthiness/Falsiness with `v-bind`
```
```suggestion
  **Note on truthiness and falsiness
```

phanan: A level 2 heading here will break the structure of the document/section. I'd suggest using the same style as other list items (Example, See also and so on).

starkovsky: Yes, you right

src/v2/api/index.md
```diff
@@ -2197,6 +2197,11 @@ type: api
   ```
 
   `.camel` is not needed if you are using string templates, or compiling with `vue-loader`/`vueify`.
+  
+- **Note on truthiness and falsiness**
+  When used with `v-bind`, the only falsy values are now: `null`, `undefined`, and `false`. This means `0` and empty strings will render as truthy. So for example, `v-bind:draggable="''"` will render as `draggable="true"`.
```
```suggestion
  When used with `v-bind`, the only falsy values are `null`, `undefined`, and `false`. This means `0` and empty strings will be rendered as truthy. For example, `v-bind:draggable="''"` will become `draggable="true"`.
```

src/v2/api/index.md
```diff
@@ -2197,6 +2197,11 @@ type: api
...
+  
+  For enumerated attributes, in addition to the falsy values above, the string `"false"` will also render as `attr="false"`.
```
```suggestion
  For enumerated attributes, in addition to the falsy values above, the string `'false'` will also be rendered as `attr="false"`.
```

phanan: Can we have an example to make this statement clearer? I'm not sure a beginner will get what "enumerated attributes" are.

starkovsky: added

starkovsky:
All changes committed

Justineo:
Actually there are at least four different ways for Vue to handle `v-bind`.

1. For some attribute/element pairs, Vue is always using the corresponding IDL attribute (property): [like `value` of `<input>`, `<select>`, `<progress>`, etc](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L11-L18).

2. For “[boolean attributes](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L33-L40)” and [xlinks](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L44-L46), Vue removes them if they are [`undefined`, `null` or `false`](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L52-L54) and adds them otherwise (see [here](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/runtime/modules/attrs.js#L66-L77) and [here](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/runtime/modules/attrs.js#L81-L85)) .

3. For “[enumerated attributes](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L20)”, Vue tried to do some [coercion](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/util/attrs.js#L24-L31) before setting the value. This is kind of problematic and we'll drop the behavior in Vue 3. See https://github.com/vuejs/vue/pull/8731 https://github.com/vuejs/vue/pull/8735.

4. For other attributes, we remove “falsy” values (`undefined`, `null`, or `false`) and set other values as-is (see [here](https://github.com/vuejs/vue/blob/bad3c326a3f8b8e0d3bcf07917dc0adf97c32351/src/platforms/web/runtime/modules/attrs.js#L92-L113)).

The added section is not quite accurate because `draggable` is now an “enumerated attributes” but `<input>`'s `checked` is not, instead it belongs to the “always use properties” category.

starkovsky:
@Justineo Thank you very match.
@NataliaTepluhina I am not sure, what do you think about last edit?

Justineo:
Though the edited version is technically more accurate, I'm not sure if putting all these into the docs is a good idea because this might lead to leak of abstraction. Ideally `v-bind` should “just work” but in practice edge cases are inevitable and users might be confused because of our “internal magic”.

WDYT? @chrisvfritz @NataliaTepluhina @phanan @sdras

NataliaTepluhina:
@Justineo I can totally understand your concerns. If we were going to put this into `Guide` part of the docs I'd say we should not add to users' cognitive load. However, this contribution touches API reference and I think we should have this explanation there for corner cases.

PR #2198, opened 2019-06-13
Add vue-daval data validator

deulos:


phanan:
Thanks for your PR. Your plugin looks cool indeed, but as of current, its user base is [too small](https://github.com/deulos/vue-daval/network/dependents) to be a significant example. If you need more exposure, I'd suggest starting with [vue-awesome](https://github.com/vuejs/awesome-vue). Cheers!

PR #2201, opened 2019-06-17
Clarify replacing an entire array.

sustained:
I'm not sure if this is really necessary or not, it's a very small change but I just noticed that the docs don't explicitly state that this is okay and thought that it might be good if they did.


phanan:
Thanks for the PR, but I don't think this is necessary – it doesn't provide any new information over the section right before it.

Issue #2195, opened 2019-06-12
Docs improvements suggestion


stratboy:
Hi! I’d like to suggest you to improve the group/list transition docs by specifying that the css transition classes will be applied to inner elements and not to the group/container itself. Took me a while to get it.

phanan:
We're open to a PR 👐

krishnanandula:
@stratboy or @phanan  which section would you like to suggest to make this change? 
[Transition-classes](http://localhost:4000/v2/guide/transitions.html#Transition-Classes)  or[ List-Transitions](http://localhost:4000/v2/guide/transitions.html#List-Transitions) ?

NataliaTepluhina:
Closing this as related PR has been merged.

PR #2207, opened 2019-06-21
added an additional hint for transition-classes

krishnanandula:
with Ref to the issue reported here https://github.com/vuejs/vuejs.org/issues/2195.

Please review and let me know if this is the right place to add the additional hint on transition classes.


NataliaTepluhina:
@krishnanandula thank you for working on this PR! The subject of the mentioned issue was a clarification about group/list transition; so I think it's better to move an additional hint to `List transitions` section.

krishnanandula:
@NataliaTepluhina thanks for the clarification, I will move 
under List Transition section.

krishnanandula:
@NataliaTepluhina  done ! can you please check and let me know if that's the right place to add? or please suggest any changes if needed.

NataliaTepluhina:
@krishnanandula thank you! Now it's definitely a right place 👍

PR #2209, opened 2019-06-21
Update references to legacy vue-cli template system

afontcu:
Hi! 👋 Just noticed that the section mentioned the old vue-cli templates, so I updated it to reflect the not-so-new-anymore vue cli preset/plugin system.

Feel free to suggest any rewording or additional modification you see fit! :)


phanan:
LGTM, thanks!

PR #2215, opened 2019-06-27
Document watch immediate behavior

### LINKED ISSUE #2213: Document watch immediate behavior
### What problem does this feature solve?
There is some confusion in the timing involved for `vm.$watch.immediate`

The first call to the provided callback happens before the teardown is created and therefore cannot be unwatched at the first call.
https://github.com/vuejs/vue/blob/0948d999f2fddf9f90991956493f976273c5da1f/src/core/instance/state.js#L346-L348

Several people have logged this thinking it's an undesired behavior.
https://github.com/vuejs/vue/issues/2276
https://github.com/vuejs/vue/issues/9134
https://github.com/vuejs/vue/issues/9501

It would help greatly to have this behavior documented so that future users don't encounter this error.

### What does the proposed API look like?
https://vuejs.org/v2/api/#vm-watch

```js
// Will cause an error, Undefined is not a function
// teardown is only available after first call.
const teardown = this.$watch('value', () => { teardown() }, { immediate: true });
```

Use a checking pattern instead.
```js
if (this.value)
   doTheThing();
else {
   const teardown = this.$watch('value', () => {
      doTheThing();
      teardown();
   });
}
```

NataliaTepluhina:
Close #2213

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1601,7 +1601,7 @@ type: api
       }
     },
     { immediate: true }
-  );
+  )
   ```
 
 ### vm.$set( target, propertyName/index, value )
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1580,7 +1580,7 @@ type: api
 
   ``` js
   // This will cause an error
-  const unwatch = this.$watch(
+  const unwatch = vm.$watch(
     'value',
     () => {
       doSomething()
@@ -1592,7 +1592,7 @@ type: api
   If you still want to call an unwatch function inside the callback, you should check its availability first:
 
   ``` js
-    const unwatch = this.$watch(
+    var unwatch = vm.$watch(
     'value',
     () => {
       doSomething()
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1580,9 +1580,9 @@ type: api
 
   ``` js
   // This will cause an error
-  const unwatch = vm.$watch(
+  var unwatch = vm.$watch(
     'value',
-    function() {
+    function () {
       doSomething()
       unwatch()
     },
@@ -1594,7 +1594,7 @@ type: api
   ``` js
     var unwatch = vm.$watch(
     'value',
-    function() {
+    function () {
       doSomething()
       if (unwatch) {
         unwatch()
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -1576,6 +1576,34 @@ type: api
   // `callback` is fired immediately with current value of `a`
   ```
 
+  Note that with `immediate` option you won't be able to unwatch the given property on the first callback call. 
+
+  ``` js
+  // This will cause an error
+  const unwatch = this.$watch(
+    'value',
+    () => {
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
@sdras @phanan could you please take a look at this PR? 🙏

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
   // `callback` is fired immediately with current value of `a`
   ```
 
+  Note that with `immediate` option you won't be able to unwatch the given property on the first callback call. 
+
+  ``` js
+  // This will cause an error
+  const unwatch = this.$watch(
```
phanan: We have a convention to use ES5 only (except for cookbook), which means no `const` or arrow functions.

phanan: I'd also suggest using `vm` instead of `this` to match the other examples.

NataliaTepluhina: Fixed 😅

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
   // `callback` is fired immediately with current value of `a`
   ```
 
+  Note that with `immediate` option you won't be able to unwatch the given property on the first callback call. 
+
+  ``` js
+  // This will cause an error
+  const unwatch = vm.$watch(
```
phanan: You still have `const` here 😅

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
...
+    'value',
+    function() {
```
phanan: Extreme nitpicking mode activated: I believe Vue CS leave a space between `function` and the opening bracket.

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
...
+    },
+    { immediate: true }
+  )
+  ```
+  If you still want to call an unwatch function inside the callback, you should check its availability first:
+
+  ``` js
+    var unwatch = vm.$watch(
+    'value',
+    function() {
```
phanan: Here as well.

src/v2/api/index.md
NataliaTepluhina: Fixed!

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
...
+    var unwatch = vm.$watch(
+    'value',
+    function () {
+      doSomething()
+      if (unwatch) {
+        unwatch()
+      }
+    },
+    { immediate: true }
+  );
```
```suggestion
  )
```

src/v2/api/index.md
```diff
@@ -1576,6 +1576,34 @@ type: api
...
+      doSomething()
+      unwatch()
+    },
+    { immediate: true }
+  )
+  ```
+  If you still want to call an unwatch function inside the callback, you should check its availability first:
+
+  ``` js
+    var unwatch = vm.$watch(
```
```suggestion
  var unwatch = vm.$watch(
```

NataliaTepluhina:
@phanan thank you so much for your nitpicking! 🙏

PR #2217, opened 2019-07-02
[Doc EN]: Some line return for consistency

Haeresis:
I think this one will be really quick @chrisvfritz ;)


sdras:
Thank you @Haeresis, looks good :)

PR #2219, opened 2019-07-03
Add another caveat I found.

xinbenlv:


Justineo:
This is not correct. Vue will take care of array methods for you:

https://jsfiddle.net/Justineo/rjgLwm10/

xinbenlv:
@Justineo  Thanks for a quick response.
Here is a shortest version I have for my case.
https://jsfiddle.net/mzL31sfq/2/

Justineo:
https://vuejs.org/v2/guide/list.html#Object-Change-Detection-Caveats

https://jsfiddle.net/Justineo/n1buf3sp/

phanan:
Thanks for the PR, but the information is incorrect as @Justineo has pointed out.

PR #2222, opened 2019-07-04
Fix logical mistake in code sample

utlime:
hasRun should be defined before callback


phanan:
Thanks for the PR, but the current code works just fine actually :).

utlime:
Hey @phainv , yeah, but it confuse when try to figure out why it needed =\
I think better to fix it

PR #2226, opened 2019-07-08
Would adding js codes for the rawHtml example make it more friendly?

OneQuid:
Would adding js codes for the `rawHtml` example make it any more friendly to beginners of either Vuejs or html/js?
I am new to Vuejs, and personally it does make things clearer.
Thank you for your efforts on Vuejs.


phanan:
Thanks for the PR, but I personally think it's not necessary. Not to mention, your `rawHtml` string doesn't 100% match the value rendered 😉.

PR #2227, opened 2019-07-10
Line 97: Instead.., ...instead.

bambalamm:
Instead shouldn't be used twice in one sentence like this (I think).


phanan:
> Instead shouldn't be used twice in one sentence like this (I think).

I happen to think so as well 😉.

PR #2231, opened 2019-07-12
Editing computed setter example to work must properly.

fthiagogv:
On the get function was inserting a default space, making the usage confusing.
On the set function without the condition the value inputted is doubled between firstName and lastName.


phanan:
Thanks, but the current code works just fine with the implied assumption of having both `firstName` and `lastName` available (just like the other samples above it). Also, we don't use ES6 syntax throughout the docs.

PR #2232, opened 2019-07-13
Added new CDN

praisedpk:
PageCDN tightly compresses files with brotli (level 11) compression. For vue.js it **saves additional 11KB** in compressed file size. So, can be a valuable addition to the docs.

Further, if you are open to work with us, we can setup the vue repo on PageCDN to fetch updates through webhooks as new releases are available on github. This way, you can easily display PageCDN links in the docs as you currently do with unpkg and give vue users benefit of smaller compressed files and [other features](https://pagecdn.com/features) like immutable caching, etc.


phanan:
Unfortunately at the moment we’d like to only suggest the most proven and popular CDN’s among developers instead. Thanks for the PR, though!

PR #2236, opened 2019-07-14
Added section discussing computed array

hitautodestruct:
Added a section that explains the use of computed setters on an array. Specifically the use of methods such as push to manipulate a computed property.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -181,6 +181,35 @@ computed: {
 
 Now when you run `vm.fullName = 'John Doe'`, the setter will be invoked and `vm.firstName` and `vm.lastName` will be updated accordingly.
 
+### Computed Setter with Arrays
+
+When setting the value of a complex type such as an `Array`, methods that appear to change only part of the data such as `.push()` or `.splice()` will return the entire array to the setter.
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR, but I find it pretty confusing. There's no acessing/mutating to the computed property `participants` in your example, for instance. Even if you meant for `this.names` to be `this.participants` instead, it would be a getter and not a setter. Can you explain?

hitautodestruct:
Yes, you're right the push was on `this.participants` and note `this.names`, I fixed the commit.

Basically the idea is to let the reader understand that when dealing with setters on an array when using `.push`,`.shift` etc. vue will pass in the entire array and not just the single `.push`ed property.

phanan:
It's incorrect then. `participants` appears to be a string (its getter uses `Array.join()` which returns a string), so calling `participants.push(fullName)` will simply not work. There's no mutator in your code as well – like I said above, you have an accessor.

hitautodestruct:
@phanan I see now, what you mean.
I'm assuming that because I'm not returning an array in `get()` I'm also not able to `set()` the array?
This behaviour should be noted in the docs.

PR #2239, opened 2019-07-18
Update plugins.md

DamianGlowala:
Propositional logic issue

### DIFF
--- a/src/v2/guide/plugins.md
+++ b/src/v2/guide/plugins.md
@@ -37,7 +37,7 @@ Vue.use(MyPlugin, { someOption: true })
 
 `Vue.use` automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/plugins.md
```diff
@@ -37,7 +37,7 @@ Vue.use(MyPlugin, { someOption: true })
 
 `Vue.use` automatically prevents you from using the same plugin more than once, so calling it multiple times on the same plugin will install the plugin only once.
 
-Some plugins provided by Vue.js official plugins such as `vue-router` automatically calls `Vue.use()` if `Vue` is available as a global variable. However in a module environment such as CommonJS, you always need to call `Vue.use()` explicitly:
+Some official plugins provided by Vue.js such as `vue-router` automatically calls `Vue.use()` if `Vue` is available as a global variable. However in a module environment such as CommonJS, you always need to call `Vue.use()` explicitly:
```
NataliaTepluhina:
```suggestion
Some official plugins provided by Vue.js such as `vue-router` automatically call `Vue.use()` if `Vue` is available as a global variable. However in a module environment such as CommonJS, you always need to call `Vue.use()` explicitly:
```

phanan: We can even remove "provided by Vue.js" because that's literally what "official" means.

PR #2241, opened 2019-07-18
Adding tip about dynamic input components.

Stegosource:
There was nothing in the docs about <components> binding to `value.prop` instead of just `value` for <input> elements.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -603,6 +603,8 @@ In the example above, `currentTabComponent` can contain either:
 
 See [this fiddle](https://jsfiddle.net/chrisvfritz/o3nycadu/) to experiment with the full code, or [this version](https://jsfiddle.net/chrisvfritz/b2qj69o1/) for an example binding to a component's options object, instead of its registered name.
 
+<p class="tip">If you are creating an &lt;input&gt; element with the dynamic component and need to bind to the `value` attribute, you will need to append the `.prop` modifier to the `value` attribute like so `:value.prop="yourData"`</p>
+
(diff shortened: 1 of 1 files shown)


Stegosource:
Hi @sdras, thanks for the feedback. It's something I would consider a caveat, but without knowing about it, there is not a great place to find this info. With that said, I also think it's more of a rare use-case so I didn't want to add too much extra noise for the average user. 

I'm happy to make changes, but maybe you have some thoughts on this? Do you think it should be written into this document? Should we add a small "Caveats" section? Does it makes more sense in the cookbook?

fer22f:
This is very very important. So important in fact that I have written this independently, because I ran into this gotcha too, but you were faster!

[I documented it like this](https://github.com/fer22f/vuejs.org/commit/1a82eb30253a0be58a4652458ef977b8e60cf559):

Keep in mind that this attribute can be used with regular HTML elements, however they will be treated as components, which means all attributes **will be bound as DOM attributes**. For some properties such as `value` to work as you would expect, you will need to bind them using the [`.prop` modifier](../api/#v-bind).

---

I hope that helps ;) because we need this documented.

Stegosource:
@fer22f Yeah, I also implemented it with `input` and `textarea`. I like your version too. I think I'm going to edit this PR to go into a bit more details and also include a link to a codesandbox or something.

sdras:
@Stegosource, any progress on this? Thanks!

sdras:
@Stegosource @fer22f, the example that @fer22f provided is exactly what we're looking for and though an example would be great, it's not too necessary- given the amount of time elapsed on this, let's close this out and see if @fer22f will PR his in. Thanks!

Stegosource:
Yeah, sorry for the delays on my part. I was meaning to get back to this, but kept putting it off. I agree with adding @fer22f 's version as it's closer to the ideal docs.

PR #2244, opened 2019-07-19
Fix font size for inline code

phanan:
Currently we're using root unit `rem` for the code, which renders code inside headers pretty weird:

image

This commit replaces `rem` with `em` so that inline code has its size related to the direct parent instead:

image


NataliaTepluhina:
Totally voting for this change :hugs:

PR #2246, opened 2019-07-23
Add: Constraint reusing data object in createElement

Grovespaz:
Reusing a data object breaks things, such as events.

I was asked by @posva to document this and create a PR.

### DIFF
--- a/src/v2/guide/render-function.md
+++ b/src/v2/guide/render-function.md
@@ -309,6 +309,23 @@ render: function (createElement) {
 }
 ```
 
+#### Data Objects Must Be Unique
+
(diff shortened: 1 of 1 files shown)


posva:
I will let the docs team look at this but I think it makes more sense to integrate the non-duplication of data into the previous example instead of adding a new section

Grovespaz:
(I forgot to add this link last Tuesday, the PR originated from this issue: https://github.com/vuejs/vue/issues/10303#event-2502744994)

phanan:
I agree with @posva. Also, the code example is syntactically incorrect.

PR #2249, opened 2019-07-26
I suppose hyphen is not required for one word names

aurimus:


sodatea:
Have you followed the link in this paragraph?
One-word names are not valid for custom components.

PR #2250, opened 2019-07-26
What does $attrs contain

aurimus:


phanan:
Thanks, but `$attrs` is actually explained [in the API section](https://vuejs.org/v2/api/#vm-attrs).

PR #2251, opened 2019-07-26
Not very clear what .sync does

aurimus:


phanan:
Thanks for the PR, but I, on the other hand, think it's quite clear what it does: 

> For convenience, we offer a shorthand for [the aforementioned pattern] with the `.sync` modifier.

So as long as you understand the pattern, you will understand `.sync`.

PR #2252, opened 2019-07-27
Another technique to stagger list transitions

ubershmekel:


phanan:
Thanks for the PR, but the technique has nothing to do with Vue (as you said yourself 🙂) so it doesn't make a lot of sense to include it into Vue's docs.

ubershmekel:
@phanan there's a technique mentioned in the current docs that works only with a 3rd party library (Velocity). The technique in this PR works just using vue's template, without a need for a 3rd party library. It makes much more sense than what's already there.

Perhaps that note is confusing. It's meant to say the technique can be used with vanilla js, though vue does make it easier thanks to ``:style="'--animation-order:'`

PR #2258, opened 2019-08-01
Update components-edge-cases.md

facka:
Add other example on how to use Webpack's asynchronous import.

### DIFF
--- a/src/v2/guide/components-edge-cases.md
+++ b/src/v2/guide/components-edge-cases.md
@@ -313,6 +313,13 @@ components: {
   TreeFolderContents: () => import('./tree-folder-contents.vue')
 }
 ```
+or
+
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR! Unfortunately, I don't think the additional snippet adds much value in this context.

facka:
@phanan yeah, you are right. Thanks for reviewing my PR. I added this example because the issue I found is that it seems that when converting the name to match the html tag is not using the name property in the component, instead is converting the name set there to kebab-case.
I realized this when I named the import `TFContents `, then the html tag should be `t-f-contents ` even if the component name is `tf-contents `.
This is way I think is better to use the kebab-case notation when doing Webpack async import

PR #2259, opened 2019-08-01
Update footer ad format

sayzlim:
Remove the icon and style it to match the docs theme. Refer to the screenshot below for the new design:

### DIFF
--- a/themes/vue/source/css/_ad.styl
+++ b/themes/vue/source/css/_ad.styl
@@ -1,40 +1,41 @@
 // main ad placement (bottom right)
 #ad
-  width: 125px
-  // text-align: center
-  position: fixed
-  z-index: 99
-  bottom: 10px
-  right: 10px
-  padding: 10px
-  background-color: #fff
-  border-radius: 3px
-  font-size: 13px
+  width 125px
+  position fixed
+  z-index 99
+  bottom 10px
+  right 10px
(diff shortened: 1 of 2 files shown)


phanan:
Thanks for the PR @sayzlim!
@yyx990803 I don't see any problem, but it would be better if you take a look as well, I guess.

themes/vue/source/css/_ad.styl
```diff
@@ -30,41 +30,37 @@
...
-        height 20px
-        border-radius 3px
-        vertical-align middle
-        position relative
-        top -1px
-    .default-title
+#bsa-native
+  .native-box
+    box-shadow inset 0 0 0 1px hsla(0, 0%, 0%, .075)
+    background-color: #f8f8f8;
```
phanan: You can remove the colon in these rules btw.

sayzlim: @phanan thanks for reviewing! I’ve removed the colon in these parts. Do we want to remove all the colons in this stylesheet?

phanan: I'd say it would be nice for you to do so, as long as we don't break anything :D

sayzlim: Please review one more time. Should be good! :)

PR #2260, opened 2019-08-04
fix: reverse message with unicode

magic-akari:
The Introduction give an example code to reverse message.
But it is not the perfect way to do that.
When the strings contain unicode such as emoji, the code will give us an unexpected result.

There are some similar code in `computed` section and `Adding Instance Properties` section.

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -213,11 +213,11 @@ To let users interact with your app, we can use the `v-on` directive to attach e
 var app5 = new Vue({
   el: '#app-5',
   data: {
-    message: 'Hello Vue.js!'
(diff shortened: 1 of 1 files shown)


Justineo:
The example code is intentionally using ES5 and although the modified version is technically more correct, I don't think it's necessary here because it doesn't help the understanding of the concepts we are introducing.

phanan:
@Justineo's feedback is spot on. Thanks for your effort @magic-akari, but we'll have to skip on this one 😛

magic-akari:
Is the ES6 Syntax unacceptable? How about this?

```diff
- this.message = this.message.split('').reverse().join('')
+ this.message = Array.from(this.message).reverse().join('')
```

PR #2264, opened 2019-08-09
Remove unnecessary JSON.stringify

3cp:
Pretty sure the JSON.stringify wrapper has no effect in webpack config file which is JS code.


3cp:
It was my mistake :-)

PR #2265, opened 2019-08-10
Darker red background for bad examples

saharbit:
Currently it feels awkward that you first see a bad example and it is not highlighted enough as one

https://i.imgur.com/CLigKEr.png
-->
https://i.imgur.com/VYSdyBH.png


phanan:
Thanks, the contrast is indeed better.

PR #2268, opened 2019-08-15
Add a link to the list of whitelisted globals

tbetous:
Add a link to the list of whitelisted globals (relative to issue #2266)

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -98,7 +98,7 @@ These expressions will be evaluated as JavaScript in the data scope of the owner
 {{ if (ok) { return message } }}
 ```
 
-<p class="tip">Template expressions are sandboxed and only have access to a <a href="https://github.com/vuejs/vue/blob/v2.6.10/src/core/instance/proxy.js#L9">whitelist of globals</a> such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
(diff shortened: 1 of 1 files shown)


src/v2/guide/syntax.md
```diff
@@ -98,7 +98,7 @@ These expressions will be evaluated as JavaScript in the data scope of the owner
 {{ if (ok) { return message } }}
 ```
 
-<p class="tip">Template expressions are sandboxed and only have access to a whitelist of globals such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
+<p class="tip">Template expressions are sandboxed and only have access to a <a href="https://github.com/vuejs/vue/blob/v2.6.10/src/core/instance/proxy.js#L9">whitelist of globals</a> such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
```
NataliaTepluhina: **nitpick**: while it's technically correct, I'd recommend replacing an `<a>` tag with a markdown link. It would be more readable and consistent with the rest of the documentation.

```suggestion
<p class="tip">Template expressions are sandboxed and only have access to a [whitelist of globals](https://github.com/vuejs/vue/blob/v2.6.10/src/core/instance/proxy.js#L9) such as `Math` and `Date`. You should not attempt to access user defined globals in template expressions.</p>
```

tbetous: Thanks for the review ! You are right ! I'll commit that right now !

PR #2269, opened 2019-08-15
fix: link to devel build for prototyping and minified build for prod

bimbiltu:
This is just something I thought was odd when going through the installation instructions


phanan:
Hmm… now I wonder why we have it that way from the first place.

PR #2270, opened 2019-08-16
docs: update index.md

Pasoul:


phanan:
Thanks for the PR, but the current version is correct.

PR #2275, opened 2019-08-22
Added a working example to Custom Events.

andregce:
While reading on Custom Events, I almost accidentally copied the non-working code.

I think the documentation would benefit from a working sample.

Here is my humble contribution to this great framework.

### DIFF
--- a/src/v2/guide/components-custom-events.md
+++ b/src/v2/guide/components-custom-events.md
@@ -25,6 +25,17 @@ Unlike components and props, event names will never be used as variable or prope
 
 For these reasons, we recommend you **always use kebab-case for event names**.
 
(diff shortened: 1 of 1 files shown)


phanan:
Thanks, but I do believe the line right above your added content is as already clear as it can be:

> For these reasons, we recommend you **always use kebab-case for event names**.

Besides, the grammar is broken :)

andregce:
Ooops sorry for the bad grammar. Got too excited trying to contribute.

But as a user I would like to insist the documentation for Custom Events is missing a working example. It's not that it is not clear, it is that the only example given is of how not to do it. Someone should add a working example.

PR #2276, opened 2019-08-24
Replace indexOf with includes in components-props

JafarAkhondali:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes, `includes` function widely supported


phanan:
Thanks, but we use ES5 intentionally.

PR #2281, opened 2019-09-01
Fix "Good" example at "Avoid v-if with v-for" rule

OziOcb:
Added the activeUsers computed property to the example.


phanan:
Thanks for the PR, but I don't think the addition is necessary for this context.

PR #2283, opened 2019-09-03
Highlight important text in Event Names example

cmitz:


phanan:
Thanks, but I don't think this change is necessary.

cmitz:
I agree that this change is probably not what that page requires.

However, I spend 15 minutes looking why my setup wasn't working, because I expected the codeblocks to be an example. It took me a while to realise it was an example of what not to do.

I suggest that the first example on the page should be a working example. I'm willing to provide that, if you think that is a good idea.

phanan:
Assuming you're talking about this block: 

```html
<!-- Won't work -->
<my-component v-on:my-event="doSomething"></my-component>
```	

I'd tend to believe that `<-- Won't work -->` brings a clear message :)

cmitz:
Yeah, but because the syntax highlighting that wasn't really clear.

I believe I'm not the only one in this case: I though I knew the syntax, but checked it for a second just in case. That literal second was definitely not enough for me to see the `<-- Won't work -->`

PR #2289, opened 2019-09-09
correct definition of modifiers

ockstadt:
<div id="hook-arguments-example" v-demo:foo.a.b="message"></div>

Colon must be a dot.


phanan:
Thanks, but the current example is correct :)

PR #2290, opened 2019-09-09
update Simple State Management from Scratch example

willhatchVista:
we're talking about mutating a property, lets use a mutable value


NataliaTepluhina:
Thank you for your impact! However, ES2015 `const` doesn't mean the value is _immutable_. The value of a constant can't be changed through reassignment, and it can't be redeclared; changing the properties of an object does mutate it but _does not_ reassign it. That's why using `const` is fine in our case.

PR #2293, opened 2019-09-10
Make the video overlay cover the header

phanan:
The video overlay's z-index is currently not large enough (10) to cover the site's header (100). This should fix the problem.

Before:
image

After:
image


phanan:
Actually, with the last change from @bencodezen this issue has already been fixed, so I'll see myself out :)

PR #2294, opened 2019-09-12
Introduce performance regression CI check

grgur:
Using [Gimbal](https://github.com/ModusCreateOrg/gimbal) and [GitHub Actions](https://github.com/vuejs/vuejs.org/actions) to ensure Performance Budgeting on the VueJS.org homepage.

This CI action will
* [x] ensure lighthouse tests don’t fall below thresholds
* [x] ensure a11y rules are observed
* [x] ensure heap memory doesn’t go over the threshold
* [x] automatically create a PR comment with items that are failing

The thresholds are loose to reflect the current state of the built site. Some of the tests are skipped as they are taken care of on the server side post-deployment.

Gimbal is not made a dependency to keep the dependency tree light (apart from the light axe plugin). However, it can be executed locally via a CLI command as such:

```sh
npx @modus/gimbal
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,5 +28,8 @@
     "hexo-server": "^0.3.1",
     "hoek": "^6.1.2",
     "request": "^2.85.0"
+  },
+  "devDependencies": {
+    "@modus/gimbal-plugin-axe": "^1.2.3"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,14 @@
 # yarn lockfile v1
 
 
+"@modus/gimbal-plugin-axe@^1.2.3":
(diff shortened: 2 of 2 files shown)


sdras:
@bencodezen since you are leading the charge for vuepresss migration, I think your input on this is important

bencodezen:
@sdras Thanks for pinging me. Sorry for missing the notification on this one and not seeing it till now.

I'll definitely need to play around with Gimbal to see how it interacts with VuePress. Will keep you posted. 👍

grgur:
Thanks @bencodezen and @sdras. Please let me know if I can help with anything there.

Issue #2297, opened 2019-09-13
Hamburger menu hidden on mobile


bencodezen:
On mobile, the hamburger icon is covered by the sidebar.

![Screen Recording 2019-09-13 at 03 50 PM](https://user-images.githubusercontent.com/4836334/64890719-4a0a0d80-d63e-11e9-8510-f6a3262469fe.gif)

Hat tip to @phanan for discovering the bug!

NataliaTepluhina:
Closing as related PR was merged

PR #2298, opened 2019-09-13
#2297 - Fix hidden hamburger menu mobile

bencodezen:
Original fix was simply updating the z-index, but felt it was time to establish a z-index manifest in order to prevent random one-offs in the future that can cause side effects like the one we saw in this bug.


NataliaTepluhina:
Thank you Ben! This is a really great improvement to our z-indexes 👍

PR #2300, opened 2019-09-14
2299 duplicate mobile menu items

bencodezen:
Ended up fixing a few issues here:

1. Made the home page "main menu" only show up on home page
2. Noticed that the sponsors were oddly aligned on mobile and gave it a little more breathing room
3. Noticed the sponsors were aligned too far down the page in comparison to the content

Feedback welcome if I misinterpreted any original design intentions.

### DIFF
--- a/themes/vue/source/css/_sidebar.styl
+++ b/themes/vue/source/css/_sidebar.styl
@@ -74,7 +74,7 @@
     -webkit-transform: translate(-280px, 0)
     transform: translate(-280px, 0)
     .sidebar-inner
-      padding: 50px 10px 10px 20px
+      padding: 60px 10px 10px 20px
       box-sizing: border-box
     .sidebar-inner-index
       padding: 10px 10px 10px 20px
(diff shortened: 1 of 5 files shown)


themes/vue/source/css/_sidebar.styl
```diff
@@ -74,7 +74,7 @@
     -webkit-transform: translate(-280px, 0)
     transform: translate(-280px, 0)
     .sidebar-inner
-      padding: 50px 10px 10px 20px
```
phanan: Can we take this opportunity to increase the bottom padding of `.sidebar-inner` to something like 60px as well? The current value (10px) makes it almost possible to tap the last menu item on mobile devices.

image

bencodezen: @phanan I've updated the bottom padding to 30px since the 60px was to account for the header and it would create too much gap.

Screen Shot 2019-09-16 at 10 18 10 AM

Let me know if you think we still need more space!

PR #2304, opened 2019-09-16
Style Guide: Elaborate on “Non-Flux State Management” (vuejs#1224)

lhenriqueanjos:


phanan:
Thanks for the effort, but I don't think it's really necessary (and the URL is localhost 😃).

lhenriqueanjos:
Thanks. Since the issue is still open, how could I have been more effective besides using the correct URL 😅?

phanan:
The issue has been closed though. And the URL is only a minor issue. What's more important is, your addition didn't really "add" anything. Here it is again:

> It is the [Official Flux-Like Implementation](http://localhost:4000/v2/style-guide/#Non-flux-state-management-use-with-caution) for Vue. The use of a flux-like implementation results in an application that is easier to debug, because you can keep track of all state mutations.

And this is what we already have:

> [...] Vuex offers not only a central place to manage state, but also tools for organizing, tracking, and debugging state changes.

I believe you'd agree that they more or less convey the same idea :)

lhenriqueanjos:
Yes, I agree! I will look for another way to contribute and add real value. Thanks for the feedback!

PR #2305, opened 2019-09-19
Fix position of '$' for prices on Theme page

tinymachine:
(In the U.S. common practice is to place the dollar symbol before the numeric value.)


NataliaTepluhina:
Thanks for the fix!

phanan:
If we want to do it properly, perhaps extract a `ProductItem` component and create a `displayPrice` computed property e.g.:

```js
displayPrice: function () {
  return this.product.price.toLocaleString('en', { 
    style: 'currency', 
    currency: 'USD' 
  })
}
```

PR #2309, opened 2019-09-22
Update .gitignore

hillaramaciej:


phanan:
Thanks for the effort, but this change is not necessary.

PR #2313, opened 2019-09-25
Corrected a parameter error

Zhaohaoran1997:
The parameter is different from the slot definition to its call.


Justineo:
`todos` is one of the props of the `TodoList` component and it is correct. Thanks.

Issue #2307, opened 2019-09-19
async/await mentioned at the end of page "Reactivity in Depth" should be part of ES 2017, not ES 2016


tomoya06:
[深入响应式原理](https://cn.vuejs.org/v2/guide/reactivity.html)一文最后一段，建议"可以使用新的 ES2016 async/await 语法完成相同的事情"。但根据[MDN文档](https://developer.mozilla.org/zh-CN/docs/Web/JavaScript/Reference/Statements/async_function)的说明，async/await特性“初始定义于ES2017”，且纳入ES2017标准。请问原文档是否应修正为“ES 2017...语法”呢？谢谢。

祝顺利

Jinjiang:
This issue is questioning about `async`/`await` we mentioned at the end of page [Reactivity in Depth](https://vuejs.org/v2/guide/reactivity.html) should be [part of ES 2017, not ES 2016](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Specifications).

phanan:
…Which is correct. Async/await is ES2017. @tomoya06 would you want to send a PR over to vuejs/vuejs.org?

Jinjiang:
Let me do this. :-)

PR #2314, opened 2019-09-25
Small fix in async/await introduction

Jinjiang:
Changed async/await from ES 2016 into ES 2017. #2307


NataliaTepluhina:
@Jinjiang thanks for the fix! 😅

PR #2316, opened 2019-09-27
Add rel=sponsored to partner links

phanan:
Adding `rel=sponsored` into sponsored and partner links. Let me know if I miss anything.


sdras:
Looks good. Just checked out and multiple rel attributes are valid as well, in case anyone is wondering.

PR #2317, opened 2019-09-27
Vue Docs Writing Guide

chrisvfritz:
This is a writing guide that the docs team has worked together to develop, to help provide rules and recommendations to ourselves and contributors for writing and maintaining docs across the Vue ecosystem.


writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Don't litter the document with tips and caveats.** If you find that more than two tips are needed per page, consider adding a caveats section to address these issues. The guide is meant to be read straight through, and tips and caveats can be overwhelming or distracting to someone trying to understand the base concepts.
+- **Avoid appeals to authority** (e.g. "you should do X, because that's a best practice" or "X is best because it gives you full separation of concerns"). Instead, demonstrate with examples the specific human problems caused and/or solved by a pattern.
+- **When deciding what to teach first, think of what knowledge will provide the best power/effort ratio.** That means teaching whatever will help users solve the greatest pains or greatest number of problems, with the relatively least effort to learn. This helps learners feel smart, powerful, and curious, so their cognitive capacity will drain more slowly.
+- **Unless the context assumes a string template or build system, only write code that works in any supported environment.**
```
phanan: What does "supported environment" mean in this case? (Or, can we elaborate that?)

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Specific is better than generic.** For example, a `<BlogPost>` component example is better than `<ComponentA>`.
+- **Relatable is better than obscure.** For example, a `<BlogPost>` component example is better than `<CurrencyExchangeSettings>`.
+- **Be emotionally relevant.** Explanations and examples that relate to something people have experience with and care about will always be more effective.
+
+### Grammar
+
+- **Avoid abbreviations** in writing and code examples (e.g. `attribute` is better than `attr`, `message` is better than `msg`), unless you are specifically referencing an abbreviation in an API (e.g. `$attrs`).
+- **When referencing a directly following example, use a colon (`:`) to end a sentence**, rather than a period (`.`).
+- **Use the oxford comma** (e.g. "a, b, and c" instead of "a, b and c"). ![Why the oxford comma is important](https://www.inkonhand.com/wp-content/uploads/2015/10/serialimage.jpg)
```
phanan: Isn't it "Oxford comma" (with a capital O)?

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **When referencing the name of a project, prioritize the broader conventions of English over internal branding conventions of that project.** For example, "webpack" and "npm" both disregard conventions such as "always start a word at the beginning of a sentence with a capital letter", "project names always use Title Case", and "acronyms are always capitalized". Instead, always write "Webpack and NPM" to provide a more consistent experience in the docs and avoid sentences like "If you don't want to use Vue CLI, you can use webpack or Rollup directly by installing them via npm or Yarn".
```
phanan: I'm not sure about this. Instead, I personally always respect the official name, for example, notice how it's always "npm" regardless of its position on the [Wikipedia entry](https://en.wikipedia.org/wiki/Npm_(software)).

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Use Title Case for headings** - at least for now, since it's what we use through the rest of the docs. There's research suggesting that sentence case (only first word of the heading starts with a capital) is actually superior for legibility and also reduces the cognitive overhead for documentation writers, since they don't have to try to remember whether to capitalize words like "and", "with", and "about".
```
phanan: If we use title case for headings (which I totally support), we can remove the following explanation.

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Always prefer simpler, plainer language over complex or jargony language.** For example:
+  - "you can use Vue with a script element" instead of "in order to initiate the usage of Vue, one possible option is to actually inject it via a script HTML element"
+  - "function that returns a function" instead of "higher order function"
```
phanan: In this particular case, I'd argue that "higher-order function" is a well-known term that is expected to be used _for_ the correct target audience. "Function that returns a functions" is more verbose and at the same time [insufficient](https://en.wikipedia.org/wiki/Higher-order_function).

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Describe the _problem_ first, then the solution.** Before showing how a feature works, it's important to explain why it exists. Otherwise, users won't have the context to know if this information is important to them (is it a problem they experience?) or what prior knowledge/experience to connect it to.
+- **While writing, don't be afraid to ask questions**, _especially_ if you're afraid they might be "dumb". Being vulnerable is hard, but it's the only way for us to more fully understand what we need to explain.
+- **Be involved in feature discussions.** The best APIs come from documentation-driven development, where we build features that are easy to explain, rather than trying to figure out how to explain them later. Asking questions at this stage (again, especially the questions we're afraid are "dumb") often help reveal potential confusions, inconsistencies, and problematic behavior before it would require a breaking change to fix.
+
+## Organization
```
phanan: Can we somehow keep this section more concise? I personally find it a bit too wordy.

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Migrations**:
+  - **Versions**: When important changes are made, it's useful to include a full list of changes, including a detailed explanation of why the change was made and how to migrate their projects.
+  - **From other projects**: How does this software compare to similar software? This is important to help users understand what additional problems we might solve or create for them, and to what extent they can transfer knowledge they already have.
+- **Style Guide**: There are necessarily some key pieces in development that need a decision, but are not core to the API. The style guide provides educated, opinionated recommendations to help guide these decisions. They shouldn't be followed blindly, but can help teams save time by being aligned on smaller details.
+- **Cookbook**: Recipes in the cookbook are written with some assumption of familiarity with Vue and its ecosystem. Each is a highly structured document that walks through some common implementation details that a Vue dev might encounter.
+
+## Writing & Grammar
```
phanan: _Very_ nitpicky: Let's avoid using abbreviations/symbols like "&" as well and replace it with "and."

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Almost always avoid humor (for English docs)**, especially sarcasm and pop culture references, as it doesn't translate well across cultures.
+- **Never assume a more advanced context than you have to.**
+- **In most cases, prefer links between sections of the docs over repeating the same content in multiple sections.** Some repetition in content is unavoidable and even essential for learning. However, too much repetition also makes the docs more difficult to maintain, because a change in the API will require changes in many places and it's easy to miss something. This is a difficult balance to strike.
+- **Specific is better than generic.** For example, a `<BlogPost>` component example is better than `<ComponentA>`.
+- **Relatable is better than obscure.** For example, a `<BlogPost>` component example is better than `<CurrencyExchangeSettings>`.
+- **Be emotionally relevant.** Explanations and examples that relate to something people have experience with and care about will always be more effective.
+
+### Grammar
```
phanan: We might suggest using a tool for grammar/spelling check, which can be tremendously useful for non-native speakers. For example, I've been extremely happy with Grammarly for English.

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Principles
+
+- **A feature doesn't exist until it's well documented.**
+- **Respect users' cognitive capacity (i.e. brain power).** When a user starts reading, they begin with a certain amount of limited brain power and when they run out, they stop learning.
+  - Cognitive capacity is **depleted faster** by complex sentences, having to learn more than one concept at a time, and abstract examples that don't directly relate to a user's work.
+  - Cognitive capacity is **depleted more slowly** when we help them feel consistently smart, powerful, and curious. Breaking things down into digestible pieces and minding the flow of the document can help keep them in this state.
+- **Always try to see from the user's perspective.** When we understand something thoroughly, it becomes obvious to us. This is called _the curse of knowledge_. In order to write good documentation, try to remember what you first needed to know when learning this concept. What jargon did you need to learn? What did you misunderstand? What took a long time to really grasp? Good documentation meets users where they are. It can be helpful to practice explaining the concept to people in person before
```
NataliaTepluhina: Added a slight rewording to avoid repeating the same adjectives

```suggestion
- **Always try to see from the user's perspective.** When we understand something thoroughly, it becomes evident to us. This is called _the curse of knowledge_. In order to write good documentation, try to remember what you first needed to know when learning this concept. What jargon did you need to learn? What did you misunderstand? What took a long time to really grasp? Proper documentation meets users where they are. It can be helpful to practice explaining the concept to people in person before
```

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Guide**: Make users feel smart, powerful, and curious, then maintain this state so that users maintain the motivation and cognitive capacity to keep learning more. Guide pages are meant to be read sequentially, so should generally be ordered from the highest to lowest power/effort ratio.
+  - **Essentials**: It should take no longer than 5 hours to read the Essentials, though shorter is better. Its goal is to provide the 20% of knowledge that will help users handle 80% of use cases. It can link to more advanced guides and the API, though these should be avoided in most cases and when provided, context should be provided so that users know they should  follow that link on their first reading. Otherwise, many users end up exhausting their cognitive capacity link-hopping, trying to fully learn every aspect of a feature before moving on, and as a result, never finish that first read-through of the Essentials. Remember that a smooth read is more important than being thorough. We want to give people the information they need to avoid a frustrating experience, but they can always come back and read further, or Google a less common problem when they encounter it.
```
NataliaTepluhina: `It can link to more advanced guides and the API, though these should be avoided in most cases and when provided, context should be provided so that users know they should  follow that link on their first reading.` - could we please reword/split this sentence a bit?

Maybe something like:
```
Essentials can link to more advanced guides and the API, though, in most cases, you should avoid such links. When they are provided, you need also provide a context so users are aware if they should follow this link on their first reading.
```

(would be also nice to have an example here)

writing-guide.md
```diff
@@ -0,0 +1,86 @@
...
+- **Style Guide**: There are necessarily some key pieces in development that need a decision, but are not core to the API. The style guide provides educated, opinionated recommendations to help guide these decisions. They shouldn't be followed blindly, but can help teams save time by being aligned on smaller details.
+- **Cookbook**: Recipes in the cookbook are written with some assumption of familiarity with Vue and its ecosystem. Each is a highly structured document that walks through some common implementation details that a Vue dev might encounter.
+
+## Writing & Grammar
+
+### Style
+
+- **Headings should describe problems**, not solutions. For example, a less effective heading might be "Using props", because it describes a solution. A better heading might be "Passing Data to Child Components with Props", because it provides the context of the problem props solve. Users won't really start paying attention to the explanation of a feature until they have some idea of why/when they'd use it.
+- **When you assume knowledge, declare it** at the beginning and link to resources for less common knowledge that you're assuming.
```
NataliaTepluhina:
```suggestion
- **When you assume knowledge, declare it** at the beginning and link to resources for less common knowledge that you're expecting.
```

writing-guide.md
phanan: If it's allowed, we should host this image in our repo to avoid possible link breakage in the future.

writing-guide.md
chrisvfritz: This is definitely one of the more controversial conventions. 😄 Opinions also seem to differ wildly across cultures. But _generally_ among the English-speaking population of the world, my informal observation has been that this really helps people who have not yet developed into experts in these ecosystems, because the branding conventions actually look like typos and disrupt their flow while reading. 

The people I see complain about _not_ respecting branding conventions are almost always ecosystem experts (and often technical writers themselves). While this group tends to be especially vocal, they're a minority and will also have the lowest baseline cognitive burden when reading, so they don't need as much help.

Does that make sense?

writing-guide.md
chrisvfritz: If we stick with title case though, we should eventually replace this with a longer explanation/list of which words should not be capitalized (e.g. "and", "with", etc). 😅 What's your reason for preferring title case over sentence case?

writing-guide.md
chrisvfritz: I'm open to changing this example, but here's my thinking. While "higher-order function" and even "higher-order component" may be well-_known_ terms (i.e. many people - or at least those more experienced in the field or coming from a computer science background - have heard of them and even used them), I would argue they're not at all well-_understood_. If I did an offline poll in an advanced workshop/meetup/conference, I feel very confident that the majority of people would not be able to give a complete and accurate definition of either. It might even be _impossible_ to give an "accurate" definition, since they seem to have different definitions depending on where you look. 😄

And their technical definitions often differ even further from how they're used by working developers. For example, I've never heard someone refer to a function that just accepts a callback a "higher-order function", but it fits the technical criteria. So do many constructors, if you count functions that return an object that has methods on it (or all constructors, if you count getters and setters). The most widely accepted technical definition of a "higher-order function" is possibly so vague as to be useless.

By avoiding jargon like this altogether and getting right to what we mean in a particular use case with "function that returns a function", we're clearer _and_ more specific. At the cost of only two extra syllables, that seems like a deal! 😄

Do you disagree?

phanan: If such a term is controversial, shouldn't we avoid it altogether and use another as an example?

writing-guide.md
chrisvfritz: I tried. 😅 You can let me know what you think. I'm open to specific suggestions.

writing-guide.md
phanan: > because the branding conventions actually look like typos and disrupt their flow while reading.

I honestly don't think this is a convincing enough reason to _not_ respect a name/branding. If I have a product named "foo," I would appreciate if it's written "foo" everywhere. At least that's what my years working in an agency taught me :)

writing-guide.md
chrisvfritz: Hm, what's your reasoning? And are you thinking everywhere, or only in non-titles?

writing-guide.md
phanan: I'm not preferring one case over another though 😃. My point is, if we're suggesting Title Case ("at least for now, since it's what we use through the rest of the docs") then there's little need to provide more information about _the other case(s)_ as it would only confuse the users I reckon.

writing-guide.md
phanan: My reasoning is actually pretty simple: We're discouraging abbreviations elsewhere (in this same document), right? "&" to me is another abbreviation 😛

writing-guide.md
chrisvfritz: I think "obvious" is actually closer to what I mean here than "evident". To me, "obvious" suggests not only understanding, but _ease_ of understanding. Once we learn something, we forget how difficult it was to learn and how difficult it may still be for others. "Evident" doesn't carry the same connotation.

Regarding "good" vs "proper", I think this is one of the occasions where repeating an adjective is actually OK (and even desirable). We're repeating an adjective in the same context, to reference a previous concept: "good documentation". Changing the second adjective to "proper" would suggest that "proper documentation" is a new concept separate from "good documentation". In general though, you're right! Repeating the same adjectives across different contexts (e.g. "during my interesting walk I saw an interesting cat") can make the writing less, um, interesting. 😜 However, this is also somewhat in conflict with our goal to keep language simpler. If our option is to replace a word with a much less common one, I think I'd prefer repeated adjectives over forcing users to expand their English vocabulary.

writing-guide.md
chrisvfritz: > If I have a product named "foo," I would appreciate if it's written "foo" everywhere.

And I'd appreciate if "foo" didn't choose a naming convention that violated larger grammar conventions, but I won't try to force them to change it on _their_ site, so I feel like I'm being more reasonable. 😄 That doesn't really tell me why their preference for lower-case is more important than the experience of our readers. I feel like respecting their branding conventions would be placing the desires of the powerful few (these big projects) over the needs of the vulnerable many (our readers, especially those who are already dealing with a significant cognitive burden from being new to the ecosystem/programming or not being fluent in English).

Can you explain your preference in more detail? Do you feel we're doing actual harm to these projects or our readers?

writing-guide.md
chrisvfritz: Ahh, yeah, I agree. I just include this for now because I'd eventually like to push for a transition to sentence case, so I want to give people time to think about it and form opinions in the meantime.

writing-guide.md
chrisvfritz: I don't think these terms are widely considered controversial - they're very widely used and I see very few people questioning their use.

writing-guide.md
chrisvfritz: My reasoning for avoiding abbreviations is that some people might not be familiar with them, or even when familiar, might have to translate in their heads to something _more_ familiar. I don't have the same worry with symbols that are on everyone's keyboards, like `&`, `%`, `#`, and `@`, even though you could argue that all of them started as abbreviations. By now, they're essentially punctuation. In my experience, even those who have never used a computer are familiar with them. Is your experience different?

writing-guide.md
chrisvfritz: I don't know where the image originally came from (it doesn't have a watermark and seems to appear on many sites), but I'll pull it in and we can replace it if someone makes a claim.

writing-guide.md
phanan: I would argue that it's easier to draw the line between _full words and abbreviations_ than between _less and more familiar_, but it's not extremely important. I guess I can live with either.

writing-guide.md
chrisvfritz: I added it to resources section.

writing-guide.md
chrisvfritz: Just clarified. Let me know what you think!

writing-guide.md
chrisvfritz: Yes. 😅

writing-guide.md
phanan: > That doesn't really tell me why their preference for lower-case is more important than the experience of our readers.

This is one of the rare cases when I would put the "experience" of our readers (which, I'd argue, shouldn't be affected that much) _lower_. To me, a name/branding is extremely important, and not for us to alter/modify at will.

phanan: I should have added that tt's also more consistent (there's only one source of truth) and less thinking for contributors to just use the name as given by the people who named it.

chrisvfritz: > I should have added that it's also more consistent

It sounds like we both value consistency - I just value consistency within a site/page/sentence for all projects rather than consistency with an individual project's site. In my case, the goal is to improve the experience of our readers, which you seem to agree is affected by this. In your case, what is the ultimate goal of the consistency you're striving for? What problem would we solve by following the branding conventions of an individual project over more global conventions?

> less thinking for contributors to just use the name as given by the people who named it

I haven't seen this to be the case in practice, as many (probably most) people don't know the branding conventions of every project. I think it's easier to remember one rule (the rule for our docs) rather than many rules - one for each project we mention. Either way though, if someone gets it wrong, it's something that's easy for us to update ourselves pre-merge.

writing-guide.md
chrisvfritz: I would argue that it's easier to draw the line between full words and abbreviations than between less and more familiar

I agree, but I also think most people are not thinking of symbols as abbreviations. Plus, if it's so universal that it's on every standard keyboard (that I've seen, at least) and even crosses language barriers, then I think it's OK. 😄 I just added a note to address this.

writing-guide.md
sdras: This looks ok to me, I know it's long, but that's also why we have the different bold and `li`, so people can scan

writing-guide.md
sdras: I think in previous PRs we have been doing what @phanan suggests- of honoring the way that the company/entity themselves call themselves. What I've seen is that otherwise they do come and PR to correct it. Some have trademarks etc where they require a particular spelling or casing. Would you be open to changing it, Chris?

chrisvfritz: Yeah, I don't mind if that's what the majority of the docs team prefers, though I think I'm still missing what the advantage would be. Even for entities with trademarks, like "npm", using the trademark as a noun rather than an adjective is already a violation (and one with greater risk of doing actual harm from my understanding, as that's what brought trademarks such as Kleenex and Xerox under threat). And I of course agree to updating in the case that we receive a cease and desist, but I still haven't heard any advantage to complying with mere requests. 🤷‍♂️

PR #2319, opened 2019-09-30
Fix: v-model example overflow problem

sercaneraslan:
Hi,

I fixed the v-model example overflow problem.

Before:

old

After:

new

URL: https://vuejs.org/v2/guide/


phanan:
Thanks, but I don't see a problem? If the text is too long there will be a horizontal scrollbar, right?

sercaneraslan:
Yes but I don't see a scrollbar at the Mac OS Chrome because it's the default behavior.

Anyway, thank you.

PR #2321, opened 2019-10-01
Fix accessibility issues relating to landmarks

heatherpekrul:
Using the Axe Coconut Chrome extension, I discovered several accessibility issues relating to landmarks on the vuejs.org website. Fixing these will make it easier for those with screen readers to use the web site.


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vuejs.org",
   "private": true,
   "hexo": {
-    "version": "3.8.0"
+    "version": "3.9.0"
```
phanan: Can you reverse the changes made to this file?

heatherpekrul: Sure - I've reverted this

package.json
```diff
@@ -29,4 +29,4 @@
     "hoek": "^6.1.2",
     "request": "^2.85.0"
   }
-}
+}
```
phanan: Please revert this change as well.

heatherpekrul: Done.

phanan:
Looks great, thanks a lot!

PR #2328, opened 2019-10-06
Update _config.yml

benjaminelliot:
Broken link for Patreon gold sponsor due to malformed URL.


phanan:
Oops, thanks for the fix!

PR #2330, opened 2019-10-07
Update partners.md

edwardscheindlin:
Updating Modus Create partner description to:

Modus Create is a digital product agency that supports clients with business and product strategy consulting, customer experience, cloud services, and Agile software delivery. Our official partnerships with Atlassian, AWS, GitHub, Ionic Framework, InVision, Cloudflare, and Vue.js reinforce our proven results with digital transformation with organizations from startups to the Fortune 100.


phanan:
As this is not adding but a (minor) modifying, I think we can go ahead and merge it without having to await for @yyx990803.

PR #2332, opened 2019-10-08
Update typescript.md

n4ks:
When creating a new project, he imported values ​​from vue-property-decorator, and not from vue-class-component


phanan:
Thank you, but isn't it clarified right above that?

> If you prefer a class-based API when declaring components, you can use the officially maintained [vue-class-component](https://github.com/vuejs/vue-class-component) decorator:

PR #2336, opened 2019-10-10
Update components-custom-events.md

schwarz-ben:
I am new to web development and it took me some time to understand what was all the fuss about here.


phanan:
> I do not think this change needs to be made.

Thanks for the effort @schwarz-ben, but I'd tend to agree with @PhilipJohnBasile.

PR #2339, opened 2019-10-13
fix avatar aspect ratio

imagentleman:
In the community partners section (https://vuejs.org/v2/guide/team.html#community-partners),  Pratik Patel's avatar image looks stretched out. 

Screen Shot 2019-10-13 at 12 43 32 AM

The image just happens to be "not a square" but instead it has a portrait aspect ratio (its height is bigger than its width). 

This pr fixes that issue by making the images do a cover resize.

Making the avatar look like this.

Screen Shot 2019-10-13 at 12 44 32 AM

### DIFF
--- a/themes/vue/source/css/_team.styl
+++ b/themes/vue/source/css/_team.styl
@@ -35,6 +35,7 @@
       flex: 0 0 80px
       img
         border-radius: 50%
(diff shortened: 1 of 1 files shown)


phanan:
This is great, thanks!

PR #2340, opened 2019-10-13
chore: add sponsor name for a11y

imagentleman:
The Standard Library sponsor on the sidebar is missing the name in the config, which leads to the image on the sidebar to be missing its alt property (https://github.com/vuejs/vuejs.org/blob/master/themes/vue/layout/partials/sponsors_sidebar.ejs#L7). Since the anchor is just that image without any text, having it would improve accessibility.


phanan:
👍 Thanks!

PR #2341, opened 2019-10-14
wording of scope explaination

swuecho:
I find the scope is very confusing and the doc is not that clear to me, so I changed some wording based on my understand. 

(for this concepts, I think nothing compared to a live doc. Is there any infrustructure to enable live doc on the doc site?

### DIFF
--- a/src/v2/guide/components-slots.md
+++ b/src/v2/guide/components-slots.md
@@ -58,27 +58,28 @@ If `<navigation-link>`'s template did **not** contain a `<slot>` element, any co
 When you want to use data inside a slot, such as in:
 
 ``` html
+<!-- parent component -->
 <navigation-link url="/profile">
   Logged in as {{ user.name }}
 </navigation-link>
 ```
 
-That slot has access to the same instance properties (i.e. the same "scope") as the rest of the template. The slot does **not** have access to `<navigation-link>`'s scope. For example, trying to access `url` would not work:
+That slot has access to the same instance properties (i.e. the parent "scope") as the rest of the template. The slot does **not** have access to `<navigation-link>`'s scope (i.e. the child "scope"). For example, trying to access `url` would not work:
 
 ``` html
 <navigation-link url="/profile">
   Clicking here will send you to: {{ url }}
   <!--
   The `url` will be undefined, because this content is passed
-  _to_ <navigation-link>, rather than defined _inside_ the
-  <navigation-link> component.
+  _to_ the child component <navigation-link> after `url` is interpolated in parent component,
+  rather than passed to <navigation-link> component directly and interpolated there.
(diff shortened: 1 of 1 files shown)


swuecho:
the section title is "compilation scope", "interpolation scope" is more accurate?

src/v2/guide/components-slots.md
```diff
@@ -58,27 +58,28 @@ If `<navigation-link>`'s template did **not** contain a `<slot>` element, any co
...
+That slot has access to the same instance properties (i.e. the parent "scope") as the rest of the template. The slot does **not** have access to `<navigation-link>`'s scope (i.e. the child "scope"). For example, trying to access `url` would not work:
 
 ``` html
 <navigation-link url="/profile">
   Clicking here will send you to: {{ url }}
   <!--
   The `url` will be undefined, because this content is passed
-  _to_ <navigation-link>, rather than defined _inside_ the
-  <navigation-link> component.
+  _to_ the child component <navigation-link> after `url` is interpolated in parent component,
```
NataliaTepluhina: **concern**: from my point of view, the term `interpolated` here add to the cognitive load and is not really descriptive, so I would prefer to leave this part as it was before

src/v2/guide/components-slots.md
```diff
@@ -58,27 +58,28 @@ If `<navigation-link>`'s template did **not** contain a `<slot>` element, any co
...
+  rather than passed to <navigation-link> component directly and interpolated there.
   -->
 </navigation-link>
 ```
 
 As a rule, remember that:
 
-> Everything in the parent template is compiled in parent scope; everything in the child template is compiled in the child scope.
+> Everything in the parent template is interpolated in parent scope; everything in the child template is interpolated in the child scope.
```
NataliaTepluhina: same as above: let's leave `compiled` here

src/v2/guide/components-slots.md
```diff
@@ -58,27 +58,28 @@ If `<navigation-link>`'s template did **not** contain a `<slot>` element, any co
...
 
 ``` html
+<!-- parent component -->
 <navigation-link url="/profile">
   Logged in as {{ user.name }}
 </navigation-link>
 ```
 
-That slot has access to the same instance properties (i.e. the same "scope") as the rest of the template. The slot does **not** have access to `<navigation-link>`'s scope. For example, trying to access `url` would not work:
+That slot has access to the same instance properties (i.e. the parent "scope") as the rest of the template. The slot does **not** have access to `<navigation-link>`'s scope (i.e. the child "scope"). For example, trying to access `url` would not work:
```
NataliaTepluhina: thank you for this change! I think it really makes an explanation more clear

PR #2342, opened 2019-10-14
more wording

swuecho:


phanan:
Thanks for the effort! The existing content reads better to me though, so I'm going to pass this for now.

PR #2344, opened 2019-10-14
Edit grammar prop validation

drewmarshburn:
Just giving subject/verb agreement.


NataliaTepluhina:
Thank you for the fix!

PR #2345, opened 2019-10-14
Correctly nest sub-menu for h3s inside h2 list item

michapietsch:
Observed behavior:
An ```ul``` with H3 links is nested inside the parent ```ul``` and as a sibling to the ```li``` containing the link to the H2.

Fix:
The sub-menu for H3s is now nested in the same way H2 sub-menu already are.

The former incorrect nesting also constitutes an a11y issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vuejs.org",
   "private": true,
   "hexo": {
-    "version": "3.8.0"
+    "version": "3.8.1"
   },
   "scripts": {
     "start": "hexo server",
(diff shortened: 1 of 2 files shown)


phanan:
Thanks for the effort! Can you guide us to an example of this change in action? A before/after comparison would be even better!

michapietsch:
Yeah, sure!

Before:

Bildschirmfoto 2019-10-15 um 11 11 08

For the third level heading "Relation to Custom Elements" the containing ```ul``` is a direct child of the parent ```ul```.

After:

Bildschirmfoto 2019-10-15 um 11 12 30

Now the ```li``` contains the ```a``` as well as the sub menu ```ul```. For sub-menus for second level headings like "Introduction" this is already handled correctly, as you can see at the top edge of the DevTools DOM tree visualization.

Hope that helps! Please let me know, if you have any more questions!

phanan:
Right, thanks for the fix!

PR #2348, opened 2019-10-16
Enforcing `rel="noopener"` for external _blank targets`

hiendv:
See [the MDN note](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target) for for information about the privacy and security concerns.

This PR is trying to add the `rel="noopener"` in the following locations:
- Algolia Attribution - https://vuejs.org/v2/search/
- Resources dropdown

I also notice `rel="noopener noreferrer"` (as a workaround for old browsers) and `rel="noopener"` are used at the same time across the codebase. Should we make them more consistent?


hiendv:
@phanan I would love to. So `rel="noopener noreferrer"` all the way right? :D

PR #2350, opened 2019-10-16
Description wasn't clear on the purpose of slots

under445:
The description on the purpose of Slots wasn't so clear at the beginning of the "Slots" section. I changed it a bit, with some examples, to explain the benefits of slots. How we can use them instead of using attributes or props for passing HTML content to other component.


phanan:
Thanks for the effort, but I'd rather keep the current version which (IMO) is more precise and to the point.

PR #2351, opened 2019-10-17
Mention setting nested object properties

opl-:
Explains that one can pass a reference to a nested object to `Vue.set`. This answers an actual question I was asked while teaching Vue.


NataliaTepluhina:
Thank you for working on this PR! However, I think this change creates an additional cognitive load and it's not necessary.

PR #2357, opened 2019-10-22
Update list.md

Aubynj:
This is not an error, but a suggestion concerning to adding empty values in fields. Thank You


phanan:
Thanks. A better (IMO) alternative would be to add a `required` attribute into the input tag instead. What do you think?

Aubynj:
Yea that make sense too. That is the best answer to my suggestion. Thanks @phanan

PR #2360, opened 2019-10-22
Update components-slots.md with 

rabbitwerks:
Added example for improper syntax, clarification on workarounds for dynamic v-slot, including examples.


phanan:
Let's focus on the other PR.

PR #2361, opened 2019-10-22
Update conditional.md

rabbitwerks:
updated the `v-if` and `v-for` caveat wording, removed p tag wrapper and a tag link, replaced with md link to list ref


phanan:
Thanks for the PR, but I don't think these changes are necessary.

rabbitwerks:
the docs dont display properly...?

PR #2365, opened 2019-10-25
use production npm install for docker examples

mike-moede:
recommend using npm production install to minimize docker image sizes

### DIFF
--- a/src/v2/cookbook/dockerize-vuejs-app.md
+++ b/src/v2/cookbook/dockerize-vuejs-app.md
@@ -22,8 +22,8 @@ WORKDIR /app
 # copy both 'package.json' and 'package-lock.json' (if available)
 COPY package*.json ./
 
-# install project dependencies
-RUN npm install
+# install project dependencies leaving out dev dependencies
+RUN npm install --production
 
 # copy project files and folders to the current working directory (i.e. 'app' folder)
 COPY . .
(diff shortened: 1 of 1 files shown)


src/v2/cookbook/dockerize-vuejs-app.md
```diff
@@ -66,7 +66,7 @@ Let's refactor our `Dockerfile` to use NGINX:
 FROM node:lts-alpine as build-stage
 WORKDIR /app
 COPY package*.json ./
-RUN npm install
+RUN npm install --production
 COPY . .
 RUN npm run build
```
machristie: How does `npm run build` work if only the production dependencies have been installed?

dominikstraessle: As of my own experience, a normal `npm run build` fails, if only the production dependencies are installed. The build command depends on the dev dependencies and therefore they have to be installed. I think it would be more clear to remove the `--production` option. Inexperienced users may be confused about the build errors.

mike-moede: Yeah, I agree.  I messed up.  I can create the pull request to get it back.

mike-moede: https://github.com/vuejs/vuejs.org/pull/2392

PR #2369, opened 2019-10-31
Add tip for Using sync with arrays

subssn21:
Ran into this issue this morning. Luckily someone wrote a blog post about it, but I feel that it should be in the documentation so that people don't have to go searching.
https://laternastudio.com/blog/using-sync-inside-a-v-for-loop-with-vuejs/

### DIFF
--- a/src/v2/guide/components-custom-events.md
+++ b/src/v2/guide/components-custom-events.md
@@ -169,3 +169,11 @@ The `.sync` modifier can also be used with `v-bind` when using an object to set 
 This passes each property in the `doc` object (e.g. `title`) as an individual prop, then adds `v-on` update listeners for each one.
 
 <p class="tip">Using <code>v-bind.sync</code> with a literal object, such as in <code>v-bind.sync="{ title: doc.title }"</code>, will not work, because there are too many edge cases to consider in parsing a complex expression like this.</p>
+
+<p class="tip">When using <code>v-bind.sync</code> with a <code>v-for</code> You must use the full array with the index and not the item being iterated over.
+
+```html
+<div v-for="(item,index) in myArray">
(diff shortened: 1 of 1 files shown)


src/v2/guide/components-custom-events.md
```diff
@@ -169,3 +169,11 @@ The `.sync` modifier can also be used with `v-bind` when using an object to set
 This passes each property in the `doc` object (e.g. `title`) as an individual prop, then adds `v-on` update listeners for each one.
 
 <p class="tip">Using <code>v-bind.sync</code> with a literal object, such as in <code>v-bind.sync="{ title: doc.title }"</code>, will not work, because there are too many edge cases to consider in parsing a complex expression like this.</p>
+
+<p class="tip">When using <code>v-bind.sync</code> with a <code>v-for</code> You must use the full array with the index and not the item being iterated over.
```
NataliaTepluhina:
```suggestion
<p class="tip">When using <code>v-bind.sync</code> with a <code>v-for</code>, you must use the full array with the index and not the item being iterated over.
```

src/v2/guide/components-custom-events.md
```diff
@@ -169,3 +169,11 @@ The `.sync` modifier can also be used with `v-bind` when using an object to set
...
+
+```html
+<div v-for="(item,index) in myArray">
```
NataliaTepluhina:
```suggestion
<div v-for="(item, index) in myArray">
```

sdras:
Thank you for the PR! I'm going to go ahead and reject this even though we agree with you, because we're dropping `.sync` soon, and adding this in will add to our cleanup when we upgrade.

PR #2371, opened 2019-11-01
Update components-props.md

ishpartko:
Is the function 'includes' the better way to check the containing smth in array? No?


NataliaTepluhina:
While you're technically correct, the documentation of Vue 2 is using ES5 standard and `includes` is ES7.

PR #2375, opened 2019-11-06
Add MDBootstrap themes

mdembna:
Add the list of MDBootstrap templates & plugins to theme-data.js


themes/vue/source/js/theme-data.js
```diff
@@ -1,4 +1,88 @@
 var themeData = [
+  {
```
phanan: Please move this block to the end.

PR #2378, opened 2019-11-06
Fix link to pass query correctly

JohnTitor:
`query` doesn't pass the query, it should be `q`.


phanan:
Indeed! Thanks for the fix.

PR #2385, opened 2019-11-10
Add video to Markdown example

jeffreybiles:
This commit adds a tutorial video to the Markdown example.  Styling is based on other video links within the docs.

The video uses the current example as a jumping off point and adds two common use cases: SSR compatibility and syntax highlighting.

Hopefully readers will find this useful as-is, but if there's anything you'd like me to change about the video, please let me know and I can do a "docs edit" based on those recommendations.

### DIFF
--- a/src/v2/examples/index.md
+++ b/src/v2/examples/index.md
@@ -6,4 +6,6 @@ order: 0
 
 > Dead simple Markdown editor.
 
+<div class="vue-screencasts"><a href="https://youtu.be/THT5FefYZCo" target="_blank" rel="noopener" title="Markdown and Syntax Highlighting Tutorial"> Watch a free video on Vue Screencasts.  It adds syntax highlighting and SSR compatibility.</a></div>
+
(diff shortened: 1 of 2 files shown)


phanan:
Thank you, but I don't think this is necessary.

jeffreybiles:
Thanks for your feedback @phanan .

To help me contribute better in the future, and thus save us both time, could you help me pinpoint what the issue was - and therefore, what I can do better next time?

I can see five possible issues:

A) Markdown is so simple that users will not benefit from a video
B) The video expanded upon what was in the example, and you'd prefer it stuck to what's already in the text
C) There's something about my presentational style that needs to be modified before it's acceptable in the docs
D) The PR was unexpected, and given the relative low priority of this page it was not worth your time to watch the video and make a specific judgement
E) The video is hosted on YouTube, which is not available in all countries

Given your phrasing of calling it "not necessary", I'd assume it's either A or D, but I'd like to confirm - but if it was just you being polite, and the issue is something else, then that will be good to know.

Thanks so much, and I hope to be able to contribute better in the future!

PR #2389, opened 2019-11-16
Drop IE 3 compatibility

MicroDroid:


phanan:
Thanks, but Vue docs intentionally support ES5.

PR #2390, opened 2019-11-17
Add braces to make code more clear

ceynri:
According to the general coding standards, if statements are best to add curly braces to avoid misunderstanding, and it can make code more clear.


phanan:
"General coding standards" are pretty subjective but yeah, personally I prefer wrapping if statements in braces too :) Thanks.

PR #2394, opened 2019-11-20
Add link from docs to v-html directive

tomquirk:
In the **Raw HTML** section of docs, link the first use of the term `v-html` to it's entry in the API docs. 

This keeps it consistent with the description of `v-bind` that follows, and improves efficiency for user when reading this page.


phanan:
Thanks!

PR #2395, opened 2019-11-20
Changed order of last section to one up

BatlethTech:
The `Non-Prop Attributes` section talks about attributes flowing by default to root element, and `Disabling Attribute Inheritance` about **not** making them flow to root so it makes sense to the latter right after former instead of being separated by `Replacing/Merging with Existing Attributes` section which is not so closely related.


phanan:
Thanks for the effort! The way I see it, the current next section still talks about passing attributes to the root component element (only this time with replacing/merging attributes) so I'd say the current flow reads better.

PR #2396, opened 2019-11-20
Use includes() rather than indexOf() !== -1

eran-nussbaum:


phanan:
Thanks, but the doc uses ES5.

eran-nussbaum:
😱 sorry

PR #2397, opened 2019-11-20
update computed.md

mohamedziada:
Line 82 
### Computed Caching vs Methods
To be 
### Computed Properties vs Methods

for sure it is cached properties but it drive confuse to name it "Computed Caching" plus you explain it down 

**computed properties are cached based on their reactive dependencies.**


NataliaTepluhina:
Thank you for your opinion! However, I think the current title is fine

PR #2401, opened 2019-11-24
fixing broken link in writing guide

one-angry-penguin:
Not entirely sure if this was the link meant, but the content and title match.


phanan:
Indeed, thanks for fixing!

PR #2402, opened 2019-11-25
Document gotcha with :is and DOM props

fer22f:
This is #2241 reborn. The motivation for it is that this snippet of code:

```html
<component :is="multiline ? 'textarea' : 'input'" v-model="value">
```

Won't work as would expect, so we are documenting that you should, instead, use `:value.prop` and `@input` instead of `v-model`.


phanan:
Pinging @sdras as she's been involved with the other from day 1.

sdras:
perfect, thank you for resubmitting!

fer22f:
No problem!

Issue #2405, opened 2019-11-29
Style guide [element-attribute-order]: missing v-slot attribute


vvanpo:
Category 6 lists the legacy `slot` attribute, but not `v-slot`.

PR #2406, opened 2019-12-03
Removed slot attribute from styleguide attributes order

NataliaTepluhina:
This PR removes deprecated `slot` attribute from Styleguide section about recommended attributes order. I didn't add `v-slot` directive there but I am open for opinions on where it should be added in the order

Close https://github.com/vuejs/vuejs.org/issues/2405

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1612,7 +1612,6 @@ This is the default order we recommend for component options. They're split into
 6. **Unique Attributes** (attributes that require unique values)
   - `ref`
   - `key`
-  - `slot`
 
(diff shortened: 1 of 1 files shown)


phanan:
LGTM!

Justineo:
> I didn't add `v-slot` directive there but I am open for opinions on where it should be added in the order

What about putting it right after `is`?

vvanpo:
> What about putting it right after `is`?

I don't think it belongs in the "definition" category, since it doesn't define what gets rendered, only where something gets rendered.

Or are you suggesting creating another category directly below "definition"? I'm not sure it really fits in any of the existing ones, so that might be the best solution.

PR #2410, opened 2019-12-05
Update form-validation.md

### LINKED ISSUE #2391: [cookbook] Form validation throws CORS error for a codepen sample
Hi,

when I read this article: https://vuejs.org/v2/cookbook/form-validation.html#Server-side-Validation I've noticed that the linked sample is not working properly.
In the section "Server-side Validation" the codepen doesn't make a server-side validation due to CORS error. I followed also a link to API and try to make a direct call to it from my browsers and it looks like server is not available any longer.
Screenshot 2019-11-17 at 10 49 51

cfjedimaster:
Updates the server side code. Fixes #2391

### DIFF
--- a/src/v2/cookbook/form-validation.md
+++ b/src/v2/cookbook/form-validation.md
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
 
 ## Server-side Validation
 
-In my final example, we built something that makes use of Ajax to validate at the server. The form will ask you to name a new product and will then check to ensure that the name is unique. We wrote a quick [OpenWhisk](http://openwhisk.apache.org/) serverless action to do the validation. While it isn't terribly important, here is the logic:
+In my final example, we built something that makes use of Ajax to validate at the server. The form will ask you to name a new product and will then check to ensure that the name is unique. We wrote a quick [Netlify](https://netlify.com/) serverless action to do the validation. While it isn't terribly important, here is the logic:
 
 ``` js
-function main(args) {
-    return new Promise((resolve, reject) => {
-        // bad product names: vista, empire, mbp
-        const badNames = ['vista', 'empire', 'mbp'];
+exports.handler = async (event, context) => {
+  
+    const badNames = ['vista', 'empire', 'mbp'];
+    let name = event.queryStringParameters.name;
 
-        if (badNames.includes(args.name)) {
-          reject({error: 'Existing product'});
-        }
+    if (badNames.includes(name)) {
+      return { 
+        statusCode: 200,         
+        headers:{
+          'Access-Control-Allow-Origin':'*'
+        },
+        body: JSON.stringify({error:'Invalid name passed.'}) 
+      }
+    } else {
+      return {
+        statusCode: 200,
+        headers:{
+          'Access-Control-Allow-Origin':'*'
+        },
+        body: JSON.stringify({ status: 'ok' })
+      }
+    }
 
-        resolve({status: 'ok'});
-    });
 }
+
 ```
 
 Basically any name but "vista", "empire", and "mbp" are acceptable. Ok, so let's look at the form.


src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
-        // bad product names: vista, empire, mbp
-        const badNames = ['vista', 'empire', 'mbp'];
-
-        if (badNames.includes(args.name)) {
-          reject({error: 'Existing product'});
-        }
+exports.handler = async (event, context) => {
+  
+    const badNames = ['vista', 'empire', 'mbp'];
+    let name = event.queryStringParameters.name;
```
phanan: Since this is never re-assigned, `const` would be a better fit.

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
+
+    if (badNames.includes(name)) {
+      return {
```
phanan: The response code shouldn't be 200. A 4xx error (e.g. 400 Bad Request or 422 Unprocessable Entity) fits better IMO.

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
-        }
+exports.handler = async (event, context) => {
+  
+    const badNames = ['vista', 'empire', 'mbp'];
+    let name = event.queryStringParameters.name;
+
+    if (badNames.includes(name)) {
+      return { 
+        statusCode: 200,         
+        headers:{
```
phanan: Can you please lint this code snippet? There are several linting errors.

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
+        },
+        body: JSON.stringify({error:'Invalid name passed.'}) 
+      }
+    } else {
+      return {
+        statusCode: 200,
+        headers:{
+          'Access-Control-Allow-Origin':'*'
+        },
+        body: JSON.stringify({ status: 'ok' })
```
phanan: Do we really need a body which basically says what the header does? Also if we remove the body, the status code might as well be 204 No Content.

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
 ``` js
-function main(args) {
-    return new Promise((resolve, reject) => {
-        // bad product names: vista, empire, mbp
-        const badNames = ['vista', 'empire', 'mbp'];
-
-        if (badNames.includes(args.name)) {
-          reject({error: 'Existing product'});
-        }
+exports.handler = async (event, context) => {
```
phanan: Since there's no `await` within the function, `async` is redundant I guess.

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,34 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
+  
+    const badNames = ['vista', 'empire', 'mbp'];
+    let name = event.queryStringParameters.name;
+
+    if (badNames.includes(name)) {
+      return { 
+        statusCode: 200,         
+        headers:{
+          'Access-Control-Allow-Origin':'*'
```
phanan: Since this function is supposed to have the same origin as the code in our context, this shouldn't be necessary. Same goes with the "success" logic.

src/v2/cookbook/form-validation.md
cfjedimaster: Considering this is very much *not* the focal point of the cookbook entry, I don't think is really necessary. In fact, I'd rather folks not even look at the server-side code as the focus is on the Vue aspect.

src/v2/cookbook/form-validation.md
cfjedimaster: Again, while I agree with you, I do not think it's necessary to worry so much about this part.

src/v2/cookbook/form-validation.md
cfjedimaster: But then it can't be run in the CodePen.

src/v2/cookbook/form-validation.md
phanan: It doesn't need to be the focal point. If we have code in our official cookbook, it better looks good.

src/v2/cookbook/form-validation.md
phanan: I respectfully disagree, for the same reason above.

src/v2/cookbook/form-validation.md
phanan: The CodePen version can keep the header but for a normal context (hence "in our context" in my original comment) the header is redundant.

src/v2/cookbook/form-validation.md
cfjedimaster: Sure, but again, this is *not* the point of the article. We don't know what backend folks are using. Heck, if I had my way I'd hide it completely and in fact, the more that I think about it the more I'd like to do that. @sdras - do you have an opinion on this? While I agree with some of the comments about the code here, I do not agree with all, and it feels like a waste of time to get deep into it. Could we just simplify things and *not* show the back end code?

phanan: I don't think you got what I mean. I do agree that this code snippet isn't the point of the article, and sure, if we can hide it I won't bother. But we're showing it here, and if it's shown, it should be in its best shape.

cfjedimaster: Hmm ok maybe. I'd still like to see what @sdras thinks about just skipping the code completely.

cfjedimaster:
I've updated the commit (hopefully correctly) to remove the server side code.

src/v2/cookbook/form-validation.md
sdras: Hi @cfjedimaster, please respond to @phanan's comments here. It's not the right path to just delete parts of the examples like this. An spent the time to review, and his suggestions are great. The point of the cookbook is an extension past Vue to use cases, it's not necessarily only Vue code we're aiming for here. 

Can you please go through the PR and work through An's suggestions? Thanks.

cfjedimaster: Ok, I wasn't trying to disregard his comments, but I honestly thought it would simplify things to focus on the client-side code. I'll bring it back (later today anyway).

cfjedimaster:
I believe I hit everything you said, *except* removing the async handler from the server-side code as Netlify didn't seem to like that (I would get CORS errors despite it still being in the "real" code).

cfjedimaster:
Just checking - are my updates ok?

bencodezen:
> Just checking - are my updates ok?

Hey @cfjedimaster! Just wanted to let you know that @phanan is out for a few days; but I'll check in with him when he gets back regarding this PR.

bencodezen:
@cfjedimaster In the meantime, could you run your code snippets through a linter / Prettier to keep it consistent in the docs? Happy to help format it for you if you don't have time.

In the future, we'll probably plan on having auto-linting / formatting config in the repo to help reduce these sort of stylistic PR changes; so we appreciate your patience with this in the meantime!

cfjedimaster:
I'd be happy with yall formatting it as you see fit. :) (Yes, being a bit lazy.)

cfjedimaster:
Just checking in. If you tell me the linting/prettier settings you want I'll do it. This PR is now over two months old and I'd like to get it wrapped. ;)

cfjedimaster:
Anything I can do to get this shipped?

src/v2/cookbook/form-validation.md
```diff
@@ -406,16 +412,15 @@ const app = new Vue({
...
-          } else {
-            // redirect to a new URL, or do something on success
-            alert('ok!');
+        fetch(apiUrl+encodeURIComponent(this.name))
+        .then(async res => {
+          if(res.status === 204) {
+            alert('Ok!')
+          } else if(res.status === 400) {
+            let errorResponse = await res.json();
+            this.errors.push(errorResponse.error);
```
phanan:
```suggestion
        fetch(apiUrl + encodeURIComponent(this.name))
        .then(async res => {
          if (res.status === 204) {
            alert('OK')
          } else if (res.status === 400) {
            let errorResponse = await res.json();
            this.errors.push(errorResponse.error);
```

src/v2/cookbook/form-validation.md
```diff
@@ -406,16 +412,15 @@ const app = new Vue({
...
           }
-        });
+        });
```
phanan:
```suggestion
        });
```

src/v2/cookbook/form-validation.md
```diff
@@ -333,21 +333,27 @@ We set up the total value as a computed value, and outside of that bug I ran int
...
+    if (badNames.includes(name)) {
+      return { 
+        statusCode: 400,         
+        body: JSON.stringify({error:'Invalid name passed.'}) 
+      }
+    } else {
+      return {
+        statusCode: 204
+      }
+    }
```
phanan:
```suggestion
        body: JSON.stringify({error: 'Invalid name passed.'}) 
      }
    }

    return {
      statusCode: 204
    }
```

src/v2/cookbook/form-validation.md
```diff
@@ -407,13 +413,12 @@ const app = new Vue({
...
-        .then(res => res.json())
-        .then(res => {
-          if (res.error) {
-            this.errors.push(res.error);
-          } else {
-            // redirect to a new URL, or do something on success
-            alert('ok!');
+        .then(async res => {
+          if (res.status === 204) {
+            alert('OK')
```
phanan:
```suggestion
            alert('OK');
```

phanan:
Hey Raymond! Terribly sorry, I totally missed this after getting back from my vacation. I've taken the liberty to ~~butcher~~ apply some linting to your PR and it should be fine to merge now. Thanks for the effort, and again, apologies for being super late on this.

cfjedimaster:
No worries and thank you for pushing this out!

PR #2413, opened 2019-12-09
Update components.md

RC-787:
Updating prop example to include all fields referenced in previous example


phanan:
Thanks for the effort! However, this is 1. IMO unnecessary for our purpose here and 2. incorrect, as `post.comments` would normally be an array and can't be used directly with `v-html` this way.

PR #2414, opened 2019-12-10
Update example of a singe file component

humb1t:
In a https://vuejs.org/v2/guide/single-file-components.html we have `Hello.vue` example, which contains `module.exports =` semantic. I hope I'm right and it's better to use `export default` semantic instead. So I replaced gist with fork of an initial example with this small change.


humb1t:
Why I think it's better to use, because `module.exports =` leads to `TypeError: "exports" is read-only`.

phanan:
Thanks, but I think the current gist is fine.

PR #2425, opened 2019-12-31
events.md: Use v-on in all examples for consistency.

dhirshjr:
I noticed that every example on this page uses v-on, but toward the bottom at the System Modifier Keys section, @ starts getting used without explanation. This can confuse someone that is just starting out.


phanan:
Thanks!

PR #2431, opened 2020-01-02
fix(docs): typo

farzadso:
This merge request fixes a little typo in the directive docs.

I doubt this mistake being intentional.


phanan:
Thanks, but it's not a typo :) A color swatch is another name for a color palette.

Issue #2429, opened 2020-01-02
Inaccuracy in comparisons guide


samuelpath:
I think that issue #668 (Add comparison to Svelte) could be reopened: https://github.com/vuejs/vuejs.org/issues/668

On 10 Jul 2017 @chrisvfritz said: "Closing for now, as Svelte might still be a little too niche to spend time on a comparison"

I believe that this is not the case anymore, given that it ranks #1 in front-end frameworks interest in the 2019 state of JS survey: https://2019.stateofjs.com/front-end-frameworks/

phanan:
I agree. We may want to wait a bit for Vue 3 though, as it will come with features and improvements that are more up to date in comparison with Svelte (as well as other frameworks).

Charuru:
Really want to see this. Just selfishly, can somebody get started on it with Vue3? I'd really like to know.

PR #2433, opened 2020-01-03
Adding placeholder Svelte section in comparison

samuelpath:
See discussion in issue #2429


NataliaTepluhina:
Thanks for your contribution! While I agree with @phanan we should have the comparison with Svelte in Vue 3 docs, a placeholder is not really necessary in Vue 2 docs.

PR #2436, opened 2020-01-05
fixed spanish language typo in dropdown menu

Yilber:
español > Español


Yilber:
@NataliaTepluhina You're welcome!

phanan:
Hmm? AFAIK you don't capitalize language names in Spanish. For example, this is from Google's setting page:

image

Yilber:
The rule says that language's name belongs to the common names category, therefore, you should not capitalize ingles, frances etc. The problem is that the rule does not tell you what to do when you are listing those names. The common approach is to capitalize the first letter when you are making a list.

You can also find the following on google too if you go to [language setting on the account's page](https://myaccount.google.com/language)

image

Yilber:
@phanan I got in touch with the RAE. I will keep you updated once they provide any feedback on this subject.

Yilber:
@phanan, @NataliaTepluhina here is the response from the RAE about the language names in dropdown menus. https://twitter.com/RAEinforma/status/1217013057118986240

It seems that Google made a small mistake.

phanan:
TIL, thanks @Yilber!

PR #2438, opened 2020-01-07
Update class-and-style.md, typos

jbratcher:
Missing single-quotes around conditional class name 'active'.  This may confuse some that copy/paste or don't notice the first condition without the quotes around the class name.


phanan:
Thanks, but this is not a typo. `v-bind:class` can take a literal object, whose keys can be without quotes.

PR #2439, opened 2020-01-07
Change String.match method to regexObj.test

jonex2:
String.match returns an array with detail information about the match and then it must be checked for truthiness. But here you just want to know if the value exists in the string. The regexObj.test method returns a boolean and performs better. 
RegexObj.test is the correct method in this case.

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -252,7 +252,7 @@ Mutation methods, as the name suggests, mutate the original array they are calle
 
 ``` js
 example1.items = example1.items.filter(function (item) {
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the effort. `String.match` returns `null` if there's no match though, so it's usable in this case.

jonex2:
Of course string.match is usable, but it's not so proper and performs not so well as regex.test (s. https://jsperf.com/test-vs-match-regex). 
I think it would be good to have the best method!
Some developers rely that this docs show good javascript code and copy it. In this case they shouldn't.

PR #2441, opened 2020-01-11
Changed an example for nested loop in List Rendering

NataliaTepluhina:
The [Displaying Filtered/Sorted Results](https://vuejs.org/v2/guide/list.html#Displaying-Filtered-Sorted-Results) mentions a case with nested `v-for` loop but it's followed with an example that does not contain any nesting; more to say, `numbers` property is explicitly placed in `data`. This can raise a question: _"Is there any sense of using a method here? If `numbers` are in `data`, we can use computed again!"_

This PR adds an actual nested loop to align the description with the example

### DIFF
--- a/src/v2/guide/list.md
+++ b/src/v2/guide/list.md
@@ -383,19 +383,19 @@ computed: {
 
 In situations where computed properties are not feasible (e.g. inside nested `v-for` loops), you can use a method:
 
-``` html
-<li v-for="n in even(numbers)">{{ n }}</li>
+```html
+<ul v-for="set in sets">
+  <li v-for="n in even(set)">{{ n }}</li>
+</ul>
 ```
 
-``` js
+```js
 data: {
-  numbers: [ 1, 2, 3, 4, 5 ]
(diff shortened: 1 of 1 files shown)


src/v2/guide/list.md
```diff
@@ -383,19 +383,19 @@ computed: {
...
-  numbers: [ 1, 2, 3, 4, 5 ]
+  sets: [[ 1, 2, 3, 4, 5 ], [6, 7, 8, 9, 10]]
 },
 methods: {
-  even: function (numbers) {
-    return numbers.filter(function (number) {
-      return number % 2 === 0
-    })
+  even(numbers) {
+    return numbers.filter(number => number % 2 === 0)
```
phanan: We use ES5 for our docs.

NataliaTepluhina: Thanks, was working on the same example for Vue 3 docs and forgot to 'translate' back to ES5 for Vue2 docs 👍 

Should be fixed now

phanan:
Sweet, thanks for your work on this!

PR #2443, opened 2020-01-13
Fix docs: change active to isActive

sajibcse68:


phanan:
Thanks for the effort, but `active` is correct here.

PR #2445, opened 2020-01-14
Update components.md

tolbxela:
The example of "Using v-model on Components" was a little unclear for me. As a newcommer at Vue I couldn't understand what is "searchText" and where it comes. I had to investigate it in jsfiddle and it took some time to clarify. 
We can make this example much easier to understand with this small addition.

```
new Vue({
    el: '#app',
    data: {
        searchText: "text"
    }
})
```

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -463,6 +463,13 @@ Vue.component('custom-input', {
     >
   `
 })
+
+new Vue({
+    el: '#app',
+    data: {
+        searchText: "text"
+    }
+})
 ```
 
 Now `v-model` should work perfectly with this component:


tolbxela:
How about creating a fiddle for every example, to make it easier to try?

phanan:
I'd say if one follows the guide, the added code like you're suggesting (creating a new Vue instance on an element with `data` and so on) should come natural by now, so I'd prefer leaving it as is. Thanks for the suggestion though!

tolbxela:
In the current version the connection between "searchText" and "value"  of the component is not so obvious and could be confusing for newcommers like me. I did read the docs statring from the beginning and it still sometimes not so easy for me to understand the examples, since they are mostly not verbose enough.  
Verbosity in the documentation is a good thing!
If you don't like to put more code in the example, then add a small explanation about, what does searchText here mean.

tolbxela:
> How about creating a fiddle for every example, to make it easier to try?

What about this?

phanan:
`v-model` is explained in details in the ["Form Input Binding" page](https://vuejs.org/v2/guide/forms.html), which comes right before "Component Basis," so again, I'd insist that if someone follows the guide he/she shouldn't have a problem infer what `v-mode="searchText"` means. 

When verbosity is often a good thing, it can be too much and distract readers from the main topic. In this case, we want to focus on "Using `v-model` on Components" and not `v-model` itself.

tolbxela:
> I'd insist that if someone follows the guide he/she shouldn't have a problem infer what `v-model="searchText"` means.

I did follow the guide and I had a problem to understand it. So, please stop saying to me I didn't do it. 

My opinion remains:
The documentation for vue.js is mostly very good, but I still struggle sometimes to transform the presented parts of code to working solutions. 
This could, of course, mean that I am just not clever enough to understand it quickly. Sure. But the documentation could also do a better work to support me in the learning process. 

As I developer learnig vue.js, I would love to have more examples and explanations how to use the framework in the docs. Right now, I have to google a lot to find examples of real-life using of the vue.js concepts.

bencodezen:
@tolbxela Thanks for expressing your concerns. We are currently looking at changing the way we provide examples inside docs so users can play with live code in addition to following the text. 

One of the examples we have found to be very useful is how Vuetify documents their components: https://vuetifyjs.com/en/components/cards#outlined-cards

Would this format have provided the support you were looking for?

tolbxela:
> Would this format have provided the support you were looking for?

For sure! This is exactly what I often miss in the vuejs docs. 

The Vuetify has beautiful docs and they did very good job with all this codepen examples. 
But their documentation has also a big drawback. It looks pretty cool but is hard to read. And there is no plain source code at all! You have to open codepen every time to check the code. This could be problematic on the mobile devices and is useless for the print or e-book readers. 

I think it would be enough to create gists and codepen/fiddle examples and link them in the current docs. It is already done in some places of vue documentation and it is really great!

bencodezen:
@tolbxela That's great feedback. We appreciate the time you took to open the issue and let us know how we can better support new developers in the Vue community. If you have any other questions or suggestions in the meantime, please don't hesitate to reach out!

lbogdan:
@tolbxela You can see the code for the example's template (and script, on some examples) if you click on the `<>` button:

image

tolbxela:
@lbogdan yes, I know it and I don't like it. It's very bad for print  and does not look good on the mobile phone. 

vuetifyjs com_en_components_cards(iPhone 5_SE)

The source code is too big and placed in a frame. You have to scroll it sometimes in two directions and it sucks. 
The Vue docs are ok in both cases. You can print it well and on the mobile phone is looks ok. 

vuejs org_v2_guide_components-slots html(iPhone 5_SE)

tolbxela:
@bencodezen You are welcome! I am very happy to contribute a little.
Thanks to all Vue makers for a great product! 👍

PR #2447, opened 2020-01-16
Added simplified Object unification

hitautodestruct:
Added a simplified creating a fresh object using ES6 object spread operator to merge props into an existing object.


phanan:
Thanks but we consciously don't use ES6 for the docs.

hitautodestruct:
@phanan you have an async await example on the same page...

PR #2448, opened 2020-01-19
Fix base component regex

kootoopas:
So that the matching filename starts with `Base`.

### DIFF
--- a/src/v2/guide/components-registration.md
+++ b/src/v2/guide/components-registration.md
@@ -199,7 +199,7 @@ const requireComponent = require.context(
   // Whether or not to look in subfolders
   false,
   // The regular expression used to match base component filenames
-  /Base[A-Z]\w+\.(vue|js)$/
+  /^Base[A-Z]\w+\.(vue|js)$/
 )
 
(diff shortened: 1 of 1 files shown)


hzchirs:
Webpack `require.context` regex matches relative path,
for example: `['./BaseXXX.vue', './BaseYYY.vue']`
use `/^Base/` will always get empty result.

NataliaTepluhina:
@hzchirs is right - the current example is correct.

kootoopas:
I see, my bad there. 

However, the current pattern accepts names that _**contain**_ `Base`:
```js
console.log(/Base[A-Z]\w+\.(vue|js)$/.test('./AABaseCCC.vue')) // This matches, although it shouldn't.
```

To assert that the last part of the path **_starts with_** `Base`, a slash before the prefix would suffice:
```js
console.log(/\/Base[A-Z]\w+\.(vue|js)$/.test('./AABaseCCC.vue')) // No match
console.log(/\/Base[A-Z]\w+\.(vue|js)$/.test('./BaseCCC.vue')) // Matches
```

PR #2452, opened 2020-01-22
Moving external example code to the examples folder

garethx:
@NataliaTepluhina Per my DM, this PR begins to implement the idea of transitioning external code examples to the examples folder to reduce the bus-factor and improve maintainability. These then use CodeSandbox's 'Import from GitHub' feature to load the examples, so people can experiment with the code and running version of the app, while the underlying code for them is in a place anyone can access and edit.

One question about using the examples folder and https://vuejs.org/v2/examples/ - do you know if having folders inside the examples folder will cause issues with the menu of examples on the examples page?

### DIFF
--- a/src/v2/examples/vue-20-accessing-parent-component-instance/index.html
+++ b/src/v2/examples/vue-20-accessing-parent-component-instance/index.html
@@ -1,0 +1,91 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Dependency Injection Google Maps Demo</title>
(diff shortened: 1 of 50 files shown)


garethx:
Hmm, https://deploy-preview-2452--vuejs.netlify.com/v2/examples/ looks ok at least.

NataliaTepluhina:
@garethx no, we have Hexo where menu is not generated automatically so it should be fine, I will check manually just in case

NataliaTepluhina:
@garethx sorry for the delay, just finished a review! Thank you so much for your work, I will go ahead and merge it to master to test if sandboxes are generated properly

PR #2457, opened 2020-01-30
Moving external example code to the examples folder

garethx:
Continuing https://github.com/vuejs/vuejs.org/pull/2452, with this PR the code for all remaining external code examples in the guide will now be in the examples folder. The benefit is reducing the bus-factor and improve maintainability by making the code accessible to update by all.


NataliaTepluhina:
Thank you for this huge improvement @garethx! The review might take a few days due to a big diff size so please don't mind a slight delay with merging 😅

garethx:
@NataliaTepluhina The template error was caused by some zero width characters. Resolving that lead to another issue with Nunjucks in Hexo erroring on use of `{{` and `}}`. I've added an exclude for those files as I can't use the raw workaround as it's actual source code that gets run, and Nunjucks doesn't need to worry about them anyway as it doesn't need to render them.

NataliaTepluhina:
Thanks for the fix! I think we can safely merge now

PR #2459, opened 2020-02-01
Add(Special Attributes: ref) note for computed props

darrelfrancis:
On Vue forum someone complained that it was not clear in the documentation that refs are not accessible at created time, or in computed properties when evaluated at created time.

https://forum.vuejs.org/t/getting-the-width-an-height-of-an-element-in-a-component/15280/5

I have given a brief note and a workaround for the case of computed props.


darrelfrancis:
Oops sorry I realise my workaround won't work! 
Doh.

PR #2460, opened 2020-02-02
Fix wrong example code on Components page

zzulu:
v-model input event have to assign the value not the event.


zzulu:
Oh, sorry. I have a mistake. It's not wrong code. Close this pr.

PR #2464, opened 2020-02-04
Fix `arabic translation repo link` & add `Secondary Maintainers` section

atefBB:


atefBB:
Hey ! I don't know about the old URL. When to add the line?

PR #2470, opened 2020-02-06
fix: document that Vue uses semver

### LINKED ISSUE #930: Clarity in the docs if Vue 2.*.* is following semver strictly.
Looking at the issues on github, it seems it might, but I couldn't find a specific statement in the docs that Vue 2.. is strictly semver, and therefore offer full backward compatibility. Have I missed it or is not stated anywhere ?

sarahdayan:
This PR adds a note on the fact that Vue uses semver.

fix #930


bencodezen:
Thanks for the addition @sarahdayan!

PR #2473, opened 2020-02-11
Intro to Vuejs

mohsinafridi:
playing with Vuejs


NataliaTepluhina:
Thank you for the contribution but I think this example is fine as it is

PR #2480, opened 2020-02-16
Update list.md

st0ky:
According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice the start parameter of splice function is resettled to array.length if it greater than it ...


NataliaTepluhina:
Thank you for the point made, but I don't think changing anything on docs to `TODO` is a good idea.

st0ky:
What I did mean was that the workaround doesn't work.. it will not change
the length of the array...

PR #2486, opened 2020-02-19
Update state-management.md

adamwirth:
making the code examples consistently use `const` instead of switching to `var` as you scroll down on https://vuejs.org/v2/guide/state-management.html#Simple-State-Management-from-Scratch

### DIFF
--- a/src/v2/guide/state-management.md
+++ b/src/v2/guide/state-management.md
@@ -35,7 +35,7 @@ Now whenever `sourceOfTruth` is mutated, both `vmA` and `vmB` will update their 
 To help solve this problem, we can adopt a **store pattern**:
 
 ``` js
-var store = {
+const store = {
   debug: true,
   state: {
     message: 'Hello!'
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
Thank you for your contribution! We're using ES5 intentionally on Vue 2 docs, this will be changed on Vue 3 docs.

adamwirth:
oh, maybe the block in this same file:


``` js
const sourceOfTruth = {}
const vmA = new Vue({
  data: sourceOfTruth
})
const vmB = new Vue({
  data: sourceOfTruth
})
```

should be `var` then?

phanan:
@adamwirth Indeed! Would you want to send a PR for it?

adamwirth:
> @adamwirth Indeed! Would you want to send a PR for it?

Ok! I made [a new PR](https://github.com/vuejs/vuejs.org/pull/2488) 😄

PR #2488, opened 2020-02-19
Update state-management.md #2

adamwirth:
Use `var` consistently. previous PR was using `const`, it should be using `var`.

Ref: https://github.com/vuejs/vuejs.org/pull/2486#issuecomment-588314467


phanan:
Thanks!

PR #2490, opened 2020-02-23
fix: clarify how a Vue app works

### LINKED ISSUE #2475: Misleading documentation - Section "Getting started"
I wasn´t able to find the right place to report my problem, so I decided to create an issue. Please feel free to tell me, where I should place problems like the following:

Yesterday I started working with vue.js. After reading and understanding the "Getting started" section (https://vuejs.org/v2/guide/index.html) I started developing. Now, my colleagues told me, that I did something wrong: For each div-container I created a new Vue instance. So instead of creating one application with many components I created many application without any components.

I told to my colleagues, that I exactly did, what the documentation says. And they aggreed: The docs suggests to create many applications, so it´s misleading.

It would be very nice, if someone could make this point clear in the documentation.

Thanks in advance

Patrick

sarahdayan:
This clarifies how a Vue app works, as discussed [here](https://github.com/vuejs/vuejs.org/issues/2475#issuecomment-589985148).

@phanan I didn't use a `tip` block since I imagine it's not a good idea to overuse them; let me know if this kind of one-off paragraphs require one though 🙂

fixes #2475

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -73,7 +73,7 @@ var app = new Vue({
 
 We have already created our very first Vue app! This looks pretty similar to rendering a string template, but Vue has done a lot of work under the hood. The data and the DOM are now linked, and everything is now **reactive**. How do we know? Open your browser's JavaScript console (right now, on this page) and set `app.message` to a different value. You should see the rendered example above update accordingly.
 
-Note that we no longer have to interact with the HTML directly. A Vue app attaches itself to a single, global DOM element (`#app` in our case) then fully controls it. The HTML is our entry point, but everything else happens within the newly created Vue Instance.
+Note that we no longer have to interact with the HTML directly. A Vue app attaches itself to a single, global DOM element (`#app` in our case) then fully controls it. The HTML is our entry point, but everything else happens within the newly created Vue instance.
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/index.md
```diff
@@ -73,6 +73,8 @@ var app = new Vue({
 
 We have already created our very first Vue app! This looks pretty similar to rendering a string template, but Vue has done a lot of work under the hood. The data and the DOM are now linked, and everything is now **reactive**. How do we know? Open your browser's JavaScript console (right now, on this page) and set `app.message` to a different value. You should see the rendered example above update accordingly.
 
+Note that we no longer have to interact with the HTML directly. A Vue app attaches itself to a single, global DOM element (`#app` in our case) then fully controls it. The HTML is our entry point, but everything else happens within the newly created Vue instance.
```
phanan: This is great! I have but one tiny issue: What's a _global_ DOM element? Can/should we refer to it more simply as a "single DOM element" to keep beginners from getting further confused?

sarahdayan: Yup you're right, I wasn't too sure about it as well. Fixing this right now!

phanan:
Beautiful! Thanks @sarahdayan!

PR #2494, opened 2020-02-25
feat: add rule on arrow functions in option values

### LINKED ISSUE #1197: Style Guide: Add Priority D rule about using arrow functions for component option values
See https://github.com/vuejs/vuejs.org/pull/1193 for reference.

sarahdayan:
This adds a Priority D rule about the usage of arrow functions in component option values, as discussed by @chrisvfritz in #1193.

fixes #1197

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -1937,6 +1937,50 @@ Vue.component('TodoItem', {
 
 
 
+### Arrow functions for component option values <sup data-p="d">use with caution</sup>
+
+**Arrow functions should be avoided for component option values.**
+
+When using ES6 arrow functions, the value of `this` is lexically bound. Using arrow functions for option values can result in errors, as `this` might be `undefined` instead of referring to the current instance.
+
+It's recommended to use the [method syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions) instead, or a regular [function expression](https://developer.mozilla.org/en-US/docs/web/JavaScript/Reference/Operators/function).
+
+{% raw %}<div class="style-example example-bad">{% endraw %}
+#### Bad
+
+``` js
+data: () => {
+  return {
+    price: 150
+  }
+},
+computed: {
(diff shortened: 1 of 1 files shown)


sarahdayan:
@NataliaTepluhina You're right, the only problem in the code is `computed` but technically the `data` properties would work. I'll update this 🙂

phanan:
I wasn't a part of the original discussion, so I guess I'll add my 2¢ here: I see this rule as a "yes and no." 

Personally, I now almost always consciously use arrow functions if my methods (as in `methods` option and lifecycle hooks) don't need to access the Vue instance. If we see `this` as a variable (because it is), albeit implicit, having it in a function that _doesn't make use of it_ sounds pretty superfluous to me. This also more or less follows the "Principle of Least knowledge," the way I see it.

While arrow functions may be a pitfall for beginners due to them not having a lexical `this`, I'd argue that it's actually _good_ to fall for it, because it'll make them investigate and better understand the problem.

phanan:
Also IMHO, using (or not) arrow function isn't a matter of _style_, and thus doesn't belong in the _Style guide_.

sarahdayan:
Hey @phanan, thanks a lot for your input 🙂 It indeed doesn't make sense to avoid using arrow functions if you don't access `this`. However, I do believe there's value in documenting how going with method syntax is _usually_ a better choice, because it always succeeds, and you don't have to change it when you start using `this` in a function that didn't beforehand.

Additionally:

1. Developers who are aware of how `this` is bound will know they can disregard the rule, and go with arrow functions when they know they don't need to access `this`. We could add a small note to explain that arrow functions are okay in that scenario.
2. Beginners may indeed fall into that scenario and wonder what's going on. It's a good opportunity for digging, and it feels like having this documentation entry can be the answer for them.

Regarding whether it belongs to the style guide, I indeed agree that using arrow function isn't a _style_ choice per se (although many developers use it for code terseness), but many entries in the style guide aren't purely stylistic (e.g., keyed `v-for`, non-flux state management).

What do you think?

PR #2497, opened 2020-02-25
fix: deduplicate change detection caveat sections

### LINKED ISSUE #2352: Repeated reactivity explanations
I noticed that both [Reactivity in Depth (section Change Detection Caveats)](https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats) and [List Rendering (section Object Change Detection Caveats)](https://vuejs.org/v2/guide/list.html#Object-Change-Detection-Caveats) mention the caveats of creating object properties and reactivity. They use slightly different, but fundamentally the same examples.

While this is certainly an important thing to mention, its repetition on both of those pages strikes me as odd and slightly confusing as a user coming back to the guides. I had a brief moment of uncertainty when I realized that it felt as though I've already read that section.

I'm not sure if this is significant enough to be worth correcting. Personally I feel like linking to the Reactivity in Depth page would be an easier to maintain and less confusing solution, though the importance of this explanation, as well as the name of the Reactivity in Depth guide in the section Internals suggesting it's an advanced topic and it's low position on the sidebar menu make me slightly hesitant to change it.

Thoughts?

sarahdayan:
This deduplicates the object change detection caveats sub-section that was on the [**Reactivity in Depth**](https://vuejs.org/v2/guide/reactivity.html#Change-Detection-Caveats) and [**List Rendering**](https://vuejs.org/v2/guide/list.html#Object-Change-Detection-Caveats) sections.

For consistency, it also moved the array change detection caveats sub-section from List **Rendering sections** to **Reactivity in Depth**, so all caveats are in a single place.

fixes #2352


src/v2/guide/reactivity.md
```diff
@@ -20,7 +20,11 @@ Every component instance has a corresponding **watcher** instance, which records
 
 ## Change Detection Caveats
 
-Due to the limitations of modern JavaScript (and the abandonment of `Object.observe`), Vue **cannot detect property addition or deletion**. Since Vue performs the getter/setter conversion process during instance initialization, a property must be present in the `data` object in order for Vue to convert it and make it reactive. For example:
+Due to limitations in JavaScript, there are types of changes that Vue **cannot detect**. However, there are ways to circumvent them to preserve reactivity.
+
+### With Objects
```
phanan: Is it "Change detection caveats _for_ objects" or "_with_ objects"? Not a native speaker myself, so…

src/v2/guide/reactivity.md
```diff
@@ -20,7 +20,11 @@ Every component instance has a corresponding **watcher** instance, which records
...
+
+Vue **cannot detect property addition or deletion**. Since Vue performs the getter/setter conversion process during instance initialization, a property must be present in the `data` object in order for Vue to convert it and make it reactive. For example:
```
phanan: Since we have emphasized that Vue `**cannot**` in the leading paragraph (L23), emphasizing it again (L27) and again (L62) sounds pretty dramatic IMHO 😅.

src/v2/guide/reactivity.md
sarahdayan: I asked a native speaker colleague and indeed, "for" is best!

src/v2/guide/reactivity.md
sarahdayan: Yep! I'll also remove the redundant "Due to limitations in JavaScript" in the **For Arrays** sub-section.

sarahdayan:
🎉 🚀

phanan:
Thanks @sarahdayan!

PR #2500, opened 2020-02-27
[Doc EN]: `transition.md`, example 4, add a forgotten instruction

Haeresis:
Hi @sarahdayan and @chrisvfritz,

Just a small change about the transition file: one instruction in live example is not present in text example associate :)

The line 526 (before)-527 (after) was copy/past to line 482 (just deploy twice (maybe third) the code below my change from tab « Files changed » to see the line I have added).

It is now resolved!


phanan:
Thanks!

Haeresis:
Thank you to you for the review @phanan \o/

PR #2505, opened 2020-03-02
docs: add a complete demo before learner start

raywill:
Without a full demo, the beginner grab the idea much harder.

With a easy demo, they could understand the following documents much easier.


phanan:
Thanks for the PR 🙌 but I don't think it's necessary.

PR #2508, opened 2020-03-03
[proposal] Show how to integrate with select2 multiple selections

maki10:


phanan:
Thanks for the PR, but I don't think it's necessary, given that it's still the same library, only with different configurations.

PR #2512, opened 2020-03-04
Add new banner for Vuemastery free weekend

pieer:


themes/vue/source/css/page.styl
```diff
@@ -14,6 +14,7 @@
 @import "_style-guide"
 @import "_modal"
 @import "_scrimba"
+@import "_vuemastery-banner"
 @import "_vue-mastery"
```
davidtwamley: For https://github.com/vuejs/vuejs.org/issues/2515 I suspect this line needs to match the file rename with `@import _vuemastery` or revert the filename change. Explains an empty https://vuejs.org/css/page.css anyways. Thanks!

PR #2513, opened 2020-03-04
It was showing a missing iframe

nkb-bd:
Added working pen


NataliaTepluhina:
Thank you for the effort! Codesandbox is back to normal now so this change is not needed

nkb-bd:
Great !

PR #2519, opened 2020-03-07
Move hyphen to correct place

jrefior:
I think it was just a mistake that the hyphen was placed between well and decoupled instead of between reasonably and well.


phanan:
Thanks for the effort, but I believe the current usage is fine as it is.

jrefior:
That's fine: after sleeping on it I realized either position could be correct depending on the intention of the author.

PR #2521, opened 2020-03-09
feat: replace awesome-vue with VueCommunity

Akryum:


sdras:
DBcUtoo

Akryum:
As there is a lot of duplicate work going on, we are discussing moving some things from awesome-vue to Vue community, and awesomejs.dev already does the job of listing packages.

Akryum:
I've opened a PR so it's more clear for everyone: https://github.com/vuejs/awesome-vue/pull/3363

bencodezen:
@Akryum Thanks for the update! As I'm reflecting on the larger JS ecosystem though, it seems the AwesomeJS pattern is more common across different libraries and frameworks. 

It seems like a better approach would be to incorporate Vue Community into AwesomeVue instead and then grow it from there. This would also help with some of the conflicts we're running into with official Vue projects versus community based ones.

What do you think?

PR #2523, opened 2020-03-10
Add Factor platform as an SSR option

arpowers:


phanan:
Thanks for your contribution. However, a more suitable place for this would be https://github.com/dobromir-hristov/vuecommunity.

arpowers:
Why's that? @phanan You have two equivalents already on the same page?

PR #2525, opened 2020-03-11
Convert security@vuejs.org email link to mailto link

sohaibalam67:
Earlier the link pointed to
`https://vuejs.org/v2/guide/security@vuejs.org` which is not valid url.
Now it is a mailto link.


phanan:
Nice catch, thanks a lot!

PR #2528, opened 2020-03-15
Normalize quotes, remove useless attributes

pepelsbey:
Google Fonts used to have single-quoted code snippet with useless `type="text/css"` attributes attached. But not anymore :)


phanan:
Thanks!

PR #2534, opened 2020-03-19
fix: fix example of regex in "Form Validation".

whisky-shusuky:
Exisiting example throw an error. Deleted code by this pull request is unnecessary.


phanan:
Indeed, this looks like a typo. Thanks!

PR #2545, opened 2020-04-08
doc(guide/reactivity): object spread syntax

simon04:


phanan:
Thanks for the PR! To prepare for a smooth migration to v3, we're currently limiting changes made to v2 docs to critical issues only. Since yours doesn't appear to be one, I'd have to skip it for now.

PR #2547, opened 2020-04-09
Add tip about `done` callbacks in transition hooks

adamalton:
It took me *hours* to figure out why my `onAfterEnter` function was not being called. It turns out that if your `onEnter` function *accepts* the `done` parameter but does not *use* the parameter then the `onAfterEnter` function never gets called.

JSFiddle to show the behaviour: https://jsfiddle.net/adamalton/7f0kca4m/54/

I considered making a bug report for it, but from looking at discussions on similar issues it seems that Vue's sniffing of the function parameters is intentional, so the behaviour is arguably deliberate. So instead I'm just adding a tip to the docs so that others after me don't need to go through the same pain 🙂.

### DIFF
--- a/src/v2/guide/transitions.md
+++ b/src/v2/guide/transitions.md
@@ -442,6 +442,8 @@ These hooks can be used in combination with CSS transitions/animations or on the
 
 <p class="tip">When using JavaScript-only transitions, **the `done` callbacks are required for the `enter` and `leave` hooks**. Otherwise, the hooks will be called synchronously and the transition will finish immediately.</p>
 
+<p class="tip">When using CSS transitions, **the `enter` and `leave` functions should not accept the `done` parameter if they do not use it**. Including the `done` parameter in the function definition signals to Vue that you want to have control over the transition, so not calling it can cause unexpected behavior.</p>
+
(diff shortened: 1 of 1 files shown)


phanan:
Thanks for the PR! To prepare for a smooth migration to v3, we're currently limiting changes made to v2 docs to critical issues only. Since yours doesn't appear to be one, I'd have to skip it for now.

adamalton:
It was pretty critical to me! 🙂 

Is the transitions functionality being removed in v3? And if not, how can I get this change into the v3 docs?

Thanks.

phanan:
Transitions will still be there, and v3 docs will be open for public contribution soon :)

adamalton:
Ok, thanks. Can this be re-opened then, and perhaps tagged appropriately, so that it can be included when the v3 docs are open?  Leaving it as "closed" seems misleading.

bencodezen:
@adamalton Absolutely. I've added a label to make sure we consider this after the new docs have been released!

bencodezen:
@adamalton Sorry for the confusion.

To give you credit for finding this issue and the suggestion, can you create an issue so we can label it and resume the conversation after the new docs are released?

Feel free to tag me when you do so I can label it accordingly.

PR #2550, opened 2020-04-11
Update conditional.md

J-rayX:
Added conditional rendering code to describe another v-show use case.


phanan:
Thanks for the PR. However, I don't think this addition is necessary.

PR #2551, opened 2020-04-12
Wrong last name algorithm

carlosmunozrodriguez:
This is a classic error when trying to get the last name on a non-English language. E.g. in Latin America is common to have two last names.
It would be  better if the expression gets simplified to `this.lastName = names[1]` assuming just one first name and one last name


phanan:
Thanks for the PR. I don't think this is necessary, however, as the code only serves as an example. Name parsing is much more complex (if not impossible).

PR #2556, opened 2020-04-20
stop opening tabs when user selects a language

clorichel:
As a user, when I change the language of the website, I don't expect to get another tab to be opened, and consequently be forced to close the original tab, loosing its history.

Example on MDN:
- go there: https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
- pick your preferred language
- _(you stay on the same tab and)_ just read your content in your preferred language!

Example on Vuejs.org:
- go there: https://vuejs.org/v2/guide/mixins.html
- pick your preferred language
- another tab is opened
- you decide to keep yet another tab or close the original tab (loosing its history)
- then read your content in your preferred language!

This user experience seems quite annoying in my opinion, hence this proposal. But anyone please feel free to elaborate if I'm wrong about that! 🤓

### DIFF
--- a/themes/vue/layout/partials/language_dropdown.ejs
+++ b/themes/vue/layout/partials/language_dropdown.ejs
@@ -1,14 +1,14 @@
 <li class="nav-dropdown-container language">
   <a class="nav-link">Translations</a><span class="arrow"></span>
   <ul class="nav-dropdown">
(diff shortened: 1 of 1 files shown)


phanan:
I agree. Thanks for the PR!

PR #2560, opened 2020-04-21
Add details link for puzzled readers

mirabilos:
Add link to documentation on just how computed dependencies are tracked.
I was really puzzled reading this at first (“WTF, does it decompile the function or how?”) and learnt of the answer only after quite some searching.

### DIFF
--- a/src/v2/guide/computed.md
+++ b/src/v2/guide/computed.md
@@ -77,7 +77,7 @@ console.log(vm.reversedMessage) // => 'eybdooG'
 
 You can open the console and play with the example vm yourself. The value of `vm.reversedMessage` is always dependent on the value of `vm.message`.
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina:
Thank you for creating the PR! However, I don't think that explaining Vue reactivity principles fits this chapter well. We have in explained within [Reactivity in Depth](https://vuejs.org/v2/guide/reactivity.html#How-Changes-Are-Tracked)

mirabilos:
Natalia Tepluhina dixit:

>Thank you for creating the PR! However, I don't think that explaining
>Vue reactivity principles fits this chapter well.

But a link to the explanation is *direly* needed in that place, for
readers getting crazy about it. Just a link to follow for only these
who need the information there (in some other sections of the guide,
such links *are* provided, which I’ve found helpful).

>We have in explained within [Reactivity in
>Depth](https://vuejs.org/v2/guide/reactivity.html#How-Changes-Are-Tracked)

Can we please get a short one-sentence link to there then, from that
place?

Looking at the thing you linked, I find it incomprehensible, even
after having read the skyronic post, though… so please open some
discussion instead of dismissing it out of hand… as a forced new‐
comer to vue.js and the craziness that contemporary ECMAscript is,
I find it direly needed.

PR #2561, opened 2020-04-22
Updated question criteria

mayunited12:
Standard, question mark should end questions. 
This change updates code to still look for question mark in the entry and still make sure question mark ends entry.


bencodezen:
Thanks for the PR @mayunited12! At this time we're in the process of updating the docs and not accepting new changes, but we appreciate your suggestion and look forward to your thoughts on the next iteration!

PR #2568, opened 2020-04-26
Adding spaces between lines

juleshwar:
At a first glance, it is difficult to comprehend which comment belongs to which line as they are sandwiched between two lines of code.
Therefore adding spaces to add clarity.


bencodezen:
Hi @juleshwar. Thanks for the PR! We're currently in the midst of migrating the docs and updating it; so we will have to decline the PR at this time so we can ensure that all code examples follow a standard. We will keep your suggestion in mind though. Appreciate it!

PR #2575, opened 2020-05-03
For emitting custom events show what does work.

tettoffensive:
When quickly going through the docs, it's confusing that it shows what doesn't work. I propose we at least add what does work.

We may even consider changing it further, so that the code blocks are just what does work and the text explanation tells more details about what doesn't work.


phanan:

PR #2577, opened 2020-05-05
Prop type validation order matters

Gotterbild:
- Update props validation example to highlight the fact that prop type validation order matters
- Add a notice about empty String is converted to Boolean `true`

Having this in docs would have saved about 3 hours of my life.


phanan:

PR #2578, opened 2020-05-06
fix: forms.html, set the vue el property

rob2468:


bencodezen:
Thanks for taking the time to create this PR @rob2468! 

While it's understandable why you are including the wrapper div around code examples, the goal here is to demonstrate how simple the HTML can be with the use of v-model rather than show the full code users need to make the example work. So we will not be merging this change.

rob2468:
@bencodezen I know. But this page does not have a consistent rule currently.
Or would you like to delete the `example-3`?
https://vuejs.org/v2/guide/forms.html

image

bencodezen:
@rob2468 😱You are totally right.

To be consistent with the rest of the examples, let's remove the wrapper div in addition to the JS block beneath it because none of the other ones have it at this time.

In the future, we should definitely consider whether to have the JS block or not on this page.

Can you create a new PR and tag me when it's ready? Thanks!

PR #2582, opened 2020-05-11
docs: fix typo "Bublé" -> "Babel"

vipranarayan14:


NataliaTepluhina:
Thank you for your contribution! However, this fix is not necessary as we actually meant [Buble](https://github.com/bublejs/buble) here

vipranarayan14:
Ohk. Never heard of that before.

PR #2583, opened 2020-05-13
fix: forms.html, remove the wrapper div

rob2468:
ref #2578 
@bencodezen


bencodezen:
Thanks for the quick turnaround @rob2468! Appreciate it!

PR #2584, opened 2020-05-15
[nitpicking] import() is not a function

FelixSchuSi:
This is just a very small detail, but a dynamic import is not a funtion(see https://v8.dev/features/dynamic-import#dynamic).


phanan:
Thanks!

PR #2586, opened 2020-05-16
Update comparison.md

zackheisnberg:


bencodezen:
Hi @zackheisnberg. Thanks for your PR! 

In the comparison guide, we aim to provide as objective as an opinion as humanly possible. Each developer has a unique journey and skillset that might make one easier or another; but that is not for us to decide. 

Appreciate the effort you made to create the PR though! And good luck with your adventures with Vue in the meantime!

PR #2587, opened 2020-05-16
How to run a development server in a Docker container

rafawalter:
Useful to speed up new developer on boarding/setup/cleanup. Also helps avoid dependency conflicts between projects


phanan:

PR #2594, opened 2020-05-24
Update conditional.md

gabrielsroka:
is it `key` or `:key` or `v-bind:key` ?


phanan:
Thanks for the effort, but it's `key` since there's no binding here.

PR #2596, opened 2020-05-27
Name Casing modify

onlystp417:
The manual of Name Casing “With PascalCase” section, mention about:
" both <MyComponentName> and <MyComponentName> are acceptable ".
but <MyComponentName> and <MyComponentName> could be misunderstand to DOM template ,
I modify it to 'MyComponentName' and 'MyComponentName',
may be much proper.

Note
====


phanan:
Thanks, but the change is incorrect. Also please read the PR template.

PR #2598, opened 2020-05-30
Check if a question ENDS with a question mark.

timothyokooboh:
The change made to the example codebase checks to see if a question ENDS with a question mark. Instead of checking if a question CONTAINS a question mark.

E.g. Do you love me?  // This will now pass as a valid question.

E.g. Do you? love me.  // This will no longer pass as a valid question.

Note
====


phanan:
Thanks for the effort, but please be aware of the PR template note.

PR #2614, opened 2020-06-04
fix: Add "boolean" and "symbol" types for "key" attribute.

mpawelski:
Booleans as keys are supported since [v2.4.2](https://github.com/vuejs/vue/releases/tag/v2.4.2)([commit](https://github.com/vuejs/vue/commit/be3dc9c6e923248bcf81eb8240dd4f3c168fac59)) and symbols since [v2.5.12](https://github.com/vuejs/vue/releases/tag/v2.5.12)([commit](https://github.com/vuejs/vue/commit/bacb911f7df09ff4868b4c848a6d7778872dff5c)).

I created PR as suggested [here](https://github.com/vuejs/vue/issues/7936#issuecomment-636984051). I think it now makes clear that this types are supported and not just "happens to work by accident". 

I noticed that [v-for](https://vuejs.org/v2/api/#v-for) has "since _version_" in type so I did the same.

### DIFF
--- a/src/v2/api/index.md
+++ b/src/v2/api/index.md
@@ -2379,7 +2379,7 @@ type: api
 
 ### key
 
-- **Expects:** `number | string`
+- **Expects:** `number | string | boolean (since 2.4.2) | symbol (since 2.5.12)`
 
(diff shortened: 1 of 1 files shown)


bencodezen:
Great update @mpawelski! Thanks for the in depth research and linking the to the resources which made the review much easier. Really appreciate it! 🙇🏽‍♂️

PR #2629, opened 2020-06-28
Update form-validation.md

nurulmac11:
iteration objects requires key.

Note
====


phanan:
Since this is not a critical error, I'm going to close it per the PR template. You're welcome to create an _issue_ which can then be labeled as `post-3.0` to tackle later.

PR #237, opened 2016-02-16
Update transitions.md

jhserodio:
change transition exemple

PR #2644, opened 2020-07-18
add option for next in guide

acidiney:
I added the option in the version select to go to the [v3](https://v3.vuejs.org)

### PR on docs-next
It seems that the v3 documentation does not follow the same sequence as for the `introduction.md' file, in all versions of vuejs it is as index.md, but in version 3 it was renamed to introduction.md.
PR on docs-next [#237](https://github.com/vuejs/docs-next/pull/237)

### DIFF
--- a/themes/vue/layout/partials/sidebar.ejs
+++ b/themes/vue/layout/partials/sidebar.ejs
@@ -21,6 +21,7 @@
           <%- titles[type] %>
           <% if (['cookbook', 'style-guide'].indexOf(type) === -1) { %>
             <select class="version-select">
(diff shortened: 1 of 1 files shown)


themes/vue/layout/partials/sidebar.ejs
```diff
@@ -21,6 +21,7 @@
           <%- titles[type] %>
           <% if (['cookbook', 'style-guide'].indexOf(type) === -1) { %>
             <select class="version-select">
+              <option value="v3">next</option>
```
bencodezen, line 24: 3.x-beta

PR #2646, opened 2020-07-21
Improve capitalize function

MorrisMalone:
Let's not assume the input string is in lowercase ;)

Note
====


NataliaTepluhina:
@MorrisMalone Thank you for the contribution! We try not to complicate examples with additional logic as it creates an additional cognitive load for readers but I really appreciate taking an effort to improve the documentation 👍🏻

PR #2666, opened 2020-08-15
Update syntax.md

andriannus:
## Description
I've read the Template Syntax page (v2 docs and v3 docs). I think there are a few sentences that need adjusting.
Bdw, I'm using Grammarly for better results.

Thank you 🙂

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -98,7 +98,7 @@ These expressions will be evaluated as JavaScript in the data scope of the owner
 {{ if (ok) { return message } }}
 ```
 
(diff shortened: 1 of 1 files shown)


src/v2/guide/syntax.md
```diff
@@ -167,7 +167,7 @@ Dynamic argument expressions have some syntax constraints because certain charac
 
 The workaround is to either use expressions without spaces or quotes, or replace the complex expression with a computed property.
 
-When using in-DOM templates (templates directly written in an HTML file), you should also avoid naming keys with uppercase characters, as browsers will coerce attribute names into lowercase:
+When using in-DOM templates (templates are directly written in an HTML file), you should also avoid naming keys with uppercase characters, as browsers will coerce attribute names into lowercase:
```
bencodezen, line 170:
```suggestion
When using in-DOM templates (i.e., templates written directly in an HTML file), you should also avoid naming keys with uppercase characters, as browsers will coerce attribute names into lowercase:
```

bencodezen:
Thanks for the PR @andriannus!

PR #2667, opened 2020-08-18
fixed profile images of Jos Gerards,Vincent Mayers,Luke Thomas. 

tola:
This is to fix profile images of Jos Gerards,Vincent Mayers,Luke Thomas.I added imageUrl for each person coz avatars.io does not work anymore. 

Note
====


bencodezen:
Thanks for the PR @tola!

PR #2669, opened 2020-08-19
Add flatlogic themes

shoudaos:


phanan:
LGTM. Would you mind creating the same for docs-next?

PR #2673, opened 2020-08-22
Update syntax.md

BrianHilson:
Giving examples using camelcase for dynamic attribute names is confusing to beginners, as this won't work when they try to use it in an HTML template. Having the explanation at the end of that section (line 170) is good, but the examples given should lead learners in the right direction whenever possible.

If people have strong feelings against snake case, then no special casing could be used instead, ie. attributename.

Note
====

### DIFF
--- a/src/v2/guide/syntax.md
+++ b/src/v2/guide/syntax.md
@@ -139,18 +139,18 @@ Starting in version 2.6.0, it is also possible to use a JavaScript expression in
 Note that there are some constraints to the argument expression, as explained
 in the "Dynamic Argument Expression Constraints" section below.
 -->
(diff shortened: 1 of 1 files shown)


BrianHilson:
@bencodezen, I can certainly see the importance of using consistence synatx throughout the documentation. Given that using the CDN is one of the first options presented in the docs (https://vuejs.org/v2/guide/installation.html#CDN), and that the guide seems to be focused on beginners, I think it's important that the examples given work for people looking to get started in the simplest way possible. I include myself in that group.

I've found the documentation I've worked through so far to be very well written, clear, and easy to understand. This example was the only one where I had to play around with the example code for a bit to get it to work.

Thanks for your response and feedback. I will create an issue for this.

Issue #2668, opened 2020-08-19
change tabs


zhytang:
Web address: https://vuejs.org/v2/guide/components-dynamic-async.html
Keep going up until you can't see.

![QQ20200819-153533-HD](https://user-images.githubusercontent.com/16872071/90606015-c6837400-e231-11ea-913a-3462b5133bf7.gif)

bencodezen:
Thanks for finding and reporting the bug @zhytang!

Justineo:
Fixed by #2700.

PR #2675, opened 2020-08-23
fix: avoid page scroll when switching tabs

willamesoares:
Not entirely sure on the root cause for this behavior reported in #2668 but I guess that is somehow related to the fact that when changing tabs we are removing then adding an entire new div to the DOM instead of manipulating the contents of an existing one. It seems like it needs to be wrapped in a div (or some other tag?).

Updated behavior
tab_vue

PS.: this can still be used/revisited post v3 so opening a PR.


Justineo:
Thanks for the PR! #2700 should be a more precise fix for this issue so we merged that one instead. But still thanks for your contribution!

PR #2682, opened 2020-08-28
Update index.md

willi-berger:
minor fix:. in line 333. should be v-bind:todo="item.text" and not v-bind:todo="item"

Note
====

### DIFF
--- a/src/v2/guide/index.md
+++ b/src/v2/guide/index.md
@@ -330,7 +330,7 @@ Now we can pass the todo into each repeated component using `v-bind`:
     -->
     <todo-item
       v-for="item in groceryList"
-      v-bind:todo="item"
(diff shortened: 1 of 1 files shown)


bencodezen:
Thanks for the PR @willi-berger! After reviewing the PR, the current code is correct as inside of `todo-item` the template refers to 

```
template: '<li>{{ todo.text }}</li>'
```

which means if we pass it `todo.text` as a prop, this line would be invalid.

Appreciate the PR though! Let me know if you have any other questions.

willi-berger:
thanks for the  quick reply, of course you are right. There was an error in my local template definition which is/was
``` 
template:'<li>{{todo}}</li>'
```

PR #2691, opened 2020-09-08
docs: fixed the tip format

Jinjiang:
Small format fix (`::: tip` is only for vuepress not for Hexo)


bencodezen (approved):
Thanks for catching this @Jinjiang!

PR #2700, opened 2020-09-22
fix(#2668): reset unexpected behavior introduced by overflow-anchor

Justineo:
The problem is caused by the `overflow-anchor` property, which was supported by Chrome a while ago.

Closes both #2668 and #2675.


NataliaTepluhina (approved):
Thank you @Justineo 🙇🏻

swanprincess386:
> The problem is caused by the `overflow-anchor` property, which was supported by Chrome a while ago.
> 
> Closes both #2668 and #2675.

PR #2701, opened 2020-09-25
feat: add v3 banners

phanan:
As discussed with @NataliaTepluhina, this PR adds the banners to promote v3.x docs more prominently in v2.x.


NataliaTepluhina (approved):
Thank you An! Looks great to me 👍🏻

PR #2702, opened 2020-09-25
chore: update the PR template now that v3 is officially out

phanan:
Now that v3 is officially out and v3 doc is 100% public, I reckon we should update the PR template ;)


NataliaTepluhina (approved):
We totally should! =)

PR #2721, opened 2020-10-26
Update computed.md

mpvismer:
It is hard to make documentation as good as this any better, but I did find the "Computed vs Watched Property" slightly confusing because it comes before the "Watchers" section. This PR moves the  "Computed vs Watched Property" to the end and removes the first sentence of it.

Note
====


3imed-jaberi (approved):
LGTM 👍

PR #2728, opened 2020-11-13
Update components-props.md

JasperWoo:
Looks like from the example the label is a `prop`. The original example definitely works, but just for completeness I think label would make the example for cohesive.


NataliaTepluhina (approved):
@JasperWoo nice catch! Thank you 👍🏻

PR #2787, opened 2021-03-17
chore: use SVG logo where applicable

phanan:
I've just realized that we're using a PNG logo everywhere when a much smaller (195B gzipped vs 3KB) SVG can do just fine. This PR does just that: replace the PNG version with SVG where applicable.

If everyone is on board, I'll port this over to v3 naturally.


themes/vue/layout/layout.ejs
```diff
@@ -21,7 +21,7 @@
     <meta property="og:type" content="article">
     <meta property="og:title" content="<%- page.title ? page.title + ' — ' : '' %>Vue.js">
     <meta property="og:description" content="<%- theme.site_description %>">
-    <meta property="og:image" content="https://<%- theme.root_domain %>/<%- url_for("/images/logo.png") %>">
+    <meta property="og:image" content="https://<%- theme.root_domain %>/images/logo.png">
```
phanan, line 24: It's quite lame that OG protocol [doesn't support SVG](https://indieweb.org/The-Open-Graph-protocol#Does_not_support_SVG_images) but I'm taking this opportunity to make this consistent with the `twitter:image` meta tag declaration.

bencodezen (approved):
Looks great. Thanks for discovering that and making the improvement!

phanan:
Oops can't believe I totally forgot about this!

PR #2807, opened 2021-04-21
docs: add limitations to TypeScript section

caugner:


caugner:
@ffxsam I would like to add these TypeScript-related tips directly to the official Vue docs, loosely based on your [vue-typescript-cookbook](https://github.com/ffxsam/vue-typescript-cookbook). Would you mind reviewing these? Maybe you also have some other limitations in mind?

(I struggled quite a lot with migrating a medium-size project to TypeScript and would have loved to find these information directly in the Vue Guide, hence this PR.)

ffxsam:
@caugner Awesome! I took a look and it looks good to me. 👍 

The only other limitation I can think of is when you wrap a method in another function like `debounce`, e.g.:

```ts
methods: {
  someHandler: _debounce(async function (this: any) => {
    // Code with references to 'this'
  })
}
```

The `this: any` is necessary, otherwise it yells at you because it has no idea what `this` looks like. But I couldn't figure out how to actually annotate `this` properly to refer to the current component (kind of a cyclical chicken & egg issue). `this: any` works, but isn't ideal.

That's all I can think of!

caugner:
@ffxsam The recipe from the cookbook didn't work for me, but I found out that debouncing a method like *that* is not recommended, see: https://github.com/ffxsam/vue-typescript-cookbook/issues/11#issue-862734850

This is the solution:

```js
mounted() {
  this.someHandler = _.debounce(this.someHandler)
}
methods: {
  someHandler() {
    // Code with references to 'this'
  }
}
```

But it's probably worth mentioning this in the limitations, so I will add it.

ffxsam:
Ah, sorry I missed your comment, I've had my hands full lately. That's great, I'll have to update my code to use that method instead. Thanks!

caugner:
@ffxsam Added in 1320d1602e57ed087455eeac40bf3d3730c9cad6.

andikaleonardo (approved):

caugner:
Hi @bencodezen :wave:, I saw that you merged some PRs in this repository :nerd_face:, and was wondering whether you or someone else from the awesome Vue core team could have a short look at this small improvement. Thanks! :pray:

caugner:
I suppose this PR won't ever land, so I will just go ahead and close it.

PR #2808, opened 2021-04-22
Fix broken link to repo explaining Instance names

AustinWoetzel:
There is a link that is supposed to provide more detail on how data can be "overwritten" in an Instance when you do not use a $ in front of the name. The repo that the link points to has been updated and the path is no longer valid. I replaced it with what I think should be the new path.

Note: the repo link is in reference to a You-Dont-Know-JS (book series). The change in file path in that repo came because they are working on a 2nd edition. The 2nd edition is not released and does not have any info on Instance names. I referenced the archived 1st edition. It is possible that by the time this PR is processed, the 2nd edition will be released and the link may need to be changed to a file path in the 2nd edition.

### DIFF
--- a/src/v2/cookbook/adding-instance-properties.md
+++ b/src/v2/cookbook/adding-instance-properties.md
@@ -58,7 +58,7 @@ new Vue({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


NataliaTepluhina (approved):
Thank you for the fix! 👍🏻

PR #2833, opened 2021-06-15
fix : return data object

BirajMainali:
Note
====


bencodezen:
Thanks for the PR @BirajMainali. At this time, the current example is still accurate and a simpler example for people to comprehend, so we'll be keeping it as is.

PR #2839, opened 2021-06-20
docs: revamp Style Guide

phanan:
As discussed, this is my attempt to revamp our Style Guide with inspiration taken from [Dart](https://dart.dev/guides/language/effective-dart/design). The list of changes is as follow:

- Instead of priorities ("essential," "strongly recommended" etc.) we use descriptive verbs: DO, DON'T, AVOID, PREFER, and CONSIDER for the guidelines
- As a side effect, each guideline's header (which is presented in the sidebar) is now a complete sentence, making it easier to skim through
- Since priorities are no longer a thing, the guidelines are now grouped by what they target, for example, "naming," "templating," "patterns & practice" etc. This, however, can use some suggestions from the team.

### DIFF
--- a/src/v2/style-guide/index.md
+++ b/src/v2/style-guide/index.md
@@ -7,45 +7,12 @@ This is the official style guide for Vue-specific code. If you use Vue in a proj
 
 For the most part, we also avoid suggestions about JavaScript or HTML in general. We don't mind whether you use semicolons or trailing commas. We don't mind whether your HTML uses single-quotes or double-quotes for attribute values. Some exceptions will exist however, where we've found that a particular pattern is helpful in the context of Vue.
 
-> **Soon, we'll also provide tips for enforcement.** Sometimes you'll simply have to be disciplined, but wherever possible, we'll try to show you how to use ESLint and other automated processes to make enforcement simpler.
 
-Finally, we've split rules into four categories:
+## Naming
 
+### DO use multi-word component names
 
-
-## Rule Categories
-
-### Priority A: Essential
-
-These rules help prevent errors, so learn and abide by them at all costs. Exceptions may exist, but should be very rare and only be made by those with expert knowledge of both JavaScript and Vue.
-
-### Priority B: Strongly Recommended
-
-These rules have been found to improve readability and/or developer experience in most projects. Your code will still run if you violate them, but violations should be rare and well-justified.
-
-### Priority C: Recommended
-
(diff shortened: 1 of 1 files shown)


phanan:
Preview available at https://deploy-preview-2839--vuejs.netlify.app/v2/style-guide/.

NataliaTepluhina:
@phanan fair warning: as the PR is quite big, the review will take some time 😅

phanan:
Totally understand. However, I’d strongly suggest looking at the deploy
preview instead (or use it as the main ref), as the git diff makes little
sense 😅

LinusBorg:
This would have an impact on eslint-plugin-vue in a way, as the presets offered by this plugin are named after the current priorities naming scheme.

So if we merge this, we should also adjust at least the plugin docs to mention how which preset maps to which section of the style guide.

phanan:
I’m not familiar with eslint-plugin-vue admittedly, but all the rules are
kept as-is, only re-organized.

LinusBorg:
To be clear I don't think this is a big issue. But to clarify:

- Right now, rules that we document as "essential" in the style guide are found in the `essential` preset (amongst others that are not part of the style guide, but simply enforce correct template syntax).
- rules the document as "recommended" are in the `recommended` preset, which also includes as rules from "essential" etc.

So we might want to consider coordinting this change with a change of language in the eslint plugin

phanan:
@LinusBorg Good point. TBH I'm not sure what direction to take. @yyx990803 @ota-meshi WDYT?

ota-meshi:
I don't think we probably need to redirect. I think we need to update the link from the eslint-plugin documentation.

I think the style guides need a "How to read the guides" section, just like the Dart documentation.
https://dart.dev/guides/language/effective-dart#how-to-read-the-guides

After that, I think the eslint-plugin will need to reconfigure the presets in the next major version as follows:

- `essential` preset ... Follow `DO` and `DONT`.
- `strongly-recommended` preset ... Follow `PREFER` and `AVOID`.
- `recommended` preset ... Follow `CONSIDER`.

However, the `essential` preset is also used in Vetur and the user cannot change the configuration, so we need to consider what to do with items that have choices.

e.g. 

> #### DO have a consistent usage strategy for directive shorthands
> Directive shorthands (: for v-bind:, @ for v-on: and # for v-slot) should be used either **always** or **never**.

https://deploy-preview-2839--vuejs.netlify.app/v2/style-guide/#DO-have-a-consistent-usage-strategy-for-directive-shorthands

I have a question that is not related to eslint. Do the `DO`, `DONT`, `PREFER`, `AVOID`, and `CONSIDER` categories work well with translated documents in each language?
I've never translated a document so I'm not familiar with it. However, it may be difficult to bring `DO` etc. to the beginning depending on the language. I thought it might affect each translation team as well.

PR #2888, opened 2021-10-04
Update components-slots.md

stephane303:
It should be parent not child here

Note
====


bangunbagustapa:
I think the docs is correct, the data only available in the child component.

stephane303:
If I take the above example:
````
<navigation-link url="/profile">
  Clicking here will send you to: {{ url }}
  <!--
  The `url` will be undefined, because this content is passed
  _to_ <navigation-link>, rather than defined _inside_ the
  <navigation-link> component.
  -->
</navigation-link>
```
What is <navigation-link>, the parent or the child ? To me it seems to be the parent, hence my suggested modification.

PR #2911, opened 2021-11-24
fix typo in class name

aletar89:
Note
====


bencodezen:
Hi @aletar89! If I'm not mistaken, the new class name only applies to Vue 3. Vue 2 still uses the normal `.enter` and not `.enter-from` class. If I'm mistaken, feel free to open another issue with a repo that shows an example of this.

Thanks for the PR!

If I'

PR #2940, opened 2022-04-09
Add missing new RegExp

matks:
I might have missed something as I'm not great with JavaScript but maybe `new RegExp()` is missing


PululuK:
FYI : https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#utiliser_test

matks:
Interesting @PululuK 👍 so it works indeed.

PR #2998, opened 2023-08-30
Update components.md

amirhoseinsalehi:
Fixed a logical issue on an input.

### DIFF
--- a/src/v2/guide/components.md
+++ b/src/v2/guide/components.md
@@ -442,7 +442,8 @@ When used on a component, `v-model` instead does this:
 ``` html
 <custom-input
   v-bind:value="searchText"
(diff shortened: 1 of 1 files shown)


Justineo:
Thank you but the original code is correct. `v-model` expects custom component emit the whole `value` to its parent.

amirhoseinsalehi:
> Thank you but the original code is correct. `v-model` expects custom component emit the whole `value` to its parent.

No problem, the issue is where you pass the whole event to the value prop which doesn't work correctly(object passed as a value).

Justineo:
The child component should do `$emit('input', value)` where `value` is string so the received `$event` as a string.

amirhoseinsalehi:
Aha, You are right.

